PR #2, opened 2020-12-08
Add hub for realtime multiplayer

peppy:
Initial logic pass. Doesn't yet have an actual client implementation (so serialization has not been tested sufficiently yet).

Potentially will be split out into its own project eventually, but made too much sense to keep it here while we can.

- [x] Depends on https://github.com/ppy/osu/pull/11127 (use local game reference to test)

### DIFF
--- a/osu.Server.Spectator.Tests/MultiplayerFlowTests.cs
+++ b/osu.Server.Spectator.Tests/MultiplayerFlowTests.cs
@@ -18,9 +18,14 @@ namespace osu.Server.Spectator.Tests
         private readonly MultiplayerHub hub;
 
         private const int user_id = 1234;
+        private const int user_id_2 = 2345;
+
         private const long room_id = 8888;
 
         private readonly Mock<IMultiplayerClient> mockReceiver;
+
+        private readonly Mock<HubCallerContext> mockContextUser1;
+        private readonly Mock<HubCallerContext> mockContextUser2;
 
         public MultiplayerFlowTests()
         {
@@ -30,16 +35,20 @@ namespace osu.Server.Spectator.Tests
 
             Mock<IGroupManager> mockGroups = new Mock<IGroupManager>();
 
-            Mock<HubCallerContext> mockContext = new Mock<HubCallerContext>();
-            mockContext.Setup(context => context.UserIdentifier).Returns(user_id.ToString());
+            mockContextUser1 = new Mock<HubCallerContext>();
+            mockContextUser1.Setup(context => context.UserIdentifier).Returns(user_id.ToString());
+
+            mockContextUser2 = new Mock<HubCallerContext>();
+            mockContextUser2.Setup(context => context.UserIdentifier).Returns(user_id_2.ToString());
 
             Mock<IHubCallerClients<IMultiplayerClient>> mockClients = new Mock<IHubCallerClients<IMultiplayerClient>>();
             mockReceiver = new Mock<IMultiplayerClient>();
             mockClients.Setup(clients => clients.Group(MultiplayerHub.GetGroupId(room_id))).Returns(mockReceiver.Object);
 
-            hub.Context = mockContext.Object;
             hub.Groups = mockGroups.Object;
             hub.Clients = mockClients.Object;
+
+            setUserContext(mockContextUser1);
         }
 
         [Fact]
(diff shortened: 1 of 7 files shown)


osu.Server.Spectator/Hubs/MultiplayerHub.cs
```diff
@@ -0,0 +1,358 @@
...
+                    throw new InvalidStateChange(oldState, newState);
+
+                case MultiplayerUserState.FinishedPlay:
+                    if (oldState != MultiplayerUserState.Playing)
+                        throw new InvalidStateChange(oldState, newState);
+
+                    break;
+
+                case MultiplayerUserState.Results:
+                    if (oldState != MultiplayerUserState.Playing)
```
bdach, line 312: Shouldn't this be checking `FinishedPlay`, or just blocking this on the client side entirely? I guess it'll be more apparent after a client implementation lands...

osu.Server.Spectator/Hubs/MultiplayerHub.cs
```diff
@@ -0,0 +1,358 @@
...
+using System.Threading.Tasks;
+using Microsoft.Extensions.Caching.Distributed;
+using osu.Game.Online.RealtimeMultiplayer;
+
+namespace osu.Server.Spectator.Hubs
+{
+    public class MultiplayerHub : StatefulUserHub<IMultiplayerClient, MultiplayerClientState>, IMultiplayerServer
+    {
+        // for the time being rooms will be maintained in memory and not distributed.
+        private static readonly Dictionary<long, MultiplayerRoom> active_rooms = new Dictionary<long, MultiplayerRoom>();
```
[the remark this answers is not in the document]
peppy: i switched away from ConcurrentDictionary for that reason specifically. Doesn't seem like the class we want here.

Keeping locking simple until performance is actually a problem (doubt it will ever be).

osu.Server.Spectator/Hubs/MultiplayerHub.cs
peppy: good catch, should be denied. i changed how the flow worked halfway through.

osu.Server.Spectator.Tests/MultiplayerFlowTests.cs
```diff
@@ -0,0 +1,375 @@
...
+
+            await hub.JoinRoom(room_id);
+            await hub.ChangeSettings(testSettings);
+
+            mockReceiver.Verify(r => r.SettingsChanged(testSettings), Times.Once);
+        }
+
+        #endregion
+
+        private void setUserContext(Mock<HubCallerContext> context) => hub.Context = context.Object;
```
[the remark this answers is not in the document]
peppy: doesn't work well as `await` calls can not occur inside a delegate.

osu.Server.Spectator/Hubs/MultiplayerHub.cs
```diff
@@ -0,0 +1,388 @@
...
+        /// <summary>
+        /// Should be called when user states change, to check whether the new overall room state can trigger a room-level state change.
+        /// </summary>
+        private async Task updateRoomStateIfRequired(MultiplayerRoom room)
+        {
+            //check whether a room state change is required.
+            switch (room.State)
+            {
+                case MultiplayerRoomState.WaitingForLoad:
+                    if (room.Users.All(u => u.State != MultiplayerUserState.WaitingForLoad))
```
bdach, line 262: From what I gathered, the user can always back out to `Idle`, in light of which this should probably be checking

```suggestion
                    if (room.Users.All(u => u.State == MultiplayerUserState.Loaded))
```

and have a special case for all users re-entering `Idle`. Here's a failing test case:

```csharp
        [Fact]
        public async Task AllUsersBackingOutFromLoadCancelsTransitionToPlay()
        {
            var room = await hub.JoinRoom(room_id);

            setUserContext(mockContextUser2);
            await hub.JoinRoom(room_id);

            await hub.ChangeState(MultiplayerUserState.Ready);

            setUserContext(mockContextUser1);
            await hub.ChangeState(MultiplayerUserState.Ready);
            await hub.StartMatch();

            Assert.Equal(MultiplayerRoomState.WaitingForLoad, room.State);

            await hub.ChangeState(MultiplayerUserState.Idle);
            setUserContext(mockContextUser2);
            await hub.ChangeState(MultiplayerUserState.Idle);

            Assert.Equal(MultiplayerRoomState.Open, room.State);
        }
```

osu.Server.Spectator/Hubs/MultiplayerHub.cs
```diff
@@ -0,0 +1,388 @@
...
+                    {
+                        foreach (var u in room.Users)
+                            await changeAndBroadcastUserState(room, u, MultiplayerUserState.Playing);
```
bdach, line 265: Upon reading `StartMatch()`, it is possible that only part of the room has previously entered gameplay, which in light of this code creates the possibility of wrongly changing state of users that have never readied up. `Room` should probably have a `UsersInGameplay` or similar (`ReadiedUpUsers`?)

Failing test case for this:

```csharp
        [Fact]
        public async Task OnlyReadiedUpUsersTransitionToPlay()
        {
            var room = await hub.JoinRoom(room_id);
            await hub.ChangeState(MultiplayerUserState.Ready);

            setUserContext(mockContextUser2);
            await hub.JoinRoom(room_id);

            setUserContext(mockContextUser1);
            await hub.StartMatch();
            Assert.Equal(MultiplayerRoomState.WaitingForLoad, room.State);
            Assert.Single(room.Users, u => u.State == MultiplayerUserState.WaitingForLoad);
            Assert.Single(room.Users, u => u.State == MultiplayerUserState.Idle);

            await hub.ChangeState(MultiplayerUserState.Loaded);
            Assert.Single(room.Users, u => u.State == MultiplayerUserState.Playing);
            Assert.Single(room.Users, u => u.State == MultiplayerUserState.Idle);
        }
```

osu.Server.Spectator/Hubs/MultiplayerHub.cs
```diff
@@ -0,0 +1,388 @@
...
+
+                    break;
+
+                case MultiplayerRoomState.Playing:
+                    if (room.Users.All(u => u.State != MultiplayerUserState.Playing))
+                    {
+                        foreach (var u in room.Users)
+                            await changeAndBroadcastUserState(room, u, MultiplayerUserState.Results);
+
+                        await changeRoomState(room, MultiplayerRoomState.Open);
```
bdach, line 280: See above pair of comments, but with `FinishedPlay`.

bdach (commented):
Sorry if these are getting annoying because this isn't finished yet - hard for me to tell what state this is in, just wanting to help out, is all.

osu.Server.Spectator/Hubs/MultiplayerHub.cs
peppy: This is really weird, I thought I already had this implemented correctly..

peppy:
Awesome feedback @bdach, thanks!

bdach (approved):
A little tired as I read this, but looks like a solid initial pass at this stage to me :+1:

smoogipoo (approved):
Lgtm

PR #3, opened 2020-12-09
Add sample client implementation for multiplayer

peppy:
- [x] Depends on https://github.com/ppy/osu-server-spectator/pull/2

This is slightly different than the spectator sample client in that it allows interactive control of a set of multiplayer clients. Should allow for some good (manual) test coverage of edge case scenarios.

Of note: signalr replies are not instant. If you perform an action and the client states are not as you expect, hit enter once to allow them to refresh again. (I've added a `Thread.Sleep` to avoid this in most cases).


bdach:
A little difficult to review this, given how intertwined this history is with #2. Gonna chill with reading this through for now.

smoogipoo (approved):

PR #4, opened 2020-12-09
Add room database retrieval and validation

peppy:
- [x] Depends on https://github.com/ppy/osu-server-spectator/pull/3

Note that this currently breaks the sample client project (or, rather, it requires a database for it to function correctly). I'm figuring out the best way to fix this, so while this PR is usable for testing against an actual database backing, don't merge just yet.

PR #5, opened 2020-12-11
Add basic token revocation / expiry checks

peppy:
Logic is based off what we already have on `osu-notification-server` (https://github.com/ppy/osu-notification-server/blob/master/src/oauth-verifier.ts#L93-L99).


smoogipoo (approved):

peppy:
This actually had a dependency on #4 for the `Database` helper class, oops.

PR #9, opened 2020-12-16
Fix clients polluting the states of other client types

smoogipoo:
When leaving a multiplayer room, the spectator client (via EndPlaySession) would remove the multiplayer state from the cache, causing the multiplayer client actions to fail.


bdach:
Yeah, this is a nasty catch with `nameof()` - it makes sense knowing that it works at compile-time, but still...

peppy (approved):

PR #11, opened 2020-12-17
Update the host UserID in the database after transfer

peppy:
This wasn't happening, causing a stale listing display and (potentially) incorrect client-side permissions if the information from the lobby is used for actual room display.

### DIFF
--- a/osu.Server.Spectator/Hubs/MultiplayerHub.cs
+++ b/osu.Server.Spectator/Hubs/MultiplayerHub.cs
@@ -5,6 +5,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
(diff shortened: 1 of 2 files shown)


osu.Server.Spectator/Hubs/MultiplayerHub.cs
```diff
@@ -197,6 +198,9 @@ public async Task TransferHost(long userId)
                     throw new Exception("Target user is not in the current room");
 
                 room.Host = newHost;
+
+                await UpdateDatabaseHost(room);
```
smoogipoo, line 202: Should also happen when the host leaves without `TransferHost()`

smoogipoo (changes requested):

PR #16, opened 2020-12-22
Add checks to ensure a valid beatmap is selected when setting are changed

peppy:
Has been deployed for testing.


bdach:
No database env to actually test but seems sane and seems to work client-side (errors are raised). Mind you the rollback of the settings does nothing (because it'll be the await that fails, and the assignment to `room.Settings` just won't happen), but I suppose at least for safety/readability it's good to have it.

As discussed on stream osu-web needs a similar checksum check on room creation.

bdach (approved):

bdach:
absolutely boneheaded review above, but have tested a few more times and it works ok now

PR #18, opened 2020-12-23
Fix dapper not correctly handling DateTimeOffset when local system is not UTC

peppy:
As it turns out, dapper just doesn't handle the time zone portion of DateTimeOffset, full stop. Things have only been working until now because we are running our components on servers which are set to UTC.

This is the proposed fix to handle DateTimeOffset correctly, see https://stackoverflow.com/questions/12510299/get-datetime-as-utc-with-dapper for more details. We should probably deploy this to all other projects we have which user dapper to avoid similar issues, especially during local development.


bdach (approved):
passes the eye test

PR #20, opened 2020-12-24
Add support for "beatmap not downloaded / downloading" user states

frenzibyte:
- [ ] Depends on ppy/osu#11289

Note that I haven't tested this due to lack of a local dev server (I tried setting one, but I blew up on the way and closed everything).

### DIFF
--- a/osu.Server.Spectator/Hubs/MultiplayerHub.cs
+++ b/osu.Server.Spectator/Hubs/MultiplayerHub.cs
@@ -240,6 +240,7 @@ namespace osu.Server.Spectator.Hubs
                 // handle whether this user should be receiving gameplay messages or not.
                 switch (newState)
                 {
+                    case MultiplayerUserState.BeatmapNotDownloaded:
                     case MultiplayerUserState.FinishedPlay:
                     case MultiplayerUserState.Idle:
                         await Groups.RemoveFromGroupAsync(Context.ConnectionId, GetGroupId(room.RoomID, true));
--- a/osu.Server.Spectator/Hubs/MultiplayerHub.cs
+++ b/osu.Server.Spectator/Hubs/MultiplayerHub.cs
@@ -515,6 +515,16 @@ namespace osu.Server.Spectator.Hubs
         {
             switch (newState)
             {
+                case MultiplayerUserState.BeatmapNotDownloaded:
+                    // any state can switch to this.
+                    break;
+
+                case MultiplayerUserState.DownloadingBeatmap:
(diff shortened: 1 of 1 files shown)


osu.Server.Spectator/Hubs/MultiplayerHub.cs
```diff
@@ -515,6 +516,16 @@ private void ensureValidStateSwitch(MultiplayerRoom room, MultiplayerUserState o
         {
             switch (newState)
             {
+                case MultiplayerUserState.BeatmapNotDownloaded:
```
peppy, line 519: probably just combine with idle below, since they are basically the same thing.

peppy (changes requested):

osu.Server.Spectator/Hubs/MultiplayerHub.cs
frenzibyte: done.

bdach:
I'm not actually sure I'd agree with being able to transition into `BeatmapNotDownloaded` from any state. Surely it doesn't make sense to do so from something like `Playing`. I think the flow should be something like the following:

obraz

I guess there's an argument to be made that permitting any state to transition to `BeatmapNotDownloaded` is more resilient against a possible dropped message about a transition to `Idle`. I'd definitely permit `DownloadingBeatmap` -> `Ready` in that case, however.

peppy:
Yeah, while thinking about this i almost proposed keeping the map download state separate from the main state enum to keep things simple. Still thinking about this one.

frenzibyte:
> I'm not actually sure I'd agree with being able to transition into `BeatmapNotDownloaded` from any state. Surely it doesn't make sense to do so from something like `Playing`. I think the flow should be something like the following:
> 
> obraz
> 
> I guess there's an argument to be made that permitting any state to transition to `BeatmapNotDownloaded` is more resilient against a possible dropped message about a transition to `Idle`. I'd definitely permit `DownloadingBeatmap` -> `Ready` in that case, however.

Was thinking like that as well, previously I had `BeatmapNotDonwloaded` only accept being transitioned from `Ready`, `Idle` and `DownloadingBeatmap`, but I thought I was overcomplicating stuff, and maybe it would potentially be allowed to delete a map from whatever state you're in, for god knows what reason. 

But I see we're set in one direction now, will change as proposed.

frenzibyte:
> Yeah, while thinking about this i almost proposed keeping the map download state separate from the main state enum to keep things simple. Still thinking about this one.

I'll have a go on this one, sounds logical.

frenzibyte:
Closing as I'm going in a different direction, as above.

PR #22, opened 2020-12-25
Disallow restricted users from joining a game

peppy:
This is not supported behaviour and currently causes client crashes as a result.

### DIFF
--- a/osu.Server.Spectator/Hubs/MultiplayerHub.cs
+++ b/osu.Server.Spectator/Hubs/MultiplayerHub.cs
@@ -52,7 +52,7 @@ namespace osu.Server.Spectator.Hubs
 
         public async Task<MultiplayerRoom> JoinRoom(long roomId)
         {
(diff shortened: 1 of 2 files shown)


osu.Server.Spectator/Hubs/MultiplayerHub.cs
```diff
@@ -405,6 +412,17 @@ protected virtual async Task EndDatabaseMatch(MultiplayerRoom room)
             }
         }
 
+        protected virtual async Task<bool> CheckUserRestrictions()
```
bdach, line 415: I'd rename to `IsUserRestricted`. With a name like `CheckUserRestrictions` it's not really clear what a `true` value means.

bdach (changes requested):

bdach (approved):
Trusting that the DB names match. Would be nice to get test coverage eventually but good enough for now.

Edit: Checked web source to compare. Seems to check out. Should really look into setting up a local instance soon-ish though...

PR #24, opened 2020-12-30
Implement EntityStore for async safe state management with per entity locking

peppy:


bdach (commented):
Stared at this for the most part of the last hour and I can't see holes in this. There are some tiny comments but after having written down the core invariants of how this works it looks pretty solid as far as I can tell right now.

bdach:
I wrote this test to try and break destroying, and it sometimes seems to deadlock - not sure why. Probably my synchronisation in the test is deficient, as in debug it sometimes throws different exceptions, but it does seem to get stuck somewhat consistently.

```csharp
        [Fact]
        public async Task TestDestroyingInConcurrentUsages()
        {
            ManualResetEventSlim secondGetStarted = new ManualResetEventSlim();

            var firstGet = await store.GetForUse(1, true);
            var secondGet = Task.Run(async () =>
            {
                secondGetStarted.Set();

                using (await store.GetForUse(1))
                {
                }
            });

            secondGetStarted.Wait(1000);
            using (firstGet)
                firstGet.Destroy();

            await Assert.ThrowsAsync<ArgumentException>(async () => await secondGet);
        }
```

bdach:
I rewrote the above to not use `Task.Run`, as I though I was potentially wrecking myself with a task pool deadlock, and used a `Thread` instead - but it seems to still fail.

```csharp
        [Fact]
        public async Task TestDestroyingInConcurrentUsages()
        {
            ManualResetEventSlim secondGetStarted = new ManualResetEventSlim();
            ManualResetEventSlim secondGetEnded = new ManualResetEventSlim();
            Exception? exception = null;

            var firstGet = await store.GetForUse(1, true);
            var secondGetThread = new Thread(async () =>
            {
                secondGetStarted.Set();

                try
                {
                    using (await store.GetForUse(1))
                    {
                    }
                }
                catch (Exception ex)
                {
                    exception = ex;
                }

                secondGetEnded.Set();
            });

            secondGetThread.Start();
            secondGetStarted.Wait(1000);

            using (firstGet)
                firstGet.Destroy();

            secondGetEnded.Wait(10_000);
            Assert.IsType<ArgumentException>(exception);
        }
```

What I'm seeing:

* Place a single breakpoint at the inner `using` inside the thread action, and another right at the end
* Run test in debug
* Step over once
* Execution is never resumed in the inner thread
* At the final assert, `secondGetEnded` was never set, exception was never set, and the started thread seems to be missing from the list of threads

Again, likely that I messed something up, but I don't get what's happening here at all...

osu.Server.Spectator/Hubs/EntityStore.cs
```diff
@@ -0,0 +1,158 @@
...
+        /// Get all tracked entities in an unsafe manner. Only read operations should be performed on retrieved entities.
+        /// </summary>
+        public KeyValuePair<long, T?>[] GetAllEntities()
+        {
+            lock (entityMapping)
+            {
+                return entityMapping
+                       .Where(kvp => kvp.Value.Item != null)
+                       .Select(entity => new KeyValuePair<long, T?>(entity.Key, entity.Value.Item))
+                       .ToArray();
```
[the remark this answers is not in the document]
bdach: Also loose thought about this - the signature is `KeyValuePair<long, T?>[]`, it seems that `null`s are valid to be stored as entities, but this method excludes `null`s. If that's to be the case, then maybe this should return a `KeyValuePair<long, T>[]` and use `.AsNonNull()` to elide the nullable?

bdach:
Added two other tests for operations that did not have coverage in this PR. Will now read through the other PR and maybe go back to my failing test above to see if I can figure out what's broken there.

osu.Server.Spectator/Hubs/EntityStore.cs
```diff
@@ -0,0 +1,158 @@
...
+            /// Mark this item as no longer used. Will remove any tracking overhead.
+            /// </summary>
+            public void Destroy()
+            {
+                // we should already have a lock when calling destroy.
+                checkValidForUse();
+
+                store.remove(id);
+                semaphore.Dispose();
+                isDestroyed = true;
```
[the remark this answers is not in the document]
peppy: I've moved this, but alone I think it's not enough. I also added a second validity check *after* the lock is satisfied.

osu.Server.Spectator/Hubs/EntityStore.cs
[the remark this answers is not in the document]
peppy: right now it's just used to send all spectators on new user connection.

osu.Server.Spectator/Hubs/EntityStore.cs
```diff
@@ -0,0 +1,158 @@
...
+                checkValidForUse(false);
+
+                if (!await semaphore.WaitAsync(lock_timeout))
+                    throw new TimeoutException($"Lock for {typeof(T)} id {id} could not be obtained within timeout period");
+            }
+
+            public void ReleaseLock()
+            {
+                if (!isDestroyed)
+                    semaphore.Release();
```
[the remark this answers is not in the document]
peppy: Can you explain what you mean here? What would a hard fail entail?

I'd hope this method is only ever called by the `EntityStore` but enforcing that isn't so simple without making interfaces or otherwise.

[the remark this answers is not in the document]
bdach: I just meant that this is the only public method of `TrackedEntity` (aside from `GetItemUnsafe()`) that does not call `checkValidForUse()` as a precondition before proceeding. Again not a huge deal.

osu.Server.Spectator/Hubs/EntityStore.cs
```diff
@@ -0,0 +1,165 @@
...
+            /// </summary>
+            public void Destroy()
+            {
+                // we should already have a lock when calling destroy.
+                checkValidForUse();
+
+                isDestroyed = true;
+
+                store.remove(id);
+                semaphore.Dispose();
```
bdach, line 138: I figured out why my test was failing here. It appears that it is in fact not safe to dispose a semaphore that other readers are waiting on. And by that I don't mean "it'll timeout", I mean "it will block forevermore". The [docs](https://docs.microsoft.com/en-us/dotnet/api/system.threading.semaphoreslim.dispose?view=net-5.0#System_Threading_SemaphoreSlim_Dispose) slightly hint at this:

> Call `Dispose` when you are finished using the `SemaphoreSlim`. The `Dispose` method leaves the `SemaphoreSlim` in an unusable state.

but I think they should be shouting that from the rooftops rather than this meek mention... If you compare the implementation of `Dispose()` and `Release()` it does look like the former just bails instead of cleaning up properly.

In light of that I think it's proper to release the semaphore before doing this here. It instantly fixes my test, for what it's worth.

bdach (changes requested):

bdach (approved):
Think this is OK now. Let's let it spin on prod and I'll merge tomorrow if nothing pops up.

peppy:
image

Going to need to alter the test to either test each scenario separately (not sure how easy that is given hoe much of an edge case this is) or handle both types of exceptions.

bdach:
On carefully re-reading the test again the type of exception that falls out of the second get comes down to who wins the race for the lock on `entityMapping`.

If the order is as follows:

```
FIRST									SECOND

secondGetStarted.Wait(1000);
										secondGetStarted.Set();
using (firstGet)
										using (await store.GetForUse(1)) [TAKES LOCK]
										{
										}
	firstGet.Destroy() [TAKES LOCK]
```

then the resultant exception is an `InvalidOperationException`. Conversely, if the order is

```
FIRST									SECOND

secondGetStarted.Wait(1000);
										secondGetStarted.Set();
using (firstGet)
	firstGet.Destroy() [TAKES LOCK]
										using (await store.GetForUse(1)) [TAKES LOCK]
										{
										}
```

then the resultant exception is an `ArgumentException`.

I don't think there's a sane way to test for this, as it's just contention on a lock. I'd say allowing both exceptions is OK to do here.

peppy:
What I *am* going to do is group those two exceptions together. While updating the test I realised that *every usage* would need to handle both of these exceptions otherwise.

peppy:
Check that now. Feels a lot better using `KeyNotFoundException`, i think.

osu.Server.Spectator/Hubs/EntityStore.cs
```diff
@@ -0,0 +1,192 @@
...
+
+        private const int lock_timeout = 5000;
+
+        /// <summary>
+        /// Retrieve an entity with a lock for use.
+        /// </summary>
+        /// <param name="id"></param>
+        /// <param name="createOnMissing"></param>
+        /// <returns>An <see cref="ItemUsage{T}"/> which allows reading or writing the item. This should be disposed </returns>
+        /// <exception cref="KeyNotFoundException">Thrown if <see cref="createOnMissing"/> was false and the item is not in a tracked state.</exception>
```
bdach, line 30: Hmm... this xmldoc is not 100% true as written. Even with `createOnMissing = true` a consumer can get this exception by accidentally winning the race on the `entityMapping` lock (gets entity from store before it's cleaned itself up). For 100% safety in that case there should just be a retry.

peppy: Should be resolved now.

osu.Server.Spectator/Hubs/EntityStore.cs
```diff
@@ -0,0 +1,203 @@
...
+            {
+                if (!entityMapping.TryGetValue(id, out item))
+                    // was not tracking.
+                    return;
+            }
+
+            await item.ObtainLockAsync();
+
+            // handles removal and disposal of the semaphore.
+            item.Destroy();
```
bdach, line 85: Maybe this should be catching (and ignoring) `InvalidOperationException` too? Hitting it would require multiple threads trying to destroy a single entity, but handling this case is simpler because receiving that exception just means someone's already done this work and there is nothing more to be done by this method.

peppy: It should be impossible to destroy from outside a locked context (see `Destroy`'s implementation). Are you saying we should add just for added safety?

bdach: That's not entirely what I meant - because the entity collection and entity locking are separate, there can be subtle order races between them. The scenario I'm thinking of is as follows:

```
FIRST												  SECOND

lock (entityMapping)
{
	if (!entityMapping.TryGetValue(id, out item))
		// succeeds
}

									                  lock (entityMapping)
									                  {
										                  if (!entityMapping.TryGetValue(id, out item))
											                  // succeeds
									                  }

await item.ObtainLockAsync(); // succeeds
item.Destroy();

													  await item.ObtainLockAsync();
													  // fails with InvalidOperationException
```

peppy: Ah, you mean on the `ObtainLockAsync` line, right? Seemed like your comment was attached to `.Destroy` which confused me. Sounds valid.

bdach: Yeah I suppose just catching on the lock obtain should suffice.

bdach:
I feel like this should be a sane enough state at this point to try merging and see if anything falls out. Still not completely 100% sure on the fine-grained locking (as in having separate locks for collection and item) - could potentially cause some weird ordering issues in extreme edge cases (i.e. a `GetForUse(..., true)` potentially not getting priority on the access after creating) - but we can deal with that as a follow-up, and get this in to get the ball rolling.

PR #25, opened 2020-12-30
Consume EntityStore and strengthen state consistency across both hubs

peppy:


peppy:
Currently seeing a permanently locked state... somehow

```

User 19301917 ending play session (Beatmap:2581495 Mods: Ruleset:0)
fail: Microsoft.AspNetCore.SignalR.HubConnectionHandler[1]
      Error when dispatching 'OnConnectedAsync' on hub.
System.TimeoutException: Lock for osu.Server.Spectator.Hubs.MultiplayerClientState id 2 could not be obtained within timeout
period
   at osu.Server.Spectator.Hubs.EntityStore`1.TrackedEntity.ObtainLockAsync() in /Users/dean/Projects/osu-server-spectator/osu.Server.Spectator/Hubs/EntityStore.cs:line 145
   at osu.Server.Spectator.Hubs.EntityStore`1.GetForUse(Int64 id, Boolean createOnMissing) in /Users/dean/Projects/osu-server-spectator/osu.Server.Spectator/Hubs/EntityStore.cs:line 41
   at osu.Server.Spectator.Hubs.StatefulUserHub`2.cleanUpState(Boolean isDisconnect) in /Users/dean/Projects/osu-server-spectator/osu.Server.Spectator/Hubs/StatefulUserHub.cs:line 85
   at osu.Server.Spectator.Hubs.StatefulUserHub`2.OnConnectedAsync() in /Users/dean/Projects/osu-server-spectator/osu.Server.Spectator/Hubs/StatefulUserHub.cs:line 46
   at Microsoft.AspNetCore.SignalR.Internal.DefaultHubDispatcher`1.OnConnectedAsync(HubConnectionContext connection)
   at Microsoft.AspNetCore.SignalR.Internal.DefaultHubDispatcher`1.OnConnectedAsync(HubConnectionContext connection)
   at Microsoft.AspNetCore.SignalR.HubConnectionHandler`1.RunHubAsync(HubConnectionContext connection)
User 17335964 beginning play session (Beatmap:2571414 Mods: Ruleset:0)
User 3830532 beginning play session (Beatmap:845391 Mods:TC Ruleset:0)
[osu.Game.Online.RealtimeMultiplayer.IMultiplayerClient] User 2 connected (83 total)!
```

Likely triggered via connection cleanup logic.

osu.Server.Spectator/Hubs/StatefulUserHub.cs
```diff
@@ -2,94 +2,107 @@
...
+                {
+                    bool isOurState = usage.Item.ConnectionId != Context.ConnectionId;
+
+                    if (isDisconnect && !isOurState)
+                        // not our state, owned by a different connection.
+                        return;
+                }
+
+                if (usage.Item != null)
+                    await CleanUpState(usage.Item);
```
bdach, line 77: Any reason not to merge this into the conditional above? This is all under a lock, so there should be no volatile changes, should there?

osu.Server.Spectator/Hubs/MultiplayerHub.cs
```diff
@@ -59,45 +47,44 @@ public async Task<MultiplayerRoom> JoinRoom(long roomId)
...
-                room.Host = roomUser;
-
-                lock (active_rooms)
+                using (var roomUsage = await ACTIVE_ROOMS.GetForUse(roomId, true))
                 {
-                    Console.WriteLine($"Tracking new room {roomId}.");
-                    active_rooms.Add(roomId, room);
+                    if (roomUsage.Item == null)
+                    {
+                        var retrievedRoom = await RetrieveRoom(roomId);
```
bdach, line 67: Side note: So this usage (and similar others) has a DB operation under a lock within, which makes all (non-read) DB operations touching a single room essentially serialised. Which is *proooobably* good (it's definitely easy to reason about); not sure how well it will scale, but you probably know better than I do off experience alone.

osu.Server.Spectator/Hubs/MultiplayerHub.cs
```diff
@@ -59,45 +47,44 @@ public async Task<MultiplayerRoom> JoinRoom(long roomId)
...
-                await UpdateDatabaseParticipants(room);
-            }
+                userUsage.Item = new MultiplayerClientState(Context.ConnectionId, CurrentContextUserId, roomId);
 
-            await UpdateLocalUserState(new MultiplayerClientState(roomId));
-
-            await MarkRoomActive(room);
+                // read only database operations are allowed to be done outside of the lock for performance reasons.
+                await UpdateDatabaseParticipants(room);
+                await MarkRoomActive(room);
```
bdach, line 84: *Are* these read-only? Both of these don't seem to be...

osu.Server.Spectator/Hubs/MultiplayerHub.cs
```diff
@@ -584,24 +550,60 @@ private void ensureIsHost(MultiplayerRoom room)
...
+        {
+            using (var roomUsage = await getLocalUserRoom(state))
             {
-                if (!active_rooms.TryGetValue(roomId, out room))
-                    failWithInvalidState($"User ({CurrentContextUserId}) is in a room ({roomId}) this hub is not aware of.");
-            }
+                var room = roomUsage.Item;
+
+                if (room == null)
+                    throw new InvalidOperationException("Attempted to operate on a null room");
```
bdach, line 570: I think throwing anything in `leaveRoom()` as is the case here can cause the user state to not be destroyed higher up in `LeaveRoom()` while it should probably be. Should most likely catch errors of this method higher up, destroy and re-throw. (in a finally block maybe?)

Also, similarly, `CleanUpState` probably shouldn't care about any exceptions thrown here? Not 100% sure on that one.

osu.Server.Spectator.Tests/SpectatorFlowTests.cs
```diff
@@ -54,8 +56,8 @@ public async Task NewUserConnectsAndStreamsData()
             mockReceiver.Verify(clients => clients.UserBeganPlaying(streamer_id, It.Is<SpectatorState>(m => m.Equals(state))), Times.Once());
 
             // check state data was added
-            var cacheState = await cache.GetStringAsync(SpectatorHub.GetStateId(streamer_id));
-            Assert.Equal(state, JsonConvert.DeserializeObject<SpectatorState>(cacheState));
+            // var cacheState = await cache.GetStringAsync(SpectatorHub.GetStateId(streamer_id));
+            // Assert.Equal(state, JsonConvert.DeserializeObject<SpectatorState>(cacheState));
```
bdach, line 60: would just trim if no longer needed

bdach (changes requested):

osu.Server.Spectator/Hubs/StatefulUserHub.cs
peppy: No reason, just an oversight

osu.Server.Spectator/Hubs/MultiplayerHub.cs
peppy: This is intentional, and yes, it may limit performance and scalability, but I am just looking to keep things simple to try and get state consistent before adding optimisations.

For reference, bancho queues all non-essential database operations (anything which *don't* need to be executed in sequence to ensure consistency) to a separate queue class, and I intend to do the same here eventually. These lower priority operations would occur outside of the lock after retrieving the data they need from the usage.

osu.Server.Spectator/Hubs/MultiplayerHub.cs
peppy: I mean read-only to the item, not the database. Will reword the comment.

peppy: Actually for simplicity and consistency I'm just going to move them inside the lock for now (turns out the access to `room.Users` wasn't thread-safe).

osu.Server.Spectator/Hubs/MultiplayerHub.cs
peppy: Moved to finally block to hopefully avoid this.

osu.Server.Spectator/Hubs/StatefulUserHub.cs
```diff
@@ -2,94 +2,106 @@
...
+            {
+                if (usage.Item != null)
+                {
+                    bool isOurState = usage.Item.ConnectionId != Context.ConnectionId;
+
+                    if (isDisconnect && !isOurState)
+                        // not our state, owned by a different connection.
+                        return;
+
+                    await CleanUpState(usage.Item);
```
bdach, line 75: As I attempted to (and probably failed to) convey earlier in https://github.com/ppy/osu-server-spectator/pull/25#discussion_r550166581, if this virtual `CleanUpState()` ever throws (as the `MultiplayerHub` one can), I think the `usage.Destroy()` lower down will never be called. Should probably be

```csharp
try
{
	if (usage.Item != null)
	// ...
}
finally
{
	usage.Destroy();
	usage.Dispose();
}
```

bdach:
Aside from the above I don't think I can see many other issues with this right now.

osu.Server.Spectator/Hubs/StatefulUserHub.cs
peppy: You bring up another interesting point: should `usage.Dispose` call destroy implicitly if it's not already destroyed? or are we happy with the system being closed enough that this is not a requirement?

bdach: Wait, dispose calling destroy? Dispose just frees the lock - otherwise all entities would get immediately deleted after usage. In fact `Destroy()` includes `Dispose()` now. In fact in the above I missed the fact that the `Dispose()` is gratuitous above - it's just a no-op at that point.

osu.Server.Spectator/Hubs/SpectatorHub.cs
```diff
@@ -18,7 +20,11 @@ public SpectatorHub([NotNull] IDistributedCache cache)
 
         public async Task BeginPlaySession(SpectatorState state)
         {
-            await UpdateLocalUserState(state);
+            using (var usage = await GetOrCreateLocalUserState())
+            {
+                usage.Item ??= new SpectatorClientState(Context.ConnectionId, CurrentContextUserId);
+                usage.Item.State = state;
+            }
```
bdach, line 27: Same as the other comment - broadcasting `state` to all users without a defensive copy inside this block is also similarly potentially unsafe (as someone has potentially grabbed the semaphore on the state and potentially started making changes to it - but this one is way smaller impact as there are no direct mutations to the state under this reference as far as I can see, the references are being replaced entirely on the entity if there's a change).

bdach (approved):
I thought of another potential issue but it's not easily resolveable in this PR without game-side changes to the data structures so going to approve for now (as I'm not sure we'll be hitting said issue any time soon).

Sorry for trickling these notes down like that. It's just that I can't seem to be able to think of everything in one go, it just sorta occurs to me step by step after a bit of thinking.

osu.Server.Spectator/Hubs/SpectatorHub.cs
peppy: Yep, but that will eventually be handled in a better way.

osu.Server.Spectator/Hubs/MultiplayerHub.cs
```diff
@@ -59,45 +47,43 @@ public async Task<MultiplayerRoom> JoinRoom(long roomId)
...
-                await UpdateDatabaseParticipants(room);
-            }
-
-            await UpdateLocalUserState(new MultiplayerClientState(roomId));
+                userUsage.Item = new MultiplayerClientState(Context.ConnectionId, CurrentContextUserId, roomId);
 
-            await MarkRoomActive(room);
-
-            return room;
+                return room;
```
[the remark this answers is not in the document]
peppy: yeah, we probably need to do a copy here. i can imagine the serialisation of users failing for reasons we've already encountered.

peppy:
Going to merge this as it hasn't *regressed* server behaviour at least.

Can work forward from a fresh slate.

osu.Server.Spectator/Hubs/StatefulUserHub.cs
peppy: Back to front, oops.

PR #26, opened 2021-01-02
Add datadog logging for entity store usages

peppy:
Bare minimum logging to track the usage of entity stores (and hopefully help track down issues)

This is already deployed live and has helped with some basic understanding. Have confirmed that everything is outputting as expected, but would appreciate a quick one-over on the calls from a correctness perspective.

### DIFF
--- a/osu.Server.Spectator/Hubs/EntityStore.cs
+++ b/osu.Server.Spectator/Hubs/EntityStore.cs
@@ -7,6 +7,7 @@ using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
 using osu.Framework.Extensions.ObjectExtensions;
+using StatsdClient;
 
 namespace osu.Server.Spectator.Hubs
 {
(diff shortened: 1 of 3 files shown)


osu.Server.Spectator/Hubs/EntityStore.cs
```diff
@@ -62,6 +70,7 @@ public async Task<ItemUsage<T>> GetForUse(long id, bool createOnMissing = false)
                     throw new KeyNotFoundException($"Attempted to get untracked entity {typeof(T)} id {id}");
```
bdach, line 70: this unlikely branch could probably also use an increment if we're tracking missing keys, similar/same as line 48

bdach:
Have looked at what the datadog operations are doing via inspecting decompiled IL in rider and it seems to do the right thing (events are enqueued onto an asynchronous worker which flushes every [1,100] ms. So I don't think adding this logging should be increasing lock contention much.

peppy:
Yep, the overhead is zero (and even if it wasn't the target server is localhost and it's using UDP). We already use this tens of thousands of times per second in bancho, web-10 and osu-web.

bdach (approved):
seems logical

PR #28, opened 2021-01-03
Remove no longer required invalid state method

frenzibyte:
That method was added for 100% code coverage, but since we don't care about that for now, this can be removed.


peppy (approved):
iu guess so

PR #29, opened 2021-01-03
Add support for beatmap availability state change handling

frenzibyte:
- [x] Depends on ppy/osu#11410
- [x] Depends on game bump

This adds the required methods to support storing each user's beatmap availability/download states and change them.

UPDATE: Tested these changes with my remaining ones lazer-side through a local web server, working perfectly.


osu.Server.Spectator.Tests/MultiplayerFlowTests.cs
```diff
@@ -563,13 +622,13 @@ private void verifyAddedToGameplayGroup(Mock<HubCallerContext> context, long roo
             => mockGroups.Verify(groups => groups.AddToGroupAsync(
                 context.Object.ConnectionId,
                 MultiplayerHub.GetGroupId(roomId, true),
-                It.IsAny<CancellationToken>()), wasAdded ? Times.Once : Times.Never);
+                It.IsAny<CancellationToken>()), wasAdded ? Times.AtLeastOnce : Times.Never);
```
bdach, line 625: explanation for this change? this weakens the assertion considerably in all other usages than the one you wanted this change for.

if you want to customise the number of times, make the times a param of this method and update existing usages.

osu.Server.Spectator/Hubs/MultiplayerHub.cs
```diff
@@ -202,6 +204,9 @@ public async Task ChangeState(MultiplayerUserState newState)
                 if (user == null)
                     failWithInvalidState("Local user was not found in the expected room");
 
+                if (user.BeatmapAvailability.State != DownloadState.LocallyAvailable)
+                    throw new InvalidStateException("Attempted to change user state while they don't have the room's beatmap.");
```
bdach, line 208: is wrong. will possibly throw if host changes beatmap while another user is still at results and keep the user stuck at results, depending on client implementation. this should only prevent changes to anything other than `Idle` if the map is not downloaded.

osu.Server.Spectator/Hubs/MultiplayerHub.cs
```diff
@@ -227,6 +232,40 @@ public async Task ChangeState(MultiplayerUserState newState)
...
+                    return;
+
+                user.BeatmapAvailability = newBeatmapAvailability;
+
+                // force state to idle when beatmap availability changes.
+                if (user.State != MultiplayerUserState.Idle)
+                {
+                    await Groups.RemoveFromGroupAsync(Context.ConnectionId, GetGroupId(room.RoomID, true));
+                    await changeAndBroadcastUserState(room, user, MultiplayerUserState.Idle);
+                }
```
bdach, line 263: this... does not actually set `user.State`. please add test coverage for this.

also i 100% guarantee doing this locally like here will break at some point if changing state to `Idle` does more things than it currently does. which is why i'm not sure that reporting downloaded state should automatically back out the user from idle state server-side. maybe this should be two operations client-side (availability changed + user transitioned to idle).

bdach (changes requested):

osu.Server.Spectator.Tests/MultiplayerFlowTests.cs
frenzibyte: thought it won't be, but yeah actually.

osu.Server.Spectator/Hubs/MultiplayerHub.cs
frenzibyte: this won't be the case, `ChangeBeatmapAvailability` will force the user state back to `Idle`, and clients probably should handle that change and take the user out of results screen and into the room sub screen.

bdach: the client is responsible for sending beatmap availability. will it be sent from the client in results screen if the host changes map? because if not, then this will throw. which is why i said "depending on client implementation".

also disagree on the latter part of that. if the user wants to keep viewing results, they should be free to do so for however long they choose.

frenzibyte: so just to be sure, you're fine with a user that is in `Ready` state, when host changes beatmap user doesn't have, the user will not get forced back to `Idle`, but let that be handled by the client instead, right?

bdach: i'm confused by this question. the server can't know if the user has the beatmap or not. unless you want to explicitly unready every user in the room on a map change. in fact, that might actually be safer.

frenzibyte: hmm, actually ok, can change the `Idle` revert to client, was thinking of the availability becoming more of something a server gets and handle whoever user has it properly, but it's starting to look like it should be more of just an information.

bdach: on second thought maybe this isn't as much of an issue as any setting change in the room reverts all users back to `Idle` already - but i'd still permit changing state to `Idle` regardless of whether you have the map or not, as it doesn't do any harm and prevents potential errors

frenzibyte: yeah sure.

osu.Server.Spectator.Tests/MultiplayerFlowTests.cs
```diff
@@ -475,6 +478,66 @@ public async void NotReadyUsersDontGetLoadRequest()
...
+
+            await hub.ChangeState(MultiplayerUserState.Loaded);
+
+            // ensure throws when changing beatmap availability while playing.
+            await Assert.ThrowsAsync<InvalidStateException>(() => hub.ChangeBeatmapAvailability(new BeatmapAvailability(DownloadState.NotDownloaded)));
+
+            await hub.ChangeState(MultiplayerUserState.FinishedPlay);
+
+            // ensure can change beatmap availability while in results, but state should be forced back to idle.
+            await hub.ChangeBeatmapAvailability(new BeatmapAvailability(DownloadState.NotDownloaded));
```
peppy, line 501: this seems wrong? what if the user was still at the results screen? why should they be forced back to idle? they are now showing an incorrect state.

peppy:
This needs conflict resolution.

osu.Server.Spectator/Hubs/MultiplayerHub.cs
```diff
@@ -202,6 +204,9 @@ public async Task ChangeState(MultiplayerUserState newState)
                 if (user == null)
                     failWithInvalidState("Local user was not found in the expected room");
 
+                if (user.BeatmapAvailability.State != DownloadState.LocallyAvailable && newState != MultiplayerUserState.Idle)
```
peppy, line 207: I'm highly unsure if this is something we want. Is there any reason for attaching these cross-state constraints for a display-only state? It seems like unnecessary complexity to me.

osu.Server.Spectator/Hubs/MultiplayerHub.cs
peppy: Fully agree on this. The download state was split out to *avoid* the complexity of changing the user's overall state, yet this change is adding weird intertwining rules which make things even harder to parse than if it was part of the main state enum.

Keep it as a display only state and don't require anything special from it. The other users in the room can choose whether to display the download state or not based on their own logic (ie. if the user is `Ready` the download state may be hidden even if the user is reporting downloading or not having the beatmap. or maybe it can just be left to display).

peppy (changes requested):

osu.Server.Spectator/Hubs/MultiplayerHub.cs
frenzibyte: Written it intentionally with safety checks such that nothing weird should ever be happening like a user playing while they don't own beatmap, but now that you all agree for the download states to be display-only, let's go with that.

osu.Server.Spectator/Hubs/MultiplayerHub.cs
frenzibyte: Same as https://github.com/ppy/osu-server-spectator/pull/29#discussion_r554730608

osu.Server.Spectator.Tests/MultiplayerFlowTests.cs
frenzibyte: Same as https://github.com/ppy/osu-server-spectator/pull/29#discussion_r554730608

osu.Server.Spectator/Hubs/MultiplayerHub.cs
frenzibyte: I'll apply changes once the dependency PR is merged first, as this is becoming a bit of a mess for me without the game package bumped.

peppy: If anything it should be the reverse directionality - if a user becomes `Ready` their beatmap download state should be assumed to be `Available` and should silently update. But I'm not even sure that is required; would opt to keep things as simple as possible to avoid the logical overhead for something which doesn't need it.

peppy:
@frenzibyte no updates on this?

frenzibyte:
> @frenzibyte no updates on this?

Done, updated game package along with the dependencies here, and removed all the checks and idle setting, ready for review and merge.

osu.Server.Spectator.Tests/MultiplayerFlowTests.cs
bdach: this is still not fixed

frenzibyte: ah....

bdach:
there are zero unit tests for this new operation now.

frenzibyte:
> there are zero unit tests for this new operation now.

I wasn't sure how useful writing a test case to check if `ChangeBeatmapAvailability(NotDownloaded)` will change the client's availability to `NotDownloaded`. I would like to know in the process at what point is it unnecessary to write a test case for something because I was really concerned about this.

bdach:
it's not only about checking that the user's state changed, it's also about checking whether that information was broadcast to the appropriate user group.

current server logic is small enough in size that in my opinion we should be aiming to as close to 100% test coverage as can reasonably get.

frenzibyte:
Gotcha, will write tests now then.

osu.Server.Spectator.Tests/MultiplayerFlowTests.cs
```diff
@@ -539,6 +551,42 @@ public async void NotReadyUsersDontGetLoadRequest()
...
+            Assert.True(room.Users.Single().BeatmapAvailability.Equals(user1Availability));
+
+            setUserContext(mockContextUser2);
+            await hub.ChangeBeatmapAvailability(user2Availability);
+            Assert.True(room2.Users.Single().BeatmapAvailability.Equals(user2Availability));
+
+            mockReceiver.Verify(c1 => c1.UserBeatmapAvailabilityChanged(user_id, It.Is<BeatmapAvailability>(b => b.Equals(user1Availability))), Times.Once);
+            mockReceiver.Verify(c1 => c1.UserBeatmapAvailabilityChanged(user_id_2, It.Is<BeatmapAvailability>(b => b.Equals(user2Availability))), Times.Never);
+
+            mockReceiver2.Verify(c2 => c2.UserBeatmapAvailabilityChanged(user_id, It.Is<BeatmapAvailability>(b => b.Equals(user1Availability))), Times.Never);
```
peppy, line 584: why was this added? it's not confirming anything the two lines above are not already.

osu.Server.Spectator/Hubs/MultiplayerHub.cs
```diff
@@ -275,6 +276,30 @@ public async Task ChangeState(MultiplayerUserState newState)
...
 
+        public async Task ChangeBeatmapAvailability(BeatmapAvailability newBeatmapAvailability)
+        {
+            using (var userUsage = await GetOrCreateLocalUserState())
+            using (var roomUsage = await getLocalUserRoom(userUsage.Item))
+            {
+                var room = roomUsage.Item;
+
+                if (room == null)
+                    throw new InvalidStateException("Attempted to operate on a null room");
```
peppy, line 287: another case of incorrect cleanup (should be resolved via #41 so not necessarily needing changes here).

peppy: This should be an `InvalidOperationException`. I'm not sure why you made it an exception which will be returned to the user (this is not how it's done anywhere else).

osu.Server.Spectator/Hubs/MultiplayerHub.cs
```diff
@@ -275,6 +276,30 @@ public async Task ChangeState(MultiplayerUserState newState)
...
+
+                var user = room.Users.Find(u => u.UserID == CurrentContextUserId);
+
+                if (user == null)
+                    throw new InvalidStateException("Local user was not found in the expected room");
```
peppy, line 292: This should also be an `InvalidOperationException`. This means there's a serious server-side error, not something triggered by the user.

peppy (changes requested):

osu.Server.Spectator.Tests/MultiplayerFlowTests.cs
```diff
@@ -539,6 +551,42 @@ public async void NotReadyUsersDontGetLoadRequest()
...
+
+            var user1Availability = BeatmapAvailability.Importing();
+            var user2Availability = BeatmapAvailability.Downloading(0.5);
+
+            setUserContext(mockContextUser1);
+            await hub.ChangeBeatmapAvailability(user1Availability);
+            Assert.True(room.Users.Single().BeatmapAvailability.Equals(user1Availability));
+
+            setUserContext(mockContextUser2);
+            await hub.ChangeBeatmapAvailability(user2Availability);
```
peppy, line 578: You aren't covering the case that this is triggered twice, where the second fire should not send a message to other users in the room. This would be visible if you run code coverage and check which lines are not covered by your tests.

osu.Server.Spectator.Tests/MultiplayerFlowTests.cs
frenzibyte: hmm, right, will remove this receiver then.

osu.Server.Spectator/Hubs/MultiplayerHub.cs
frenzibyte: Other similar cases actually had this `InvalidStateException`, which caused me into thinking that it's fine using this exception, but I've forced them all to be `InvalidOperationException` while I was excluding coverage for the other cases (https://github.com/ppy/osu-server-spectator/pull/43)

peppy (approved):

PR #30, opened 2021-01-03
Add an extra check to ensure two connections from the same user_id cannot join the same room

peppy:
As per the inline comments, this cannot happen in theory. If we are hitting this point, connection uniqueness has not been maintained (two connections / tracked states from the same user have been allowed on the server at once).

The addition of this check allows rooms to continue to remain in a good state even when the above occurs, at worst limiting the damage to the user (until their connections are lost).

### DIFF
--- a/osu.Server.Spectator/Hubs/MultiplayerHub.cs
+++ b/osu.Server.Spectator/Hubs/MultiplayerHub.cs
@@ -43,9 +43,7 @@ namespace osu.Server.Spectator.Hubs
             bool isRestricted = await CheckIsUserRestricted();
 
             if (isRestricted)
-            {
                 throw new InvalidStateException("Can't join a room when restricted.");
-            }
 
             using (var userUsage = await GetOrCreateLocalUserState())
             {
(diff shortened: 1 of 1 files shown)


bdach:
On second thought, do we actually know for sure that those were two disparate connections, rather than one sending duplicate messages or dropping some? I could well be convinced that client-side protections against the latter case are insufficient, and single-connection checks would do nothing against that.

Either way looks like there shouldn't be any harm in merging this. I might add a test case to be doubly sure.

peppy:
If it's the same connection it will abort a few lines earlier:

https://github.com/ppy/osu-server-spectator/pull/30/files#diff-90658690a1d84ef42d78f75a985211e9140da7fdadbcb3b5562eee9c5d894fd7R50-R54

This is intentionally returned to the client (see exception type) as opposed to the newly added one.

bdach:
Right, I did miss that, and there is even a test case for that. Seems non-trivial to test this new conditional without some deeper fiddling. I'm looking at possibilities of `room.Users` getting corrupted in some way but can't see any evidence of that - or at least not ones that #31 wouldn't theoretically fix already.

Eh, guess I click the button and we see if we can get detailed enough logging to nail whatever is causing this...

bdach (approved):

PR #31, opened 2021-01-03
Add some extra safeties against database failures

peppy:
This sees to multiple failures that could occur if a database execution was to fail for one reason or another. The common case I've seen come up is on the `updateParticipants`, which is now protected against all failures, but on the way I also noticed that due to no local exception handling, in combination with haphazard order of execution, this could result in a room or user state becoming inconsistent.

I've reordered all such operations to allow the database calls to occur first, and early-abort on error (allowing the client or server to retry the operation post-failure).

This also adds exception handling to the cleanup operation on new connection, to avoid the case where a previous connection's cleanup failure results in state being inconsistent (and never recovered). Now it will retry every time the user tries to reconnect to the server, and disallow a connection from the bad-state user if retries don't help.

### DIFF
--- a/osu.Server.Spectator/Hubs/MultiplayerHub.cs
+++ b/osu.Server.Spectator/Hubs/MultiplayerHub.cs
@@ -591,19 +591,19 @@ namespace osu.Server.Spectator.Hubs
                 if (user == null)
                     failWithInvalidState("User was not in the expected room.");
 
-                room.Users.Remove(user);
-
-                await UpdateDatabaseParticipants(room);
-
-                if (room.Users.Count == 0)
+                // handle closing the room if the only participant is the user which is leaving.
+                if (room.Users.Count == 1)
                 {
-                    Console.WriteLine($"Stopping tracking of room {room.RoomID} (all users left).");
-                    roomUsage.Destroy();
-
                     await EndDatabaseMatch(room);
 
+                    // only destroy the usage after the database operation succeeds.
+                    Console.WriteLine($"Stopping tracking of room {room.RoomID} (all users left).");
+                    roomUsage.Destroy();
                     return;
                 }
+
+                room.Users.Remove(user);
+                await UpdateDatabaseParticipants(room);
 
                 var clients = Clients.Group(GetGroupId(room.RoomID));
 
--- a/osu.Server.Spectator/Hubs/MultiplayerHub.cs
+++ b/osu.Server.Spectator/Hubs/MultiplayerHub.cs
@@ -360,13 +360,20 @@ namespace osu.Server.Spectator.Hubs
         {
             Debug.Assert(room.Host != null);
 
-            using (var conn = Database.GetConnection())
+            try
             {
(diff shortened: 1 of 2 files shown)


osu.Server.Spectator/Hubs/StatefulUserHub.cs
```diff
@@ -35,8 +35,18 @@ public override async Task OnConnectedAsync()
...
+            {
+                // if a previous connection is still present for the current user, we need to clean it up.
+                await cleanUpState(false);
+            }
+            catch
+            {
+                // if any exception happened during clean-up, don't allow the user to reconnect.
+                // this limits damage to the user in a bad state if their clean-up cannot occur (they will not be able to reconnect until the issue is resolved).
+                Context.Abort();
+                throw;
```
bdach, line 48: Think #30 might be rendered obsolete/unnecessary with this change? I'm not yet totally sold on merging checks like the one in that PR without at least partial understanding of what the mechanism of failure is there, as it might turn out we'll be patching such cases until the end of time because there's a larger underlying problem - and this seems like it could be a plausible explanation.

peppy: #30 is still required. it's still getting hit with both deployed, and even if it wasn't, i'd still want it in place as per the comment (the more failsafes the better; primary goal is to keep as much of the server working for as long as possible, isolating any issues)

this is how i always design server components: protect and then fix.

bdach:
Looking at other db operations that could potentially lead to inconsistent state, `setNewHost` is also potentially susceptible to a desync (db operation is between server-local room update and broadcasting updated room state to clients). Similar issue goes for `ChangeSettings`.

peppy:
I can see a slight issue in `setNewHost`, but not sure I see one in `ChangeSettings`, can you elaborate there?

bdach:
that's my bad, missed the rollback there. that was even fixed recently... looks fine.

osu.Server.Spectator/Hubs/MultiplayerHub.cs
```diff
@@ -430,10 +440,11 @@ protected override async Task CleanUpState(MultiplayerClientState state)
         private async Task setNewHost(MultiplayerRoom room, MultiplayerRoomUser newHost)
         {
             room.Host = newHost;
+            await Clients.Group(GetGroupId(room.RoomID)).HostChanged(newHost.UserID);
 
-            await UpdateDatabaseHost(room);
 
-            await Clients.Group(GetGroupId(room.RoomID)).HostChanged(newHost.UserID);
+            // don't care too much about failures here (is only for display), so run after local state is updated.
+            await UpdateDatabaseHost(room);
```
bdach, line 447: still not 100% safe to do that i don't think. if the usage in `leaveRoom()` throws, it won't notify clients that the user has left.

probably safest to mirror `UpdateDatabaseParticipants()` for now.

bdach (approved):

PR #33, opened 2021-01-04
Fix user's joined state not being updated as part of room transaction

peppy:
This shouldn't cause any issues with the current scope of this server,
but may in the future (for instance, if a user is kicked from a room).

### DIFF
--- a/osu.Server.Spectator/Hubs/MultiplayerHub.cs
+++ b/osu.Server.Spectator/Hubs/MultiplayerHub.cs
@@ -80,12 +80,12 @@ namespace osu.Server.Spectator.Hubs
                     room.Users.Add(roomUser);
 
                     await UpdateDatabaseParticipants(room);
+
+                    await Clients.Group(GetGroupId(roomId)).UserJoined(roomUser);
+                    await Groups.AddToGroupAsync(Context.ConnectionId, GetGroupId(roomId));
+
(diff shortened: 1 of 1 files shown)


bdach:
I'm not sure I can entirely see where the concern about kicking is coming from (any kick operation implementation would surely have to take the room lock - unless it's the item creation itself that's the potential issue there?).

I can, however, agree with this change for different reasons (just looks to be safer overall, and should reduce [hold-and-wait](https://en.wikipedia.org/wiki/Deadlock#Necessary_conditions) behaviour in other methods that take both of these locks - I'm not saying that I'm necessarily worried about any deadlocks in this code, but this should in theory decrease time spent by other operations on waiting for both locks, as the previous implementation had a theoretical window where it released just one of two).

On a side note, we have to be extra careful about the order of taking locks in the multiplayer hub - currently it looks good as everything takes user first and then room, but we must be vigilant about new operations such that they don't flip that order.

bdach (approved):

peppy:
The concern would be that when kicking, locks may not be obtained on the user object (as the way we currently take locks would result in potential deadlocks without careful consideration - you'd need to take out a lock on the target user to kick, then the room inside of that).

With the rewritten code, a lock on the room would be enough to perform an operation on a foreign user in the same room, as long as the operation is related to the room that user is joined to.

PR #34, opened 2021-01-04
Add more logging to attempt to track down remaining issues

peppy:
- [x] Depends on https://github.com/ppy/osu-server-spectator/pull/33 due to structural changes

### DIFF
--- a/osu.Server.Spectator/Hubs/MultiplayerHub.cs
+++ b/osu.Server.Spectator/Hubs/MultiplayerHub.cs
@@ -50,7 +50,7 @@ namespace osu.Server.Spectator.Hubs
                 if (userUsage.Item != null)
                 {
                     // if the user already has a state, it means they are already in a room and can't join another without first leaving.
-                    throw new InvalidStateException("Can't join a room when already in another room.");
+                    throw new InvalidStateException($"Can't join room {roomId} when already in room {userUsage.Item.CurrentRoomID}.");
                 }
 
                 // add the user to the room.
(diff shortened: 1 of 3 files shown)


osu.Server.Spectator/Hubs/MultiplayerHub.cs
```diff
@@ -50,7 +50,7 @@ public async Task<MultiplayerRoom> JoinRoom(long roomId)
                 if (userUsage.Item != null)
                 {
                     // if the user already has a state, it means they are already in a room and can't join another without first leaving.
-                    throw new InvalidStateException("Can't join a room when already in another room.");
+                    throw new InvalidStateException($"Can't join room {roomId} when already in room {userUsage.Item.CurrentRoomID}.");
```
bdach, line 53: Sure we don't want a `LogAndThrow()` sort of thing for these? These won't log the current user context ID, or will they?

Some throws already include user IDs within exception messages - might supersede those with that method if it is added.

Also as a side thought, we're exposing these to the users (as any errors in multi operations are just marshaled as plain server-side exception through the wire), so we might want something that's more structured, like a user-facing message along with a context object that gets logged to console on top of that. Maybe going too far with that one, I don't know. Not sure how permanent this logging is intended to be.

osu.Server.Spectator/Hubs/StatefulUserHub.cs
```diff
@@ -33,15 +34,17 @@ protected StatefulUserHub(IDistributedCache cache)
...
 
             try
             {
                 // if a previous connection is still present for the current user, we need to clean it up.
                 await cleanUpState(false);
             }
-            catch
+            catch (Exception e)
             {
+                Log("State cleanup failed");
```
bdach, line 46: Did this mean to log the exception? The `e` local is not used.

osu.Server.Spectator/Hubs/MultiplayerHub.cs
peppy: Invalid state are client-side issues, so this should probably just be reverted. As you say, this should be a user parsable for now.

osu.Server.Spectator/Hubs/StatefulUserHub.cs
peppy: It's already logged via the `throw` below; i've removed the unaccessed variable.

bdach (approved):

PR #36, opened 2021-01-06
Fix state management mishaps in old connection cleanup

peppy:
Due to yet another oversight, there was the potential for an old connection to unsafely destroy a user context (without running any clean up code).

The oversight here is that the early `return` will still run the `finally` code, which includes not just a `dispose` but also a `destroy`:

https://github.com/ppy/osu-server-spectator/blob/351e32ce8336a203445d82a5e53b498f21c46364/osu.Server.Spectator/Hubs/StatefulUserHub.cs#L89-L90

for the same user:
- Connection 1 connects and creates a state
- Connection 2 connects, cleans up previous state
- Connection 2 create a state
- Connection 1 disconnects, and DESTROYS THE NEW STATE

Added test coverage for the above and other connection cleanup expectations.


bdach (approved):
sure

PR #38, opened 2021-01-06
Improve component namespacing

bdach:
I know the entity change was planned before, but I also took the opportunity to clean up database too.

The `{-> Osu}Database` rename was unfortunate but necessary to prevent namespace/class name clashes. Not sure about that name, but it was the sanest thing I could come up with without going overly enterprise:tm: (with like `DatabaseConnectionFactory` or something). Such a rename might be coming up anyway, as I'm figuring out how to split out database access to be more testable.


bdach:
For reference, the end goal of these changes that I was aiming towards can be seen here: https://github.com/bdach/osu-server-spectator/compare/namespacing...bdach:database-testability-wip

Buyer beware: I have not ran the above code, just adjusted test cases to pass. I believe it should have a chance of working, but it might also just die at runtime if I forgot about something (or if stackoverflow lied to me that I can do something like `ConfigureJwtBearerOptions` at all and have the db factory DI'd in there). Doesn't help that to test this I'd have to spend a whiiiiiiile to spin up a test environment.

osu.Server.Spectator/Database/OsuDatabase.cs
```diff
@@ -4,9 +4,9 @@
 using System;
 using MySqlConnector;
 
-namespace osu.Server.Spectator
+namespace osu.Server.Spectator.Database
 {
-    public class Database
+    public class OsuDatabase
```
peppy, line 9: `DatabaseHelper` maybe? or just `DB` (similar to `RNG`) since this is intended to be a concise call.

bdach: I've gone with `DB`. I have a strong dislike for class names with `Helper` in them, always feels like open season for a random grab-bag of unrelated stuff.

peppy (approved):

PR #39, opened 2021-01-09
Decouple multiplayer logic from database accesses

bdach:
Continuation of work I was hinting at in #38, addressing one of my largest issues with the current server code.

What this pull helps achieve:

* Isolation of multiplayer logic and data access - [no need to duplicate error conditions in test code anymore](https://github.com/ppy/osu-server-spectator/pull/35#discussion_r552100339).
* Ability to mock data access and inject any special cases or errors for testing purposes, paving the way for improved edge case test coverage.
* Aesthetically, crops out all the SQL from the hub and makes the file smaller.
* As a byproduct the authentication setup also gets its own separate class which is also sort of nice (the pretext for that is that I had to be able to DI the DB factory in somehow).

Unfortunately the diff is quite large because of the sheer amount of code moved around. I'm not sure if there's any reasonable way to split this.

This is probably more of an RFC rather than 100% ready to go code, although I did test this today alongside with a web instance and a lazer instance. Did not spot anything out of the ordinary (at least not on the server side, more about that in a bit over at lazer). Willing to listen to feedback on structure and naming for sure.


peppy (approved):
Looks good!

PR #40, opened 2021-01-11
👷 Actions: Add Docker build and push

ThePooN:
Builds a Docker image and push to Docker Hub every time a tag is pushed.


peppy (approved):
I guess this will do as a starting point. We may want to rename the image / project in the future though.

PR #42, opened 2021-01-19
Add flow to automatically destroy state if item is null after usage

### LINKED ISSUE #41: Leaving a state with a null item inhibits cleanup
There are various cases where we retrieve a user state but don't set the state against it due to an error or otherwise early exit:

https://github.com/ppy/osu-server-spectator/blob/5c8cd28513559162c1b7878ac3cb32598bacaa1c/osu.Server.Spectator/Hubs/MultiplayerHub.cs#L56-L60

https://github.com/ppy/osu-server-spectator/blob/5c8cd28513559162c1b7878ac3cb32598bacaa1c/osu.Server.Spectator/Hubs/MultiplayerHub.cs#L203-L204

https://github.com/ppy/osu-server-spectator/blob/5c8cd28513559162c1b7878ac3cb32598bacaa1c/osu.Server.Spectator/Hubs/MultiplayerHub.cs#L245-L246

This seems fine at first glance, but looking at state clean-up code, the user state will not be removed completely if the item is null:

https://github.com/ppy/osu-server-spectator/blob/5c8cd28513559162c1b7878ac3cb32598bacaa1c/osu.Server.Spectator/Hubs/StatefulUserHub.cs#L85

This leads to states remaining in a "stuck" state until the user next connects. A non-disconnect will be able to reuse the state on next connect, so this isn't a critical issue, but one we should still resolve.

Proposals:

- Destroy a state usage on dispose if the item is null
- Somehow enforce non-null on the item in a different way
- Have StatefulUserHub track connectionId outside of `ClientState`

First option being the simplest, if we are happy with that direction.

peppy:
Closes #41 in the simplest way possible. Need feedback on visibility/readability, but seems to work as expected.

### DIFF
--- a/osu.Server.Spectator.Tests/MultiplayerFlowTests.cs
+++ b/osu.Server.Spectator.Tests/MultiplayerFlowTests.cs
@@ -145,6 +145,35 @@ namespace osu.Server.Spectator.Tests
         #region Joining and leaving
 
         [Fact]
+        public async Task UserCantJoinWhenRestricted()
+        {
+            mockDatabase.Setup(db => db.IsUserRestrictedAsync(It.IsAny<int>())).ReturnsAsync(true);
+
+            // ensure the same user can't join a room if already in a room.
+            await Assert.ThrowsAsync<InvalidStateException>(() => hub.JoinRoom(room_id));
+
+            // ensure no state was left behind.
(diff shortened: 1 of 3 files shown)


bdach (approved):

bdach:
Seems perfectly innocuous to me.

PR #43, opened 2021-01-19
Exclude code coverage from unexpected cases

frenzibyte:
This excludes code coverage from cases that are unexpected to happen, and cannot really be covered in tests.

I was hoping this could be achieved by [reusing the helper method removed](https://github.com/ppy/osu-server-spectator/pull/28), but the cases are still included in code coverage, and I believe that's because the condition to trigger the exception is still inside code coverage, not in the excluded helper method.

Therefore I've extracted each case to its own excluded method:
https://github.com/ppy/osu-server-spectator/blob/56c5f2d094751542e9df2ae9761814853c98cd19/osu.Server.Spectator/Hubs/MultiplayerHub.cs#L575-L582 https://github.com/ppy/osu-server-spectator/blob/56c5f2d094751542e9df2ae9761814853c98cd19/osu.Server.Spectator/Hubs/MultiplayerHub.cs#L584-L592

### DIFF
--- a/osu.Server.Spectator/Hubs/MultiplayerHub.cs
+++ b/osu.Server.Spectator/Hubs/MultiplayerHub.cs
@@ -238,15 +238,8 @@ namespace osu.Server.Spectator.Hubs
             using (var userUsage = await GetOrCreateLocalUserState())
             using (var roomUsage = await getLocalUserRoom(userUsage.Item))
             {
-                var room = roomUsage.Item;
-
-                if (room == null)
-                    throw new InvalidOperationException("Attempted to operate on a null room");
-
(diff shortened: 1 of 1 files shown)


bdach:
feels so wrong on multiple levels.

* doing this for absolutely 100% coverage feels like a pointless exercise
* not even true that this cannot be covered, it's just impractical (would require more mocking)
* actively harms readability (how does `getLocalUserRoom` differ from `getRoomOrThrow`?)

bdach:
also resolving #41 will probably leave this (and the existing throws) null and void

frenzibyte:
Closing for now as incorrect and likely useless.

PR #46, opened 2021-01-22
Clone the full MultiplayerRoom before returning to the user to avoid collection enumeration failures

peppy:
As seen in server logs, there are cases where `JoinRoom` operations fail to serialise the message to the end user due to collection modification during enumeration:

```
Microsoft.AspNetCore.SignalR.HubConnectionContext[6]
      Failed writing message. Aborting connection.
      System.InvalidOperationException: Collection was modified;
enumeration operation may not execute.
         at System.Collections.Generic.List`1.Enumerator.MoveNextRare()
```

After this occurs, things seem to fall apart in terms of server consistency. I'm not quite sure why this is the case yet (since the serialisation failure should occur after the lock has been returned, and should not affect any transactional server operation) but figure it is best to fix this as a first step.

The "Aborting connection" part is a touch suspicious - maybe it is bypassing the cleanup logic in cases like this. Will require some further investigation.

Unfortunately, this makes the tests are bit uglier than before.

As a side note, I tried using AutoMapper for the cloning of the room, but ran into issues with reference types somewhere in the hierarchy.  While the json method is horribly inefficient, it is a very simple way to ensure serialization consistency (since we're using the same method for the SignalR level serialization). Also, as it's only happening in the `JoinRoom` method (which is quite rare in comparison to other operations) this should do fine for now.

### DIFF
--- a/osu.Server.Spectator.Tests/MultiplayerFlowTests.cs
+++ b/osu.Server.Spectator.Tests/MultiplayerFlowTests.cs
@@ -125,23 +125,24 @@ namespace osu.Server.Spectator.Tests
         public async Task HostTransfer()
         {
             setUserContext(mockContextUser1);
-            var room = await hub.JoinRoom(room_id);
+            await hub.JoinRoom(room_id);
 
             setUserContext(mockContextUser2);
-            Assert.Equal(room, await hub.JoinRoom(room_id));
+            await hub.JoinRoom(room_id);
 
             setUserContext(mockContextUser1);
             await hub.TransferHost(user_id_2);
 
             mockReceiver.Verify(r => r.HostChanged(user_id_2), Times.Once);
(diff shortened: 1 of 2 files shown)


smoogipoo:
Just so I understand this...

`EntityStore` ensures that one thread can access a `MultiplayerRoom` at one time, but the collection being modified here and causing the exception is the list of users _inside_ the `MultiplayerRoom`?

bdach:
That seems correct. [That was something I had considered once](https://github.com/ppy/osu-server-spectator/pull/25#discussion_r550748213), but wasn't sure it was an actual concern.

peppy:
The issue is the serialisation happens outside the locked portion (see `JoinRoom`'s return type).

smoogipoo (approved):

PR #48, opened 2021-01-22
Use .NET 5.0 in Docker build

ThePooN:


peppy:
I'd say this needs to be applied in more places than just this project, most likely.

peppy (approved):

PR #49, opened 2021-01-26
Add MessagePack support

peppy:
This adds MessagePack as a supported protocol, in addition to existing Newtonsoft.JSON.

- [x] Depends on https://github.com/ppy/osu/pull/11607 + game dll bump


bdach:
Diff looks good (but not approving since the eventual game bump that will have to happen so that this can be merged will dismiss an approve anyway).

smoogipoo (approved):

PR #50, opened 2021-02-01
Add support for optional mods (aka freemod)

smoogipoo:
Prereqs:
- [x] https://github.com/ppy/osu/pull/11641
- [x] New nuget package.


peppy (approved):
Implementation looks fine. Will need to be updated with `UserMods` -> `Mods` and `Mods` -> `RequiredMods` renames.

PR #51, opened 2021-02-05
Update projects to target SignalR 5.0

peppy:


smoogipoo (approved):

peppy:
I've combined this with the freemods PR to avoid some weird version conflicts (test projects aren't explicitly referencing SignalR so required an osu.Game update to work... but osu.Game update requires code changes in freemods branch).

PR #54, opened 2021-02-17
Fix ability to update room settings with custom rulesets

frenzibyte:
Can be reproduced by creating a room regularly in multiplayer, then updating room settings to select a beatmap of a custom ruleset.

**NOTE:** I haven't tested this yet, but it looks to be theoretically working given that the other guards there exist and are working properly.

Wasn't initially sure about just throwing in the number `3` as the maximum legacy ruleset ID (despite of remaining constant), so I went ahead and used `ILegacyRuleset.MAX_LEGACY_RULESET_ID` instead, feels nicer that way.

Also guarded against negative ruleset ID as that sounds like a possibility to happen (from the type of `ruleset_id` in `multiplayer_playlist_item` being signed short).

---

Added 100% code coverage of the whole `updateDatabaseSettings` method while at it.

---

EDIT: This could potentially be the reason why rooms with playlist items of custom rulesets were being received in multiplayer lounge somewhat, though I'm not sure how that could happen given that the user must set the filter to the right custom ruleset ID /shrug.

### DIFF
--- a/osu.Server.Spectator.Tests/MultiplayerFlowTests.cs
+++ b/osu.Server.Spectator.Tests/MultiplayerFlowTests.cs
@@ -770,7 +770,7 @@ namespace osu.Server.Spectator.Tests
         [Fact]
         public async Task ChangingSettingsToNonExistentBeatmapThrows()
         {
-            mockDatabase.Setup(d => d.GetBeatmapChecksumAsync(3333)).ReturnsAsync((string)null!);
+            mockDatabase.Setup(d => d.GetBeatmapChecksumAsync(3333)).ReturnsAsync((string?)null);
 
             MultiplayerRoomSettings testSettings = new MultiplayerRoomSettings
             {
--- a/osu.Server.Spectator.Tests/MultiplayerFlowTests.cs
+++ b/osu.Server.Spectator.Tests/MultiplayerFlowTests.cs
@@ -15,6 +15,7 @@ using Moq;
 using osu.Game.Online;
 using osu.Game.Online.Multiplayer;
 using osu.Game.Online.Rooms;
+using osu.Game.Rulesets;
 using osu.Server.Spectator.Database;
 using osu.Server.Spectator.Database.Models;
 using osu.Server.Spectator.Entities;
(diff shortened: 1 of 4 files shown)


osu.Server.Spectator.Tests/MultiplayerFlowTests.cs
```diff
@@ -766,6 +767,52 @@ public async Task RoomSettingsUpdateNotifiesOtherUsers()
             mockReceiver.Verify(r => r.SettingsChanged(testSettings), Times.Once);
         }
 
+        [Fact]
+        public async Task ChangingSettingsToNonExistentBeatmapThrows()
+        {
+            mockDatabase.Setup(d => d.GetBeatmapChecksumAsync(3333)).ReturnsAsync((string)null!);
```
bdach, line 773: While the database does seem to list the `checksum` column as nullable, that would mean that the return type of `GetBeatmapChecksumAsync()` should *actually* be `Task<string?>`. Because note that `DatabaseAccess` uses `Single()`, not `SingleOrDefault()`. So a null here would have to physically reside *in the database*.

I'm not sure if there is a case of a null checksum in production, but either way - either this test should be removed due to testing an unrealistic scenario, or the types should be adjusted to avoid this `null!` nonsense.

This is all setting aside the fact of adding tests unrelated to the actual goal of this pull.

bdach (dismissed):

osu.Server.Spectator.Tests/MultiplayerFlowTests.cs
frenzibyte: > While the database does seem to list the checksum column as nullable, that would mean that the return type of GetBeatmapChecksumAsync() should actually be Task<string?>. Note that DatabaseAccess uses Single(), not SingleOrDefault(). So a null here would have to physically reside in the database.

Yeah, I was a bit hesitant about it, but got too busy to look through it, my bad though. I've went ahead and made `GetBeatmapChecksumAsync()` return nullable instead. 

Not sure about straight up changing it to `SingleOrDefault()` though, @peppy can you confirm whether you intended making `GetBeatmapChecksumAsync()` use `QuerySingleAsync()` instead of `QuerySingleOrDefaultAsync()`, sounds like it should've been the latter?

> This is all setting aside the fact of adding tests unrelated to the actual goal of this pull.

I have added these tests while I was at this, don't see any complexity to move it into its own PR. (not to mention that I wasn't really comfortable at all with adding just one only test for the spotted edge case)

osu.Server.Spectator/Hubs/MultiplayerHub.cs
```diff
@@ -170,6 +172,8 @@ private async Task<MultiplayerRoom> retrieveRoom(long roomId)
                 var playlistItem = await db.GetCurrentPlaylistItemAsync(roomId);
                 var beatmapChecksum = await db.GetBeatmapChecksumAsync(playlistItem.beatmap_id);
 
+                Debug.Assert(beatmapChecksum != null);
```
peppy, line 175: Are you sure about using an `Assert` here rather than throwing an exception? It's going to allow deployed builds to get in a potentially bad state (as it will be ignored). See below where it's throwing an exception in the same case.

Note that it should be considered a server-side exception at this point (so not an `InvalidStateException`, since the user doesn't need to know what went wrong).

osu.Server.Spectator/Hubs/MultiplayerHub.cs
```diff
@@ -417,11 +421,14 @@ private async Task clearDatabaseScores(MultiplayerRoom room)
 
         private async Task updateDatabaseSettings(MultiplayerRoom room)
         {
+            if (room.Settings.RulesetID < 0 || room.Settings.RulesetID > ILegacyRuleset.MAX_LEGACY_RULESET_ID)
+                throw new InvalidStateException("Attempted to select a beatmap of a custom ruleset.");
```
peppy, line 425:
```suggestion
                throw new InvalidStateException("Attempted to select an unsupported ruleset.");
```

It has nothing to do with the beatmap (the setting is at a room level and may be a convert).

peppy (changes requested):

peppy (approved):

PR #55, opened 2021-02-17
Insert new items into the playlist with each match start

smoogipoo:
Prereqs:
- [x] https://github.com/ppy/osu-web/pull/7223
- [x] https://github.com/ppy/osu/pull/11808

On match end:
1. The current playlist item is duplicated.
2. `expired = 1` is set on the old playlist item.
3. The duplicate item is propagated to clients via `SettingsChanged`.

Here's how it looks at a database level: https://drive.google.com/file/d/1xDLePId0IIVdqU3Qce437A9CRYSH0v95/view?usp=sharing

Compatibility between old and new clients:
- Old clients can join rooms from the new clients that are on their first playlist item. Any more than that and they die on a `Playlist.SingleOrDefault()` call somewhere.
- Old clients can play the first playlist item, but then die on an added osu-web exception (https://github.com/ppy/osu-web/blob/d92c83fb80467796da3b892647fa426842f868e6/app/Http/Controllers/Multiplayer/Rooms/Playlist/ScoresController.php#L196) otherwise they'd submit to an older playlist_item_id.
- Old clients can't change beatmaps - they don't populate `PlaylistItemId` and die on an added exception (https://github.com/ppy/osu-server-spectator/compare/master...smoogipoo:multiplayer-no-playlist-mangling?expand=1#diff-90658690a1d84ef42d78f75a985211e9140da7fdadbcb3b5562eee9c5d894fd7R437-R438)

### DIFF
--- a/osu.Server.Spectator/Database/DatabaseAccess.cs
+++ b/osu.Server.Spectator/Database/DatabaseAccess.cs
@@ -160,6 +160,7 @@ namespace osu.Server.Spectator.Database
         {
             var currentItem = await GetCurrentPlaylistItemAsync(room.RoomID);
 
+            await connection.ExecuteAsync("UPDATE multiplayer_playlist_items SET played=1 WHERE id=@id", currentItem);
             await connection.ExecuteAsync(
                 "INSERT INTO multiplayer_playlist_items (room_id, beatmap_id, ruleset_id, allowed_mods, required_mods, created_at, updated_at)"
                 + " VALUES (@room_id, @beatmap_id, @ruleset_id, @allowed_mods, @required_mods, NOW(), NOW())",
--- a/osu.Server.Spectator/Database/DatabaseAccess.cs
+++ b/osu.Server.Spectator/Database/DatabaseAccess.cs
@@ -171,9 +171,9 @@ namespace osu.Server.Spectator.Database
 
         public async Task EndMatchAsync(MultiplayerRoom room)
         {
-            // Remove all non-played items from the playlist. For now this is only the current item.
+            // Remove all non-expired items from the playlist as they have no scores.
             var currentItem = await GetCurrentPlaylistItemAsync(room.RoomID);
-            await connection.ExecuteAsync("DELETE FROM multiplayer_playlist_items WHERE id=@id", currentItem);
+            await connection.ExecuteAsync("DELETE FROM multiplayer_playlist_items WHERE expired = 0");
 
             // Close the room.
             await connection.ExecuteAsync("UPDATE multiplayer_rooms SET ends_at = NOW() WHERE id = @RoomID", new
--- a/osu.Server.Spectator/Database/DatabaseAccess.cs
+++ b/osu.Server.Spectator/Database/DatabaseAccess.cs
@@ -156,17 +156,28 @@ namespace osu.Server.Spectator.Database
             }
         }
 
-        public async Task<long> CommitPlaylistItem(MultiplayerRoom room)
+        public async Task<long> CreatePlaylistItemAsync(multiplayer_playlist_item item)
         {
-            var currentItem = await GetCurrentPlaylistItemAsync(room.RoomID);
-
-            await connection.ExecuteAsync("UPDATE multiplayer_playlist_items SET expired = 1, updated_at = NOW() WHERE id=@id", currentItem);
             await connection.ExecuteAsync(
                 "INSERT INTO multiplayer_playlist_items (room_id, beatmap_id, ruleset_id, allowed_mods, required_mods, created_at, updated_at)"
                 + " VALUES (@room_id, @beatmap_id, @ruleset_id, @allowed_mods, @required_mods, NOW(), NOW())",
-                currentItem);
+                item);
 
-            return (await GetCurrentPlaylistItemAsync(room.RoomID)).id;
+            var lastItem = await connection.QuerySingleAsync<multiplayer_playlist_item>(
+                "SELECT * FROM multiplayer_playlist_items WHERE id = (SELECT MAX(id) FROM multiplayer_playlist_items WHERE room_id = @RoomId)", new
+                {
+                    RoomID = item.room_id
+                });
+
+            return lastItem.id;
+        }
+
+        public async Task ExpirePlaylistItemAsync(long playlistItemId)
+        {
+            await connection.ExecuteAsync("UPDATE multiplayer_playlist_items SET expired = 1, updated_at = NOW() WHERE id = @PlaylistItemId", new
+            {
+                PlaylistItemId = playlistItemId
+            });
         }
 
         public async Task EndMatchAsync(MultiplayerRoom room)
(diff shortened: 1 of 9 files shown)


osu.Server.Spectator/Database/DatabaseAccess.cs
```diff
@@ -59,7 +59,8 @@ public Task<multiplayer_room> GetRoomAsync(long roomId)
 
         public Task<multiplayer_playlist_item> GetCurrentPlaylistItemAsync(long roomId)
         {
-            return connection.QuerySingleAsync<multiplayer_playlist_item>("SELECT * FROM multiplayer_playlist_items WHERE room_id = @RoomID", new
+            // Todo: Add ordering.
+            return connection.QueryFirstAsync<multiplayer_playlist_item>("SELECT * FROM multiplayer_playlist_items WHERE room_id = @RoomId AND expired = 0", new
```
bdach, line 63: Not sure about the `Single` -> `First` change. Seems like keeping `Single` would help in checking the invariant that there can be only one non-expired playlist item at a time.

osu.Server.Spectator/Database/DatabaseAccess.cs
```diff
@@ -154,22 +156,40 @@ public async Task UpdateRoomParticipantsAsync(MultiplayerRoom room)
             }
         }
 
-        public async Task ClearRoomScoresAsync(MultiplayerRoom room)
+        public async Task<multiplayer_playlist_item> GetPlaylistItemFromRoomAsync(long roomId, long playlistItemId)
```
bdach, line 159: If `QueryFirstOrDefault`, then
```suggestion
        public async Task<multiplayer_playlist_item?> GetPlaylistItemFromRoomAsync(long roomId, long playlistItemId)
```

osu.Server.Spectator/Database/DatabaseAccess.cs
```diff
@@ -154,22 +156,40 @@ public async Task UpdateRoomParticipantsAsync(MultiplayerRoom room)
...
-            return connection.ExecuteAsync("UPDATE multiplayer_rooms SET ends_at = NOW() WHERE id = @RoomID", new
+            await connection.ExecuteAsync("UPDATE multiplayer_playlist_items SET expired = 1, updated_at = NOW() WHERE id = @PlaylistItemId", new
+            {
+                PlaylistItemId = playlistItemId
+            });
+        }
+
+        public async Task EndMatchAsync(MultiplayerRoom room)
+        {
+            // Remove all non-expired items from the playlist as they have no scores.
```
bdach, line 188: I'm not sure that is correct? Playlist items are expired by `selectNextPlaylistItem()`, which in turn fires on the room transitioning to `Playing` state. I imagine this could therefore discard the scores associated with the last playlist item before the room closes upon all users leaving it.

osu.Server.Spectator/Database/DatabaseAccess.cs
```diff
@@ -154,22 +156,40 @@ public async Task UpdateRoomParticipantsAsync(MultiplayerRoom room)
...
+            await connection.ExecuteAsync("DELETE FROM multiplayer_playlist_items WHERE expired = 0");
```
bdach, line 189: This delete query does not filter by room ID, therefore accidentally nuking all non-expired playlist items globally across all rooms.

bdach (changes requested):

osu.Server.Spectator/Database/DatabaseAccess.cs
smoogipoo: This is not going to be the case in the future. I envisioned two extra modes for the multiplayer server:
1. Host rotate - players will be able to queue up maps and the server will automatically pick them based on something like
`SELECT * FROM multiplayer_playlist_items WHERE room_id = @RoomId AND expired = 0 ORDER BY playlist_order ASC` instead of this line here.
2. Tournament - the server sets up a playlist where the user can change to any given item from that playlist but not add additional items to it. Again this line will change.

More refactoring is required before that path though.

osu.Server.Spectator/Database/DatabaseAccess.cs
smoogipoo: `expired = 1` is set from `Playing` -> `FinishedPlay`, but I think you're right in two cases:
1. Players start play, then all players leave. Technically score submission hasn't occurred yet but the `multiplayer_scores` table has been populated.
2. Players start play, some submit scores, then _all_ players leave. I'm not sure in this case if the room will ever transition to the `FinishedPlay` state before it's closed...

Need more opinions here... Should this even be a thing? The only reason I have this is for visualising the progression in the room - the lounge would otherwise show 5 maps where only 4 have been played, and similarly for something like match history (mp links).

smoogipoo: What I could do is set the expired state on match start, and keep the select-next part happening on match end. The question I have then is how to handle the client-side visual part of this, where the API has no concept of the "current" playlist item and hides all items with `expired = 1`.

osu.Server.Spectator/Database/DatabaseAccess.cs
peppy: This is indeed the correct path forward. We want to be able to queue future playlist items, either as the host or via other systems.

osu.Server.Spectator/Database/DatabaseAccess.cs
peppy: The two cases you mention are probably best handled in a separate PR/change. The multiplayer server should be responsible for deciding if a match was "completed" or not, based on whether any single score was submitted, rather than requiring all scores to submit (which I presume from the above description is what is happening right now?).

Basically there would be some method which is run on standard flow (finished play) but also on room close, which decides what to do about any ongoing play, marking it as finished etc.

From a database perspective, marking as `expired` after the play ends sounds correct.

smoogipoo: > rather than requiring all scores to submit

Yeah, that's correct. Right now it'll nuke the playlist item if the room never transitions to `FinishedPlay`, which can happen via the two cases I mentioned. It actually won't nuke scores since deletes don't cascade, but it still breaks the DB relations (score can point to a playlist item that has been deleted).

So the question is then, should the server do this deletion at all?

peppy: Only if no scores were submitted. Or never works fine, too. Since the play did start.

smoogipoo: In that case I'd change this to `DELETE FROM multiplayer_playlist_items p WHERE p.room_id = @room_id AND p.expired = 0 AND (SELECT COUNT(*) FROM multiplayer_scores s WHERE s.playlist_item_id = p.id) = 0`. Sound good?

peppy: The logic sounds fine, but I'm not sure about how efficient this execution will be. I think it will do for now, but down the line we will probably want to decide which items to delete locally (the server should have the knowledge I think?) rather than doing a nested query.

peppy (approved):
Seems fine, pending deployment discussion.

PR #56, opened 2021-03-25
Fix room state not being updated if a user is disconnected while playing

peppy:
Closes https://github.com/ppy/osu/issues/12010.


bdach (approved):
seems good :+1:

PR #62, opened 2021-06-22
Add mod validation for `RequiredMods` and `AllowedMods`

peppy:
This adds the same level of validation to required/allowed mods (at the point of a host changing room settings) that the previous PR added for `UserMods`. Note that as a result, some of the previous tests have been updated as they were throwing newly added exception cases.

- [x] Depends on (and builds on) https://github.com/ppy/osu-server-spectator/pull/60

### DIFF
--- a/osu.Server.Spectator/Hubs/MultiplayerHub.cs
+++ b/osu.Server.Spectator/Hubs/MultiplayerHub.cs
@@ -459,13 +459,13 @@ namespace osu.Server.Spectator.Hubs
             }
 
             if (!ModUtils.CheckCompatibleSet(requiredMods, out var invalid))
-                throw new InvalidStateException($"Invalid combination of required mods: {invalid.Select(m => m.Acronym)}");
+                throw new InvalidStateException($"Invalid combination of required mods: {string.Join(',', invalid.Select(m => m.Acronym))}");
 
             // check aggregate combinations with each allowed mod individually.
             foreach (var allowedMod in allowedMods)
             {
                 if (!ModUtils.CheckCompatibleSet(requiredMods.Concat(new[] { allowedMod }), out invalid))
-                    throw new InvalidStateException($"Invalid combination of required mods: {invalid.Select(m => m.Acronym)}");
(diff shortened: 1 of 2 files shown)


osu.Server.Spectator/Hubs/MultiplayerHub.cs
```diff
@@ -438,6 +443,32 @@ private async Task changeUserMods(IEnumerable<APIMod> newMods, MultiplayerRoom r
...
+            }
+
+            if (!ModUtils.CheckCompatibleSet(requiredMods, out var invalid))
+                throw new InvalidStateException($"Invalid combination of required mods: {string.Join(',', invalid.Select(m => m.Acronym))}");
+
+            // check aggregate combinations with each allowed mod individually.
+            foreach (var allowedMod in allowedMods)
+            {
+                if (!ModUtils.CheckCompatibleSet(requiredMods.Concat(new[] { allowedMod }), out invalid))
+                    throw new InvalidStateException($"Invalid combination of required and allowed mods: {string.Join(',', invalid.Select(m => m.Acronym))}");
```
bdach, line 468: Just going to note that there are some game-side cases in which this may run counter to players' expectations. Consider someone wanting to have Sudden Death as a required mod and Perfect as a free mod, which technically kinda makes sense as Perfect is a stricter Sudden Death. With this change that is not permitted anymore. But I'm not sure whether it's a huge deal right now.

bdach (approved):
I've adjusted some exception messages to be better. Looks pretty good to me :+1: 

Just going to leave one stray remark, up to you to decide whether we care about it or not for now.

osu.Server.Spectator/Hubs/MultiplayerHub.cs
peppy: This is an interesting one for sure. I think we *do* want to allow such cases, but this should probably either be done in `CheckCompatibleSet` or via a new `ModUtils` method, so we keep this logic all in one place.

smoogipoo (approved):
lgtm

PR #63, opened 2021-07-01
Add support for `in_room` flag to avoid truncating `multiplayer_rooms_high` table

peppy:
Requires osu-web changes. Will add tests going forward; just PRing so osu-web peoples can see how things will work.

- [x] Depends on https://github.com/ppy/osu-web/pull/7819 for database changes

### DIFF
--- a/osu.Server.Spectator/Database/DatabaseAccess.cs
+++ b/osu.Server.Spectator/Database/DatabaseAccess.cs
@@ -127,13 +127,12 @@ namespace osu.Server.Spectator.Database
             {
                 using (var transaction = await connection.BeginTransactionAsync())
                 {
-                    await connection.ExecuteAsync("INSERT IGNORE INTO multiplayer_rooms_high (room_id, user_id, in_room) VALUES (@RoomID, @UserID, 1)", new
+                    // the user may have previously been in the room and set some scores, so need to update their presence if existing.
+                    await connection.ExecuteAsync("INSERT INTO multiplayer_rooms_high (room_id, user_id, in_room) VALUES (@RoomID, @UserID, 1) ON DUPLICATE KEY UPDATE in_room = 1", new
                     {
                         RoomID = room.RoomID,
                         UserID = user.UserID
                     }, transaction);
-
-                    // TODO: set in_room to 1 if the user is already in the table.
 
                     await connection.ExecuteAsync("UPDATE multiplayer_rooms SET participant_count = @Count WHERE id = @RoomID", new
                     {
--- a/osu.Server.Spectator/Database/DatabaseAccess.cs
+++ b/osu.Server.Spectator/Database/DatabaseAccess.cs
@@ -121,35 +121,56 @@ namespace osu.Server.Spectator.Database
             }
         }
 
-        public async Task UpdateRoomParticipantsAsync(MultiplayerRoom room)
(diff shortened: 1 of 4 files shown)


osu.Server.Spectator/Database/DatabaseAccess.cs
```diff
@@ -187,17 +210,19 @@ public async Task ExpirePlaylistItemAsync(long playlistItemId)
...
             // Remove all non-expired items from the playlist as they have no scores.
-            await connection.ExecuteAsync(
-                "DELETE FROM multiplayer_playlist_items p WHERE p.room_id = @RoomID AND p.expired = 0 AND (SELECT COUNT(*) FROM multiplayer_scores s WHERE s.playlist_item_id = p.id) = 0",
+            await connection.ExecuteAsync("DELETE FROM multiplayer_playlist_items p WHERE p.room_id = @RoomID AND p.expired = 0 AND (SELECT COUNT(*) FROM multiplayer_scores s WHERE s.playlist_item_id = p.id) = 0",
                 new
                 {
                     RoomID = room.RoomID
                 });
 
+            int totalUsers = connection.QuerySingle<int>("SELECT COUNT(*) FROM multiplayer_rooms_high WHERE room_id = @RoomID", new { RoomID = room.RoomID });
```
smoogipoo, line 219: I'm confused, what exactly is this doing? Setting `participant_count` to 1?

peppy: to the total count of users (that ever participated in the room).

smoogipoo: Ah, I see. I thought it was selecting `multiplayer_rooms`, not `multiplayer_rooms_high`.

smoogipoo (approved):

smoogipoo (approved):

osu.Server.Spectator/Database/DatabaseAccess.cs
```diff
@@ -119,35 +119,57 @@ public async Task UpdateRoomHostAsync(MultiplayerRoom room)
...
+                {
                     foreach (var u in room.Users)
                     {
-                        await connection.ExecuteAsync("INSERT INTO multiplayer_rooms_high (room_id, user_id) VALUES (@RoomID, @UserID)", new
+                        await connection.ExecuteAsync("UPDATE multiplayer_rooms_high SET in_room = 0 WHERE room_id = @RoomID AND user_id = @UserID", new
                         {
                             RoomID = room.RoomID,
                             UserID = u.UserID
                         }, transaction);
                     }
```
bdach, line 163: This can't be right, or can it? It's a loop over all users in the room, so this'll set `in_room = 0` on all of them...?

The following inspection is what tipped me off:

2021-07-03-131219_1231x155_scrot

bdach (changes requested):

osu.Server.Spectator/Database/DatabaseAccess.cs
peppy: Great catch. Not having tests on the sql itself is kind of bad, which is why I wasn't merging this until actually testing live

bdach (approved):

peppy:
Deployed and seems to be working as expected!

PR #70, opened 2021-07-25
Split multiplayer test cases to multiple different test classes

frenzibyte:
As the `MultiplayerFlowTests` grew bigger and bigger very quickly, it became hard to navigate between and write new test cases (at least for me).

Therefore, (and as been requested), I've separated all the test cases to multiple different test classes, starting off by taking each of one region and define a test class for it, then adjusted things a bit (moved single user flow and multi user flow tests to its own `MultiplayerFlowTest`).

---

One unfortunate thing I was forced to do is disabling Xunit test parallelization, as currently `StatefulUserHub` and `MultiplayerHub` itself read from a static entity store and reset on each test setup.

Parallelization only happens between different test classes, not between each test case of one class, that's why it wasn't noticed at all previously.

There are alternative ways to disable parallelization, such as using `[Collection(nameof(StatefulUserHub<IMultiplayerClient, MultiplayerClientState>)]` on `MultiplayerTest`, for example. (collection name doesn't matter, as long as it's the same name for all relevant test classes. more on parallelization [here](https://xunit.net/docs/running-tests-in-parallel.html))

But since `StatefulUserHub` itself is used in almost all test classes, it may be better just disabling parallelization for now, until we replace those statics with something more parallelization-friendly (cc @peppy).


bdach:
The way forward for removing the static fields would probably be to leverage dependency injection in the stateful hubs. At runtime the entity store can be registered as a singleton for DI to preserve the current behaviour.

I would probably prefer that to happen before this, but I'll see what everyone else has to say about that.

peppy:
I'm personally not too fussed about the parallel disabling as a temporary thing. I'm in the process of adding even more tests and wanted to get the tests into a good state as that single file was getting out of control.

Still need to do a full review on this change, though.

osu.Server.Spectator.Tests/Multiplayer/ModValidationTest.cs
```diff
@@ -0,0 +1,370 @@
...
+using Moq;
+using osu.Game.Online.API;
+using osu.Game.Online.Multiplayer;
+using osu.Game.Rulesets.Catch.Mods;
+using osu.Game.Rulesets.Osu.Mods;
+using Xunit;
+
+namespace osu.Server.Spectator.Tests.Multiplayer
+{
+    public class ModValidationTest : MultiplayerTest
```
peppy, line 17: Any reason for not using plurals for class names? As it stands it reads a bit weird. Not using plural would only seem correct if this was testing a class called `ModValidation`, I believe.

peppy (changes requested):

osu.Server.Spectator.Tests/Multiplayer/ModValidationTest.cs
frenzibyte: I haven't done that as it's not usual osu!lazer-side so I wasn't sure, will update them to your way.

frenzibyte:
> The way forward for removing the static fields would probably be to leverage dependency injection in the stateful hubs. At runtime the entity store can be registered as a singleton for DI to preserve the current behaviour.
> 
> I would probably prefer that to happen before this, but I'll see what everyone else has to say about that.

I'll give that an attempt now and PR it separately to not increase the complexity of reviewing and merging this PR.

peppy (approved):

PR #72, opened 2021-07-27
Use `osuweb` database user ID in run configurations

frenzibyte:
That's what an osu!web setup would use by default (see [`.env.example`](https://github.com/ppy/osu-web/blob/ff0dcdda9e421a21852146120dcb9ce49cff5d1d/.env.example#L17)).

The `launchSettings.json` would also be used when running using `dotnet run` which is a good thing for this.

### DIFF
--- a/osu.Server.Spectator/Properties/launchSettings.json
+++ b/osu.Server.Spectator/Properties/launchSettings.json
@@ -9,15 +9,6 @@
     }
   },
   "profiles": {
-    "IIS Express": {
-      "commandName": "IISExpress",
-      "launchBrowser": false,
-      "launchUrl": "ranklookup",
-      "environmentVariables": {
-        "DB_USER": "osuweb",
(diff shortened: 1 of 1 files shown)


osu.Server.Spectator/Properties/launchSettings.json
```diff
@@ -14,6 +14,7 @@
       "launchBrowser": false,
       "launchUrl": "ranklookup",
```
peppy, line 15: pretty sure this configuration is completely invalid (was copied from another project). might as well remove it.

frenzibyte: you mean removing the whole IIS express profile?

peppy: correct.

peppy (approved):

PR #73, opened 2021-07-27
Replace static entity stores with cached singletons

frenzibyte:
Replaces the static `ACTIVE_USERS` and `ACTIVE_ROOMS` entity stores with singletons to allow enabling test parallelization back.

Using the new introduced "hub entities" encapsulating classes cached as singletons for use in `StatefulUserHub` and `MultiplayerHub`, and are instantiated manually in tests as a better replacement of the previous static `Reset()` method.

All tests pass now with parallelization on, I've also done a sanity test using a local server and both multiplayer and spectator look to work properly.


osu.Server.Spectator.Tests/StatefulUserHubTest.cs
```diff
@@ -105,11 +104,12 @@ public async Task SameUserOldConnectionDoesntDestroyNewState()
 
         private class TestStatefulHub : StatefulUserHub<object, ClientState>
         {
-            public EntityStore<ClientState> UserStore => ACTIVE_STATES;
+            public EntityStore<ClientState> ActiveUsers { get; }
```
bdach, line 107: Not sure why this rename has been shoved into this diff, inflating it significantly. It seems mostly unnecessary.

osu.Server.Spectator/Entities/UserHubEntities.cs
```diff
@@ -0,0 +1,13 @@
...
+
+using osu.Server.Spectator.Hubs;
+
+namespace osu.Server.Spectator.Entities
+{
+    public abstract class UserHubEntities<TUserState>
+        where TUserState : ClientState
+    {
+        public readonly EntityStore<TUserState> ActiveUsers = new EntityStore<TUserState>();
+    }
```
bdach, line 12: Do these need to exist at all? The `EntityStores` can be registered directly, by specifying the constructed generic types (`EntityStore<SpectatorClientState>` and `EntityStore<MultiplayerClientState>`), no?

frenzibyte: Definitely, I just thought it would be a better idea to encapsulate things rather than caching each `EntityStore` per hub, though it may be an overcomplicated thing so not sure anymore.

osu.Server.Spectator.Tests/StatefulUserHubTest.cs
frenzibyte: It's just a name change in the end to match the entity store name, but I could revert that I guess.

osu.Server.Spectator/Entities/UserHubEntities.cs
peppy: The first example you give here is wrong. The multiplayer server shouldn't even have spectator states. I agree that these classes should not exist, though. Too many layers of abstraction.

frenzibyte: I wrote how it would look for both the `MultiplayerHub` constructor and the `SpectatorHub` constructor. I have just made these classes to shorten the dependencies through abstraction, but sure, can break that to direct `EntityStore`s if that's a better choice.

peppy (approved):

osu.Server.Spectator/Extensions/ServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,26 @@
...
+
+namespace osu.Server.Spectator.Extensions
+{
+    public static class ServiceCollectionExtensions
+    {
+        public static IServiceCollection AddHubEntities(this IServiceCollection serviceCollection)
+        {
+            return serviceCollection.AddSingleton<EntityStore<SpectatorClientState>>()
+                                    .AddSingleton<EntityStore<MultiplayerClientState>>()
+                                    .AddSingleton<EntityStore<Room>>();
```
frenzibyte, line 18: @peppy this seems to be the wrong room type to cache a store of, but you might beat me on fixing that though.

PR #79, opened 2021-08-11
Add kick support

peppy:
- [x] Depends on https://github.com/ppy/osu/pull/14225

### DIFF
--- a/osu.Server.Spectator.Tests/Multiplayer/HostManagementTests.cs
+++ b/osu.Server.Spectator.Tests/Multiplayer/HostManagementTests.cs
@@ -99,7 +99,7 @@ namespace osu.Server.Spectator.Tests.Multiplayer
             Receiver.Verify(r => r.UserKicked(It.IsAny<MultiplayerRoomUser>()), Times.Never);
 
             using (var room = await Rooms.GetForUse(ROOM_ID))
-                Assert.True(room.Item?.Users.Any(u => u.UserID != USER_ID_2));
+                Assert.True(room.Item?.Users.Any(u => u.UserID == USER_ID));
         }
 
         [Fact]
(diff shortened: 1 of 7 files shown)


osu.Server.Spectator.Tests/Multiplayer/HostManagementTests.cs
```diff
@@ -53,5 +56,67 @@ public async Task HostLeavingCausesHostTransfer()
...
+
+            SetUserContext(ContextUser2);
+            await Hub.JoinRoom(ROOM_ID);
+
+            await Assert.ThrowsAsync<NotHostException>(() => Hub.KickUser(USER_ID));
+
+            Receiver.Verify(r => r.UserKicked(It.IsAny<MultiplayerRoomUser>()), Times.Never);
+
+            using (var room = await Rooms.GetForUse(ROOM_ID))
+                Assert.True(room.Item?.Users.Any(u => u.UserID != USER_ID_2));
```
bdach, line 119: This one is definitely wrong, though. `ContextUser2` was doing the kicking while not being host, so there was never any danger that they would disappear (in any sane scenario, that is). Should presumably read

```suggestion
                Assert.True(room.Item?.Users.Any(u => u.UserID == USER_ID));
```

too.

bdach (changes requested):
looks good otherwise

osu.Server.Spectator.Tests/Multiplayer/HostManagementTests.cs
peppy: These both have the same meaning with two users right? Definitely could read better but I believe it’s logically testing the same thing.

bdach: Uhh right, those assertions are in fact equivalent iff there are 2 users. Not sure how it happened that I misread this... I think testing the desired outcome directly rather than its double negative may be safer/more understandable?

peppy: yeah, i'll change it for readability

bdach (approved):

PR #81, opened 2021-08-19
Don't throw exception when attempting to leave a room when not joined

peppy:
This is already deployed to production. It was used to make client-side event flow simpler in the case of kick operations.


frenzibyte (approved):

bdach:
See https://github.com/ppy/osu/pull/14291 for rationale.

Even if the client didn't explicitly try to leave the room on its own in the kick case, there are still plausible scenarios wherein a user is being kicked at the same time that they're trying to leave the room "normally" (not as a response to the kick), causing the exception either way.

smoogipoo (approved):

PR #83, opened 2021-11-11
Replace usage of `TypeNameHandling.All` with custom type converter

peppy:
- [x] Depends on https://github.com/ppy/osu/pull/15567.
- [x] osu.Game package bump (next release)


bdach:
Done some testing too, seems to work fine.

bdach (approved):

bdach:
I've pushed a game package bump here after yesterday's release.

There are more packages to bump overall (i.e. signalr has 6.0.0 versions out) but since they're majors I'm a bit skeeved out something with .NET 6 will break, so I left that for later.

PR #84, opened 2021-11-15
Add server-side support for multiplayer queue modes

smoogipoo:
Prereqs:
- [x] https://github.com/ppy/osu-web/pull/8336
- [x] https://github.com/ppy/osu/pull/15640

This PR only contains the host-only and free-for-all queueing modes.

Changes to tests are mostly to make the playlist interactions testable by hooking up the mock database methods, or to update to the new structure.


osu.Server.Spectator.Tests/Multiplayer/MultiplayerQueueTests.cs
```diff
@@ -0,0 +1,122 @@
...
+            {
+                var room = usage.Item;
+                Debug.Assert(room != null);
+
+                Assert.Equal(2, (await Database.Object.GetAllPlaylistItems(ROOM_ID)).Length);
+            }
+        }
+
+        [Fact]
+        public async Task JoiningRoomInvokesPlaylistItemAddedCallbackForEveryItem()
```
bdach, line 104: This test is failing for me locally. Adding an `await Hub.RequestAllPlaylistItems();` fixes, but honestly I'd more expect it to pass as written (as per [my client-side pull comment](https://github.com/ppy/osu/pull/15640#discussion_r749626537)).

bdach (commented):
Skimmed this one too. I'm tired and there's a lot of code, but it mostly passes my eye test. I haven't tried to majorly break it though, need to return with a fresh head for that.

osu.Server.Spectator/Hubs/ServerMultiplayerRoom.cs
```diff
@@ -27,13 +29,22 @@ public MatchTypeImplementation MatchTypeImplementation
...
             : base(roomId)
         {
             this.hubCallbacks = hubCallbacks;
 
             // just to ensure non-null.
             matchTypeImplementation = createTypeImplementation(MatchType.HeadToHead);
+            QueueImplementation = new MultiplayerQueue(this, dbFactory, hubCallbacks);
+        }
+
+        public async Task Initialise()
```
peppy, line 44: Is this split out from the constructor for the sole purpose of setting properties to their correct values before calling?

Wondering if it could potentially be moved to the `Settings` setter?

osu.Server.Spectator/Hubs/ServerMultiplayerRoom.cs
```diff
@@ -27,13 +29,22 @@ public MatchTypeImplementation MatchTypeImplementation
...
+        public readonly MultiplayerQueue QueueImplementation;
+
+        public ServerMultiplayerRoom(long roomId, IDatabaseFactory dbFactory, IMultiplayerServerMatchCallbacks hubCallbacks)
             : base(roomId)
         {
             this.hubCallbacks = hubCallbacks;
 
             // just to ensure non-null.
             matchTypeImplementation = createTypeImplementation(MatchType.HeadToHead);
+            QueueImplementation = new MultiplayerQueue(this, dbFactory, hubCallbacks);
```
peppy, line 41: Hmm, don't really like the fact that `this` is passed down here. The `Initialise` method flow feels a bit sloppy the whole way down. Wondering if `Settings` could also be propagated down in a `Settings` setter rather than this method.

It would mean using a `new` or adding `virtual` to the base `MultiplayerRoom`, but interested in hearing thoughts.

peppy: I guess the `MultiplayerRoom` is used in about 50 places so it's required to be passed. My main concern is how much cross-class operations which are going on here (the queue is doing writes to the room in many cases). Feels a bit confusing to follow.

In addition, the `ChangeMode` call takes in a new mode but is that even required if the settings are already available via the room reference? If it's to stay as-is, I'd probably want to see an assertion that the new queue mode matches the one in the room's settings.

osu.Server.Spectator/Hubs/MultiplayerQueue.cs
```diff
@@ -0,0 +1,315 @@
...
+                    case QueueModes.FairRotate:
+                        // Todo: Group playlist items by (user_id -> count_expired), and select the first available playlist item from a user that has available beatmaps where count_expired is the lowest.
+                        throw new NotImplementedException();
+                }
+            }
+
+            long lastItem = room.Settings.PlaylistItemId;
+            room.Settings.PlaylistItemId = CurrentItem.ID;
+
+            if (notifyHub && CurrentItem.ID != lastItem)
```
peppy, line 311: Is there a reason to block this? If it's just an optimisation then I don't think it's required (in the interest of keeping things as simple as possible).

osu.Server.Spectator/Hubs/MultiplayerQueue.cs
```diff
@@ -0,0 +1,315 @@
...
+        }
+
+        /// <summary>
+        /// Verifies all proposed mods are valid for the room's ruleset, returning instantiated <see cref="Mod"/>s for further processing.
+        /// </summary>
+        /// <param name="rulesetID">The legacy ruleset ID to check against.</param>
+        /// <param name="proposedMods">The proposed mods.</param>
+        /// <param name="valid">A list of valid deserialised mods.</param>
+        /// <returns>Whether all <see cref="proposedMods"/> were valid.</returns>
+        private static bool populateValidModsForRuleset(int rulesetID, IEnumerable<APIMod> proposedMods, out List<Mod> valid)
```
peppy, line 243: Ignoring that this `static` method is wedged between non-`static` counterparts, it feels like this shouldn't be here. Did I dream that you were splitting out mod comparison methods like this into a common location? I'd much rather this (and the method above) were in `ModUtils` on the osu! side, or if that doesn't make sense, a counterpart helper class in this project.

osu.Server.Spectator/Hubs/MultiplayerQueue.cs
```diff
@@ -0,0 +1,315 @@
...
+        /// <summary>
+        /// Add a playlist item to the room's queue.
+        /// </summary>
+        /// <param name="item">The item to add.</param>
+        /// <param name="user">The user adding the item.</param>
+        /// <exception cref="NotHostException">If the adding user is not the host in host-only mode.</exception>
+        /// <exception cref="InvalidStateException">If the given playlist item is not valid.</exception>
+        public async Task AddItem(MultiplayerPlaylistItem item, MultiplayerRoomUser user)
+        {
+            if (mode == QueueModes.HostOnly && (room.Host == null || !user.Equals(room.Host)))
```
peppy, line 128: What's the case of `null` host?

osu.Server.Spectator/Database/IDatabaseAccess.cs
```diff
@@ -84,5 +91,11 @@ public interface IDatabaseAccess : IDisposable
         /// Marks the given <paramref name="room"/> as ended and no longer accepting new players or scores.
         /// </summary>
         Task EndMatchAsync(MultiplayerRoom room);
+
+        /// <summary>
+        /// Retrieves all playlist items.
+        /// </summary>
+        /// <param name="roomId">The room to retrieve playlist items from.</param>
+        Task<multiplayer_playlist_item[]> GetAllPlaylistItems(long roomId);
```
peppy, line 99: Should this also have an `Async` suffix?

```suggestion
        Task<multiplayer_playlist_item[]> GetAllPlaylistItemsAsync(long roomId);
```

osu.Server.Spectator/Database/DatabaseAccess.cs
```diff
@@ -189,13 +180,34 @@ public async Task RemoveRoomParticipantAsync(MultiplayerRoom room, MultiplayerRo
...
+                VALUES (@room_id, @beatmap_id, @ruleset_id, @allowed_mods, @required_mods, NOW(), NOW())",
                 item);
 
             return await connection.QuerySingleAsync<long>("SELECT max(id) FROM multiplayer_playlist_items WHERE room_id = @room_id", item);
         }
 
+        public async Task UpdatePlaylistItemAsync(multiplayer_playlist_item item)
+        {
+            await connection.ExecuteAsync(
+                @"UPDATE multiplayer_playlist_items SET
```
peppy, line 193: Try to avoid writing database queries with whitespace in the string, it makes viewing in mysql's processlist or potentially creating query rules for it quite hard. See original code which uses `+ ""` for each line instead.

peppy (changes requested):
Initial pass, haven't gone through tests or overall behaviour yet - just pointing out things which can likely be fixed/discussed in the mean time.

osu.Server.Spectator/Hubs/MultiplayerQueue.cs
bdach: If I remember correctly the only call that doesn't notify is in `Initialise`. In which case it makes sense to not notify, but maybe this method can be factored differently as to not require the flag. Not sure.

peppy: Why does it make sense? Does the notify trigger something client-side which we want to avoid? My thinking was that it would be a no-op therefore there's no real reason to block it.

peppy: Not a huge deal either way, just looking for any way to reduce complexity. Can be left in place if it's deemed best for the server-client flow.

osu.Server.Spectator.Tests/Multiplayer/MultiplayerQueueTests.cs
smoogipoo: Should be fixed now because the method doesn't exist anymore.

osu.Server.Spectator/Hubs/MultiplayerQueue.cs
smoogipoo: Removed. It's just a no-op instead.

osu.Server.Spectator/Database/DatabaseAccess.cs
smoogipoo: I'll take note of that, because I've done the same in other projects (e.g. https://github.com/ppy/osu-queue-score-statistics/pull/6), updated everything relevant in this file.

osu.Server.Spectator/Hubs/MultiplayerQueue.cs
smoogipoo: This was to work around the issue of `MultiplayerRoomUser` implementing `Equals(X)` instead of `Equals(X?)`. I've fixed this osu!-side now, so removed the extra `null` check.

Do note, though, that `null` is the default value for host. Could also be null in the future with something like automated rooms.

osu.Server.Spectator/Database/Models/database_queue_mode.cs
```diff
@@ -0,0 +1,52 @@
...
+        fair_rotate
+    }
+
+    public static class DatabaseQueueModeExtensions
+    {
+        public static QueueModes ToQueueMode(this database_queue_mode mode)
+        {
+            switch (mode)
+            {
+                default:
```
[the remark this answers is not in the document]
smoogipoo: I was copying `database_match_type`, but I agree. I've updated both to throw exceptions instead.

osu.Server.Spectator/Hubs/MultiplayerQueue.cs
smoogipoo: It's such a specific use-case that I'm not convinced it's worth, but sure.

osu.Server.Spectator/Hubs/ServerMultiplayerRoom.cs
smoogipoo: > Hmm, don't really like the fact that this is passed down here.

I'm following existing implementation (match types) which do the same thing.

> adding `virtual` to the base `MultiplayerRoom`

Completely against this. It's a model, not a business class. We've gone over this in the lead up to `ServerMultiplayerRoom`.

> It would mean using a `new`

A `new` what? `new MultiplayerSettings Settings { get; set; }`? It doesn't work for my purposes of `Initialise`, because I need async.

> the queue is doing writes to the room in many cases

In the same way as the match type implementations are doing writes to the room in many cases. I don't necessarily disagree because the match type implementations blew my mind the first time coming back to this project in... 4 months, but I disagree that this is something to be addressed here.

> the `ChangeMode` call takes in a new mode but is that even required if the settings are already available via the room reference

Again, also applicable to the match type implementation, I'm not doing anything new here. I will add an assert.

osu.Server.Spectator/Hubs/ServerMultiplayerRoom.cs
smoogipoo: See answer below.

osu.Server.Spectator/Hubs/ServerMultiplayerRoom.cs
smoogipoo: Check whether the newest changes to `ChangeMode()` look better: https://github.com/ppy/osu-server-spectator/pull/84/commits/fed23800c923298f0e1990ba05b43c78efe83726

osu.Server.Spectator/Hubs/MultiplayerQueue.cs
```diff
@@ -118,7 +103,7 @@ public async Task AddItem(MultiplayerPlaylistItem item, MultiplayerRoomUser user
             if (room.Settings.QueueMode == QueueMode.HostOnly && !user.Equals(room.Host))
                 throw new NotHostException();
 
-            if (room.Playlist.Count(i => i.OwnerID == user.UserID && !i.Expired) >= PER_USER_LIMIT)
+            if (room.Settings.QueueMode != QueueMode.HostOnly && room.Playlist.Count(i => i.OwnerID == user.UserID && !i.Expired) >= PER_USER_LIMIT)
```
peppy, line 106: Hmm, I was thinking the limit would only apply for round robin, but I guess this is okay?

osu.Server.Spectator.Tests/Multiplayer/MultiplayerFairPlayQueueTests.cs
```diff
@@ -0,0 +1,173 @@
...
+using System.Threading.Tasks;
+using Moq;
+using osu.Game.Online.Multiplayer;
+using osu.Game.Online.Multiplayer.Queueing;
+using osu.Game.Online.Rooms;
+using Xunit;
+
+namespace osu.Server.Spectator.Tests.Multiplayer
+{
+    public class MultiplayerFairPlayQueueTests : MultiplayerTest
```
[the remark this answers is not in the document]
peppy: or whatever new term we go with. just not sure where "fair play" came from.

osu.Server.Spectator/Hubs/ServerMultiplayerRoom.cs
peppy: Marginally I guess. Not a blocker either way, just reads a bit weird.

osu.Server.Spectator/Hubs/IMultiplayerServerMatchCallbacks.cs
```diff
@@ -31,5 +32,32 @@ public interface IMultiplayerServerMatchCallbacks
         /// <param name="room">The room to send the event to.</param>
         /// <param name="user">The user whose state has changed.</param>
         Task UpdateMatchUserState(MultiplayerRoom room, MultiplayerRoomUser user);
+
+        /// <summary>
+        /// Let the hub know that a playlist item has been added.
+        /// </summary>
+        /// <param name="room">The room to send the event to.</param>
+        /// <param name="item">The added item.</param>
+        Task OnPlaylistItemAdded(ServerMultiplayerRoom room, MultiplayerPlaylistItem item);
```
peppy, line 41: All the existing methods specify `MultiplayerRoom` in this interface, would prefer these either do too, or the existing usages are updated to use `ServerMultiplayerRoom`. Probably the latter.

Reasoning is that the purpose of this interface when viewed in isolation becomes a touch uncertain with mixed usage.

osu.Server.Spectator/Hubs/MultiplayerHub.cs
```diff
@@ -178,32 +174,18 @@ private async Task<ServerMultiplayerRoom> retrieveRoom(long roomId)
...
-                var beatmapChecksum = await db.GetBeatmapChecksumAsync(playlistItem.beatmap_id);
-
-                if (beatmapChecksum == null)
-                    throw new InvalidOperationException($"Expected non-null checksum on beatmap ID {playlistItem.beatmap_id}");
-
-                var requiredMods = JsonConvert.DeserializeObject<IEnumerable<APIMod>>(playlistItem.required_mods ?? string.Empty) ?? Array.Empty<APIMod>();
-                var allowedMods = JsonConvert.DeserializeObject<IEnumerable<APIMod>>(playlistItem.allowed_mods ?? string.Empty) ?? Array.Empty<APIMod>();
-
-                var room = new ServerMultiplayerRoom(roomId, this)
+                var room = new ServerMultiplayerRoom(roomId, databaseFactory, this)
```
peppy, line 177: I know we agreed on passing in the `databaseFactory` here, but for the `Initialise()` calls specifically, I'd feel a lot happier with them if they were `Initialise(databaseFactory)`, implying that the `ctor` didn't do the actual room/database setup via the method structure.

Doesn't need to change, just mentioning in passing.

osu.Server.Spectator/Hubs/IMultiplayerServerMatchCallbacks.cs
smoogipoo: The latter is the only one possible, because it's actually being used as a `ServerMultiplayerRoom` to access the queue (to validate mods).

Or otherwise I could also re-query the room, but that seems wasteful...

osu.Server.Spectator/Hubs/MultiplayerHub.cs
smoogipoo: So passing database factory into the ctor but also passing it into the initialise? Or are you expecting initialise to store the factory?

osu.Server.Spectator/Hubs/IMultiplayerServerMatchCallbacks.cs
peppy: Latter is fine 👍

osu.Server.Spectator/Hubs/MultiplayerHub.cs
peppy: Wasn't expecting any change. But what you mention would be better if it can work, I think?

osu.Server.Spectator/Hubs/MultiplayerQueue.cs
```diff
@@ -0,0 +1,279 @@
...
+            }
+        }
+
+        /// <summary>
+        /// Checks whether the given mods are compatible with the current playlist item's mods and ruleset.
+        /// </summary>
+        /// <param name="proposedMods">The mods to check against the current playlist item.</param>
+        /// <param name="validMods">The set of mods which _are_ valid.</param>
+        /// <returns>Whether all mods are valid for the current playlist item.</returns>
+        public bool ValidateMods(IEnumerable<APIMod> proposedMods, [NotNullWhen(false)] out IEnumerable<APIMod>? validMods)
```
peppy, line 157: Looks like this only uses the `CurrentItem`, would this potentially sit better in the model class? Not sure if there would be uses osu!-side, but being in the queue implementation makes it seem like the validation will change per queue type.

peppy: Also I'd probably call this `ValidateUserMods` if that's the intention, since there's a second different validation method direction below this which is used in different scenarios.

osu.Server.Spectator/Hubs/ServerMultiplayerRoom.cs
```diff
@@ -27,13 +29,22 @@ public MatchTypeImplementation MatchTypeImplementation
             }
         }
 
-        public ServerMultiplayerRoom(long roomId, IMultiplayerServerMatchCallbacks hubCallbacks)
+        public readonly MultiplayerQueue QueueImplementation;
```
peppy, line 32: Shall we drop the `Implementation` suffix on this (and the existing usage)? Doesn't seem like it's required.

osu.Server.Spectator/Hubs/MultiplayerQueue.cs
smoogipoo: By model you mean `MultiplayerRoom`?

Hmm... Potentially. The model only contains `PlaylistItemId` so no real connection to the model itself, but there's two ways around that:
- Replace `MultiplayerRoomSettings.PlaylistItemId` with a `MultiplayerPlaylistItem`, denoting the current item.
- Do a manual lookup every time (`CurrentItem => Playlist.First(item => item.ID == settings.PlaylistItemId)`.

Is that what you're envisioning?

smoogipoo: Perhaps in the future multiplayer could be updated so the host calls `ChangeSettings()` with the new settings for the current playlist item...? Rather than `AddItem()` to do the same (currently) or a separate `ChangeItem()` in the future.. Though we may still want other players to be able to change items they've enqueued, so not sure how helpful that addition is.

peppy: I meant `MultiplayerPlaylistItem`

currently:
`match.Queue.ValidateMods(..)`
post:
`match.CurrentItem.ValidateMods(..)`

smoogipoo: Yeah, so should I be removing `MultiplayerRoomSettings.PlaylistItemId` then?

smoogipoo: Moved them to extension methods on `MultiplayerPlaylistItem`, currently only available on the server. Eventually will need to take in an `IRulesetStore` or something, at which point it can be moved to the osu!-side model.

Didn't move the current item out of the queue and into `ServerMultiplayerRoom` yet because that needs a bit more finessing - I was scared of potential O(n^2) operation when validating mods for all users in the room (e.g. upon host changing the playlist item).

osu.Server.Spectator/Hubs/MultiplayerHub.cs
smoogipoo: Moved `IDatabaseFactory` into `Initialise()`, see if you agree: https://github.com/ppy/osu-server-spectator/pull/84/commits/bf5372a77ea027d8f47dc1e696b14bde2b1418dd

osu.Server.Spectator/Hubs/MultiplayerQueue.cs
```diff
@@ -0,0 +1,210 @@
...
+    public class MultiplayerQueue
+    {
+        public const int PER_USER_LIMIT = 3;
+
+        public MultiplayerPlaylistItem CurrentItem => room.Playlist[currentIndex];
+
+        private readonly ServerMultiplayerRoom room;
+        private readonly IMultiplayerServerMatchCallbacks hub;
+
+        private IDatabaseFactory dbFactory = null!;
```
peppy, line 27: Would you be against making this nullable and adding null check / throw on any method which requires it? Feels safer (ensures that `Initialise` is called).

osu.Server.Spectator/Database/Models/database_queue_mode.cs
```diff
@@ -24,11 +24,11 @@ public static QueueMode ToQueueMode(this database_queue_mode mode)
...
 
-                case database_queue_mode.free_for_all:
-                    return QueueMode.FreeForAll;
+                case database_queue_mode.all_players:
+                    return QueueMode.AllPlayers;
 
-                case database_queue_mode.fair_rotate:
-                    return QueueMode.FairRotate;
+                case database_queue_mode.all_players_rr:
+                    return QueueMode.AllPlayersRR;
```
peppy, line 31: hmm, i think this one should maybe be `AllPlayerRoundRobin`... not sure

smoogipoo: Renamed to the full name in all cases not too bad overall.

osu.Server.Spectator/Hubs/MultiplayerQueue.cs
```diff
@@ -0,0 +1,216 @@
...
+        /// </summary>
+        public async Task UpdateFromQueueModeChange()
+        {
+            if (dbFactory == null) throw new InvalidOperationException($"Call {nameof(Initialise)} first.");
+
+            if (room.Settings.QueueMode == QueueMode.HostOnly && room.Playlist.All(item => item.Expired))
+            {
+                // When changing to host-only mode, ensure that exactly one non-expired playlist item exists by duplicating the currrent item.
+                using (var db = dbFactory.GetInstance())
+                    await duplicateCurrentItem(db);
```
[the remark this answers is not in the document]
smoogipoo: We've since decided to not clear the previous non-expired items, so the implementation is correct. I've fixed up the comment.

peppy:
I've updated the signalr packages to 6.0 because the [changes don't look too large](https://docs.microsoft.com/en-us/aspnet/core/release-notes/aspnetcore-6.0?view=aspnetcore-6.0). Have tested this to work with 5.x clients fine.

peppy (approved):

PR #85, opened 2021-11-15
Add fair play multiplayer queueing mode

smoogipoo:
Prereqs:
- [ ] https://github.com/ppy/osu-server-spectator/pull/84


smoogipoo:
Merged into https://github.com/ppy/osu-server-spectator/pull/84 because it's easier to manage as structures move around.

PR #87, opened 2021-12-02
Rework multiplayer queue to expose item ordering to clients

smoogipoo:
Prereqs:
- [x] https://github.com/ppy/osu-web/pull/8408
- [x] https://github.com/ppy/osu-web/pull/8411
- [x] https://github.com/ppy/osu/pull/15890 (circular dependency)

Adds the new `PlaylistOrder` field to `MultiplayerPlaylistItem` and related classes, which allows the server to communicate the order with clients.

### DIFF
--- a/osu.Server.Spectator.Tests/Multiplayer/MultiplayerAllPlayersRoundRobinQueueTests.cs
+++ b/osu.Server.Spectator.Tests/Multiplayer/MultiplayerAllPlayersRoundRobinQueueTests.cs
@@ -44,47 +44,16 @@ namespace osu.Server.Spectator.Tests.Multiplayer
                 BeatmapChecksum = "5555"
             });
 
-            using (var usage = Hub.GetRoom(ROOM_ID))
-            {
-                var room = usage.Item;
-                Debug.Assert(room != null);
-
-                // Item 1 selected (user 1)
-                Assert.Equal(1, room.Settings.PlaylistItemId);
-            }
+            checkCurrentItem(1);
 
             await runGameplay();
-
-            using (var usage = Hub.GetRoom(ROOM_ID))
-            {
-                var room = usage.Item;
-                Debug.Assert(room != null);
-
-                // Item 4 selected (user 2)
-                Assert.Equal(4, room.Settings.PlaylistItemId);
-            }
+            checkCurrentItem(4);
 
             await runGameplay();
-
-            using (var usage = Hub.GetRoom(ROOM_ID))
-            {
-                var room = usage.Item;
-                Debug.Assert(room != null);
-
-                // Item 2 selected (user 1)
-                Assert.Equal(2, room.Settings.PlaylistItemId);
-            }
+            checkCurrentItem(2);
 
             await runGameplay();
-
-            using (var usage = Hub.GetRoom(ROOM_ID))
-            {
-                var room = usage.Item;
-                Debug.Assert(room != null);
-
-                // Item 3 selected (user 1). User 2 has no more items available.
-                Assert.Equal(3, room.Settings.PlaylistItemId);
-            }
+            checkCurrentItem(3);
         }
 
         [Fact]
@@ -112,37 +81,57 @@ namespace osu.Server.Spectator.Tests.Multiplayer
             });
 
             await runGameplay();
-
-            using (var usage = Hub.GetRoom(ROOM_ID))
-            {
-                var room = usage.Item;
-                Debug.Assert(room != null);
-
-                // After playing the first item, the second item (by player 1) is selected.
-                Assert.Equal(2, room.Settings.PlaylistItemId);
-            }
+            checkCurrentItem(2);
 
             await Hub.ChangeSettings(new MultiplayerRoomSettings { QueueMode = QueueMode.AllPlayersRoundRobin });
-
-            using (var usage = Hub.GetRoom(ROOM_ID))
-            {
-                var room = usage.Item;
-                Debug.Assert(room != null);
-
-                // After changing to fair-play mode, we expect the third item (by player 2) to be selected.
-                Assert.Equal(3, room.Settings.PlaylistItemId);
-            }
+            checkCurrentItem(3);
 
             await Hub.ChangeSettings(new MultiplayerRoomSettings { QueueMode = QueueMode.AllPlayers });
+            checkCurrentItem(2);
+        }
 
-            using (var usage = Hub.GetRoom(ROOM_ID))
+        [Fact]
+        public async Task CurrentItemOrderedByPlayedAt()
+        {
+            Database.Setup(d => d.GetBeatmapChecksumAsync(3333)).ReturnsAsync("3333");
+
+            await Hub.JoinRoom(ROOM_ID);
+            await Hub.ChangeSettings(new MultiplayerRoomSettings { QueueMode = QueueMode.AllPlayersRoundRobin });
+
+            SetUserContext(ContextUser2);
+            await Hub.JoinRoom(ROOM_ID);
+            SetUserContext(ContextUser);
+
+            await Hub.AddPlaylistItem(new MultiplayerPlaylistItem
             {
-                var room = usage.Item;
-                Debug.Assert(room != null);
+                BeatmapID = 3333,
+                BeatmapChecksum = "3333"
+            });
 
-                // After changing back to free-for-all mode, we expect the second item (by player 1) to be selected.
-                Assert.Equal(2, room.Settings.PlaylistItemId);
-            }
+            await runGameplay();
+            await runGameplay();
+
+            // Item 3: Now the only non-expired item in the room.
+            await Hub.AddPlaylistItem(new MultiplayerPlaylistItem
+            {
+                BeatmapID = 3333,
+                BeatmapChecksum = "3333"
+            });
+
+            SetUserContext(ContextUser2);
+
+            // Item 4: Because this user hasn't had any items played, this item will be moved to the start.
+            await Hub.AddPlaylistItem(new MultiplayerPlaylistItem
+            {
+                BeatmapID = 3333,
+                BeatmapChecksum = "3333"
+            });
+
+            await runGameplay();
+            await runGameplay();
+
+            // Item 3 should be the "current" item, as it is in order of play rather than order of addition.
+            checkCurrentItem(3);
         }
 
         private async Task runGameplay()
@@ -167,6 +156,18 @@ namespace osu.Server.Spectator.Tests.Multiplayer
             await Hub.ChangeState(MultiplayerUserState.Idle);
 
             SetUserContext(ContextUser);
+        }
+
+        private void checkCurrentItem(long expectedItemId)
+        {
+            using (var usage = Hub.GetRoom(ROOM_ID))
+            {
+                var room = usage.Item;
+                Debug.Assert(room != null);
+
+                // After changing back to free-for-all mode, we expect the second item (by player 1) to be selected.
+                Assert.Equal(expectedItemId, room.Settings.PlaylistItemId);
+            }
         }
     }
 }
--- a/osu.Server.Spectator.Tests/Multiplayer/MultiplayerAllPlayersRoundRobinQueueTests.cs
+++ b/osu.Server.Spectator.Tests/Multiplayer/MultiplayerAllPlayersRoundRobinQueueTests.cs
@@ -165,7 +165,6 @@ namespace osu.Server.Spectator.Tests.Multiplayer
                 var room = usage.Item;
                 Debug.Assert(room != null);
 
-                // After changing back to free-for-all mode, we expect the second item (by player 1) to be selected.
                 Assert.Equal(expectedItemId, room.Settings.PlaylistItemId);
             }
         }
--- a/osu.Server.Spectator.Tests/Multiplayer/MultiplayerHostOnlyQueueTests.cs
+++ b/osu.Server.Spectator.Tests/Multiplayer/MultiplayerHostOnlyQueueTests.cs
@@ -86,8 +86,8 @@ namespace osu.Server.Spectator.Tests.Multiplayer
 
                 // Players received callbacks.
                 Receiver.Verify(r => r.PlaylistItemAdded(It.Is<MultiplayerPlaylistItem>(i => i.ID == newItem.ID)), Times.Once);
-                Receiver.Verify(r => r.PlaylistItemChanged(It.Is<MultiplayerPlaylistItem>(i => i.ID == oldItem.ID && i.Expired)), Times.Once);
-                Receiver.Verify(r => r.SettingsChanged(room.Settings), Times.Exactly(2));
+                Receiver.Verify(r => r.PlaylistItemChanged(It.Is<MultiplayerPlaylistItem>(i => i.ID == oldItem.ID && i.Expired)));
+                Receiver.Verify(r => r.SettingsChanged(room.Settings));
             }
 
             // And a second time...
@@ -116,8 +116,8 @@ namespace osu.Server.Spectator.Tests.Multiplayer
 
                 // Players received callbacks.
                 Receiver.Verify(r => r.PlaylistItemAdded(It.Is<MultiplayerPlaylistItem>(i => i.ID == newItem.ID)), Times.Once);
-                Receiver.Verify(r => r.PlaylistItemChanged(It.Is<MultiplayerPlaylistItem>(i => i.ID == oldItem.ID && i.Expired)), Times.Once);
-                Receiver.Verify(r => r.SettingsChanged(room.Settings), Times.Exactly(3));
+                Receiver.Verify(r => r.PlaylistItemChanged(It.Is<MultiplayerPlaylistItem>(i => i.ID == oldItem.ID && i.Expired)));
+                Receiver.Verify(r => r.SettingsChanged(room.Settings));
             }
         }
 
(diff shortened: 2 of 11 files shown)


smoogipoo:
This now also sets `PlaylistOrder` for historical items too, resolving https://github.com/ppy/osu/pull/15890#discussion_r760825945.

I'm not 100% sure on the implementation of `updatePlaylistOrder` - ideally I'd keep the number of `PlaylistItemChanged` callbacks to a minimum, but it should be adequate for the time being and can be iterated on later.

Also, I haven't changed the round robin algorithm, which is also something for a further PR.

osu.Server.Spectator/Database/Models/multiplayer_playlist_item.cs
```diff
@@ -19,13 +19,21 @@ public class multiplayer_playlist_item
...
-        public DateTimeOffset? created_at { get; set; }
-        public DateTimeOffset? updated_at { get; set; }
+
+        /// <summary>
+        /// Read-only.
+        /// </summary>
+        public DateTimeOffset? created_at { get; set; } = DateTimeOffset.Now;
+
+        /// <summary>
+        /// Read-only.
```
peppy, line 32: Arguably this should be updated each time we change data in a row (I don't think we're enforcing this at a mysql level as laravel handles this for us, but just outlining the expectation).

If this *has* to be readonly, probably worth adding reasoning as to why.

osu.Server.Spectator/Hubs/MultiplayerQueue.cs
```diff
@@ -7,6 +7,7 @@
 using System.Collections.Generic;
 using System.Linq;
 using System.Threading.Tasks;
+using NuGet.Packaging;
```
peppy, line 10: oops

osu.Server.Spectator/Hubs/MultiplayerQueue.cs
```diff
@@ -228,32 +229,41 @@ private async Task updatePlaylistOrder(IDatabaseAccess db)
...
                         perUserCounts[candidateItem.OwnerID]++;
                     }
 
-                    orderedItems = processedItems;
                     break;
             }
 
-            for (int i = 0; i < orderedItems.Count; i++)
+            // For expired items, it's important that they're ordered in ascending order such that the last updated item is the last in the list.
+            // This is so that the updated_at database column doesn't get refreshed as a result of change in ordering.
```
peppy, line 241: This really sounds like you want another column that isn't `updated_at`. That's what we do in all other instances where we don't expect the column to be updated on write.

Actually important, since if laravel happens to touch one of these rows for any reason in the future, it *is* going to modify the column as it stands.

peppy (changes requested):

osu.Server.Spectator/Database/Models/multiplayer_playlist_item.cs
smoogipoo: It’s updated directly in the queries of `DatabaseAccess`. It’s read-only in that if you call `db.UpdatePlaylistItem(item)`, the `updated_at` value isn’t taken from the item but calculated at a query level.

peppy: Hmm.. so it sounds like the correct behaviour then. But this xmldoc doesn't really explain that and will likely lead to confusion.

peppy: Is it possible to make these `get` only or protect them in some other way? Or just remove the xmldoc because it's kinda implied and will be overwritten anyways.

smoogipoo: Will change to `Changes to this property will not be persisted to the database.`

osu.Server.Spectator/Hubs/MultiplayerQueue.cs
smoogipoo: Hmmm... So I guess we do need an `expired_at` anyway.

peppy: I'd say that's safest, yeah.

Can happen as a separate change if that's easier.

peppy:
I'm fine with this PR as it stands, but I guess it's best to leave it open until the database structural changes are completed?

osu.Server.Spectator/Hubs/MultiplayerQueue.cs
```diff
@@ -77,17 +81,16 @@ public async Task FinishCurrentItem()
             using (var db = dbFactory.GetInstance())
             {
                 // Expire and let clients know that the current item has finished.
-                await db.ExpirePlaylistItemAsync(CurrentItem.ID);
+                await db.MarkPlaylistItemAsPlayedAsync(CurrentItem.ID);
                 CurrentItem.Expired = true;
+                CurrentItem.PlayedAt = DateTimeOffset.Now;
```
bdach, line 86: Dunno if this is me being paranoid, but I find the fact that both the DB and the server generate this timestamp independently (via SQL `NOW()` and via `DateTimeOffset.Now` respectively) kinda sketchy... I would hope this property is display-only, but if this is to be used in actual logic then that may become an issue.

osu.Server.Spectator/Hubs/MultiplayerQueue.cs
```diff
@@ -155,61 +156,106 @@ public async Task AddItem(MultiplayerPlaylistItem item, MultiplayerRoomUser user
...
         /// Updates <see cref="CurrentItem"/> and the playlist item ID stored in the room's settings.
         /// </summary>
         private async Task updateCurrentItem()
         {
-            MultiplayerPlaylistItem newItem;
+            MultiplayerPlaylistItem nextItem = room.Playlist
+                                                   .Where(i => !i.Expired)
+                                                   .OrderBy(i => i.PlaylistOrder)
+                                                   .FirstOrDefault()
+                                               ?? room.Playlist.Last();
```
bdach, line 189: Shouldn't this fallback "last" case also be sorted by `PlaylistOrder` just for sanity?

osu.Server.Spectator/Hubs/MultiplayerQueue.cs
```diff
@@ -155,61 +156,106 @@ public async Task AddItem(MultiplayerPlaylistItem item, MultiplayerRoomUser user
...
+        private async Task updatePlaylistOrder(IDatabaseAccess db)
+        {
+            List<MultiplayerPlaylistItem> orderedActiveItems;
 
             switch (room.Settings.QueueMode)
             {
                 default:
-                    // Pick the first available non-expired playlist item, or default to the last item for when all items are expired.
-                    newItem = room.Playlist.FirstOrDefault(i => !i.Expired) ?? room.Playlist.Last();
+                    orderedActiveItems = room.Playlist.Where(item => !item.Expired).OrderBy(item => item.ID == 0 ? int.MaxValue : item.ID).ToList();
```
bdach, line 210: So I presume the 0 special case is to accommodate the `addItem()` usage? Because in that case the model is sorted before being saved to DB and therefore being assigned an ID?

If that is correct I'd probably leave an inline comment here because at face value I find that behaviour kinda weird...

osu.Server.Spectator.Tests/Multiplayer/MultiplayerHostOnlyQueueTests.cs
```diff
@@ -86,8 +86,8 @@ public async Task RoomHasNewPlaylistItemAfterMatchStartWithOneItem()
 
                 // Players received callbacks.
                 Receiver.Verify(r => r.PlaylistItemAdded(It.Is<MultiplayerPlaylistItem>(i => i.ID == newItem.ID)), Times.Once);
-                Receiver.Verify(r => r.PlaylistItemChanged(It.Is<MultiplayerPlaylistItem>(i => i.ID == oldItem.ID && i.Expired)), Times.Once);
-                Receiver.Verify(r => r.SettingsChanged(room.Settings), Times.Exactly(2));
+                Receiver.Verify(r => r.PlaylistItemChanged(It.Is<MultiplayerPlaylistItem>(i => i.ID == oldItem.ID && i.Expired)));
```
bdach, line 89: Was there a particular reason for dropping the invocation count part of these verify calls?

smoogipoo: When finishing a playlist item there's two updates that occur - one to mark the item as expired, and another to reorder _all_ playlist items.
This may not be required anymore because expired items don't get reordered anymore - their `PlaylistOrder` value doesn't change.

That being said, I don't think the changes to this test are that important? The way I see it is they're testing implementation details rather than semantics, and I originally was going to go even further but left that for a future PR. Do you disagree with that?

osu.Server.Spectator/Hubs/MultiplayerQueue.cs
smoogipoo: I am tending to agree. We should do a re-query here...

osu.Server.Spectator/Hubs/MultiplayerQueue.cs
smoogipoo: True, good catch.

osu.Server.Spectator.Tests/Multiplayer/MultiplayerHostOnlyQueueTests.cs
bdach: I don't necessarily disagree with dropping the invocation count checks; on one side they may be useful to ensure there are no unnecessary / misplaced calls, but it does make the tests more brittle to implementation details changing. I don't think I would object to removing, but I didn't really have a problem with them being here either. Not sure, I mostly just asked out of curiosity more than anything.

osu.Server.Spectator/Hubs/MultiplayerQueue.cs
smoogipoo: I've cleaned this up a bit by setting `long.MaxValue` in `addItem`, which removes these checks scattered across this method. The only remaining one is at the very bottom of this method which previously had and still does have an inline comment.

osu.Server.Spectator.Tests/Multiplayer/MultiplayerHostOnlyQueueTests.cs
smoogipoo: Have reverted for the time being just to reduce the scope of changes.

osu.Server.Spectator/Database/DatabaseAccess.cs
```diff
@@ -168,20 +168,19 @@ public async Task RemoveRoomParticipantAsync(MultiplayerRoom room, MultiplayerRo
             }
         }
 
-        public async Task<multiplayer_playlist_item?> GetPlaylistItemFromRoomAsync(long roomId, long playlistItemId)
+        public async Task<multiplayer_playlist_item> GetPlaylistItemAsync(long playlistItemId)
         {
-            return await connection.QueryFirstOrDefaultAsync<multiplayer_playlist_item>("SELECT * FROM multiplayer_playlist_items WHERE id = @Id AND room_id = @RoomId", new
+            return await connection.QueryFirstOrDefaultAsync<multiplayer_playlist_item>("SELECT * FROM multiplayer_playlist_items WHERE id = @Id", new
```
bdach, line 173: Hmm, a few small things here:

* The nullable type annotation was dropped, while the linq query still uses `FirstOrDefaultAsync`. Would prefer either bringing the annotation back, or using `QueryFirst`/`QuerySingle`.
* Also not super sure about dropping the room check. May be better/nicer to keep it for safety, to avoid accidentally performing cross-room fetches. This is also relevant to `MarkPlaylistItemAsPlayedAsync`, which also uses a single playlist item ID rather than playlist item ID + room ID.

osu.Server.Spectator.Tests/Multiplayer/MultiplayerAllPlayersRoundRobinQueueTests.cs
```diff
@@ -168,5 +157,17 @@ private async Task runGameplay()
...
         }
+
+        private void checkCurrentItem(long expectedItemId)
+        {
+            using (var usage = Hub.GetRoom(ROOM_ID))
+            {
+                var room = usage.Item;
+                Debug.Assert(room != null);
+
+                // After changing back to free-for-all mode, we expect the second item (by player 1) to be selected.
```
bdach, line 168: comment no longer applies in this context

```suggestion
```

osu.Server.Spectator/Database/DatabaseAccess.cs
smoogipoo: Fixed the `FirstOrDefaultAsync()`. I intended to change it but forgot.

I personally find providing a room ID superfluous, but I've also applied that change.

bdach:
Code seems fine to me now, but I haven't tested this in any real capacity yet.

osu.Server.Spectator/Hubs/MultiplayerQueue.cs
```diff
@@ -155,61 +155,107 @@ public async Task AddItem(MultiplayerPlaylistItem item, MultiplayerRoomUser user
...
         }
 
         /// <summary>
         /// Updates <see cref="CurrentItem"/> and the playlist item ID stored in the room's settings.
         /// </summary>
         private async Task updateCurrentItem()
         {
-            MultiplayerPlaylistItem newItem;
+            // Pick the next non-expired playlist item by playlist order, or default to the most-recently-expired item.
+            MultiplayerPlaylistItem nextItem = room.Playlist.Where(i => !i.Expired).OrderBy(i => i.PlaylistOrder).FirstOrDefault()
```
peppy, line 187: To confirm, the database ordering is only valid for non-expired items, and will overlap with expired? So it's basically a throwaway value which is meaningless (should explicitly never be read) once the item is expired?

Wondering if the `expired` column should be removed and this one should be used instead, setting it to `-1` for expired.

osu.Server.Spectator/Hubs/MultiplayerQueue.cs
```diff
@@ -155,61 +155,107 @@ public async Task AddItem(MultiplayerPlaylistItem item, MultiplayerRoomUser user
...
+            BeatmapChecksum = CurrentItem.BeatmapChecksum,
+            RulesetID = CurrentItem.RulesetID,
+            AllowedMods = CurrentItem.AllowedMods,
+            RequiredMods = CurrentItem.RequiredMods
+        });
+
+        private async Task addItem(IDatabaseAccess db, MultiplayerPlaylistItem item)
+        {
+            // Add the item to the list with an initial "infinite" ID, in order to notify updatePlaylistItem()
+            // that the item should be de-prioritised and events should not be propagated for it.
```
peppy, line 171: This explains why the ID is set to `MaxValue` (which i'm not sure I agree with yet) but doesn't explain why the playlist order needs to be updated in this order. I'd need to understand why that is required before making a call on whether this `long.MaxValue` logic is something I can get behind or not.

smoogipoo: I hoped that this being in "addItem" would explain why the item should be deprioritised. Because it's the last item added?

I could just as well have given it a value of `Current_Max_ID + 1` and it would be the same ordering. I'm not sure how I could explain this any better - it's at the end of the list.

osu.Server.Spectator/Hubs/MultiplayerQueue.cs
smoogipoo: Yeah it's pretty much an undefined value for expired items. I originally had a doc on it but the semantics have changed a few times and the doc is now gone.

osu.Server.Spectator/Hubs/MultiplayerQueue.cs
peppy: Is there a reason the order of these operations couldn't be reversed to get the same behaviour? The `playlist_order` does seem to be null initially, so I think that would still be in a manageable state?

smoogipoo: I don't understand what you mean. Reverse which operations? The adding to list is required for the ordering to happen.

peppy: Something along the lines of this:
20211206 123729 (JetBrains Rider-EAP)

smoogipoo: I do not want to propagate `PlaylistItemChanged` events before `PlaylistItemAdded` events.

peppy: so.. 

```csharp

        private async Task addItem(IDatabaseAccess db, MultiplayerPlaylistItem item)
        {
            item.ID = await db.AddPlaylistItemAsync(new multiplayer_playlist_item(room.RoomID, item));

            room.Playlist.Add(item);

            await hub.OnPlaylistItemAdded(room, item);

            await updatePlaylistOrder(db);
        }

```

?

smoogipoo: And so client side the item shows up in the incorrect position and then pops back into place later on? I suppose it's fine otherwise.

peppy: I think I'd prefer that to this workaround. Seems like something we can clean up in the future. Prefer that the server logic is as simple as possible for now.

peppy (approved):

peppy:
I've pushed this out to the dev server for testing.

PR #89, opened 2021-12-06
Rework round robin ordering to not pin users to the front of the queue

smoogipoo:
Resolves https://github.com/ppy/osu/issues/15903

As per the tests, the following order now takes place. Notice in particular how items no longer jump the queue.

```
- User 1 adds two items (1, 2)

{ item: 1, user: 1 } <
{ item: 2, user: 1 }

- Item played

{ item: 2, user: 1 } <

- User 2 adds two items (3, 4)

{ item: 2, user: 1 } <
{ item: 3, user: 2 }
{ item: 4, user: 2 }

- User 1 adds an item (5)

{ item: 2, user: 1 } <
{ item: 3, user: 2 }
{ item: 5, user: 1 }
{ item: 4, user: 2 }
```

### DIFF
--- a/osu.Server.Spectator.Tests/Multiplayer/MultiplayerAllPlayersRoundRobinQueueTests.cs
+++ b/osu.Server.Spectator.Tests/Multiplayer/MultiplayerAllPlayersRoundRobinQueueTests.cs
@@ -2,6 +2,7 @@
 // See the LICENCE file in the repository root for full licence text.
 
 using System.Diagnostics;
+using System.Linq;
 using System.Threading.Tasks;
 using Moq;
 using osu.Game.Online.Multiplayer;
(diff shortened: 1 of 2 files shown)


osu.Server.Spectator.Tests/Multiplayer/MultiplayerAllPlayersRoundRobinQueueTests.cs
```diff
@@ -13,125 +14,81 @@ namespace osu.Server.Spectator.Tests.Multiplayer
...
 
-            await Hub.ChangeSettings(new MultiplayerRoomSettings { QueueMode = QueueMode.AllPlayersRoundRobin });
-            checkCurrentItem(3);
-
-            await Hub.ChangeSettings(new MultiplayerRoomSettings { QueueMode = QueueMode.AllPlayers });
-            checkCurrentItem(2);
-        }
-
-        [Fact]
-        public async Task CurrentItemOrderedByPlayedAt()
```
smoogipoo, line 94: I've removed these other tests because they specifically tested for jumping in the front of the queue. This class only contains one test now which is the full round robin sequence.

This test in particular is upheld by the last few lines at the bottom of the new test, that ensure the current item doesn't change after the lest item in the room is played.

peppy (approved):
Not 101% sure on this, but seems much better than what we have, so let's get this out with the release.

PR #93, opened 2021-12-08
Implement playlist item removal

smoogipoo:
Prereqs:
- [x] https://github.com/ppy/osu/pull/16010

Not super sure about the condition for not being able to delete in host-only mode, but that can be changed at any time.

### DIFF
--- a/osu.Server.Spectator/Hubs/MultiplayerQueue.cs
+++ b/osu.Server.Spectator/Hubs/MultiplayerQueue.cs
@@ -152,6 +152,41 @@ namespace osu.Server.Spectator.Hubs
         }
 
         /// <summary>
+        /// Removes a playlist item from the room's queue.
+        /// </summary>
+        /// <param name="playlistItemId">The item to remove.</param>
+        /// <param name="user">The user removing the item.</param>
+        public async Task RemoveItem(long playlistItemId, MultiplayerRoomUser user)
+        {
+            if (dbFactory == null) throw new InvalidOperationException($"Call {nameof(Initialise)} first.");
+
+            if (room.Settings.QueueMode == QueueMode.HostOnly)
+                throw new InvalidStateException("Items cannot be removed in host-only mode.");
+
+            if (room.Playlist.Count == 1)
+                throw new InvalidStateException("The singular item in the room cannot be removed.");
+
+            var item = room.Playlist.FirstOrDefault(item => item.ID == playlistItemId);
+
+            if (item == null)
+                throw new InvalidStateException("Item does not exist in the room.");
+
+            if (item == CurrentItem)
+                throw new InvalidStateException("The room's current item cannot be removed.");
+
+            if (item.OwnerID != user.UserID)
+                throw new InvalidStateException("Attempted to remove an item which is not owned by the user.");
+
+            using (var db = dbFactory.GetInstance())
+                await db.RemovePlaylistItemAsync(room.RoomID, playlistItemId);
+
+            room.Playlist.Remove(item);
+            await hub.OnPlaylistItemRemoved(room, playlistItemId);
+
+            await updateCurrentItem();
+        }
+
+        /// <summary>
         /// Duplicates <see cref="CurrentItem"/> into the database.
         /// </summary>
         /// <param name="db">The database connection.</param>
--- a/osu.Server.Spectator/Hubs/MultiplayerQueue.cs
+++ b/osu.Server.Spectator/Hubs/MultiplayerQueue.cs
@@ -160,9 +160,6 @@ namespace osu.Server.Spectator.Hubs
         {
             if (dbFactory == null) throw new InvalidOperationException($"Call {nameof(Initialise)} first.");
 
-            if (room.Settings.QueueMode == QueueMode.HostOnly)
-                throw new InvalidStateException("Items cannot be removed in host-only mode.");
-
             var item = room.Playlist.FirstOrDefault(item => item.ID == playlistItemId);
 
             if (item == null)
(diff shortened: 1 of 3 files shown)


osu.Server.Spectator/Hubs/MultiplayerQueue.cs
```diff
@@ -151,6 +151,41 @@ public async Task AddItem(MultiplayerPlaylistItem item, MultiplayerRoomUser user
...
+        /// Removes a playlist item from the room's queue.
+        /// </summary>
+        /// <param name="playlistItemId">The item to remove.</param>
+        /// <param name="user">The user removing the item.</param>
+        public async Task RemoveItem(long playlistItemId, MultiplayerRoomUser user)
+        {
+            if (dbFactory == null) throw new InvalidOperationException($"Call {nameof(Initialise)} first.");
+
+            if (room.Settings.QueueMode == QueueMode.HostOnly)
+                throw new InvalidStateException("Items cannot be removed in host-only mode.");
```
bdach, line 164: Yeah, I'm not sure about this either. With the way how I understand queue mode to work (in that items from previous queue mode stay there after change), you could have a situation wherein a room was in round robin and there were several items enqueued by users, and then if the host changes to host-only, nobody can remove any of those items anymore, host included.

A simple compromise may be to drop this check altogether and rely on the "cannot delete current item" check for sanity. I think that should be enough to ensure users don't get themselves into a broken state? That's also just my opinion, probably a good idea to wait for another.

bdach:
As a loose aside I'm not sure this strictly depends on the game-side PR. Functionally I can see it, but mechanically speaking this seems to compile fine without.

peppy:
Would hosts being able to delete other users' items be a follow up? Another case which has come up commonly is a user leaving or being kiced from a room and the remaining players not necessarily wanting to play the beatmap left behind.

osu.Server.Spectator/Hubs/MultiplayerQueue.cs
smoogipoo: > drop this check altogether

I agree with that also.

> Also not sure that the client should be authoritative with the display of the delete button

Is there a reason why the server should be authoritative on everything? How far do you go with this - should there also be a flag telling the client whether they can kick participants rather than the host check? Should there be a flag telling the client whether they can add items to the queue rather than the queue mode + host check? Should there be a flag telling the client whether they can ready up and/or start the match based on the conditions currently in `MultiplayerReadyButton`?  
It feels like unnecessary complexity to me.

And to illustrate the complexity, if it were to happen, how would you envision such a flag being passed to clients? Would there be a dictionary in `MultiplayerPlaylistItem` mapping each user id in the room to whether they can delete the item? Or would there be a flag in `MultiplayerSettings` telling all clients whether items _may be possible to be_ deleted, but then the clients would still have to do their own `OwnerID` check?

smoogipoo:
I can allow hosts to always delete items, if you think that makes more sense?

peppy:
I think so, yes. In the case where host would be disallowed to remove, I think the room would be in a special state where the host is delegated to an api owner or banchobot or something.

osu.Server.Spectator/Hubs/MultiplayerQueue.cs
peppy: The answer is "no" to all the above. The reason for it here was that the conditionals here are (were?) quite complicated.

The objective was to simplify things, not complicate. The flag would be sent as part of the `MultiplayerPlaylistItem`, not externally.

smoogipoo: If the objective were to simplify, then this is not it. Instead of 4 lines in one class, one of which was redundant, and which is reduced to 2 in my upcoming changes, you'd have multiple lines across `MultiplayerRoom`, `MultiplayerClient`, `PlaylistItem`, `TestMultiplayerClient`, and `DrawableRoomPlaylistItem` just to get such a flag to work.

So I disagree with that path.

smoogipoo:
Applied requested changes re: host being able to always remove items, and allowing removal in host-only mode. Seems to work nicely.

osu.Server.Spectator/Hubs/MultiplayerQueue.cs
```diff
@@ -151,6 +151,38 @@ public async Task AddItem(MultiplayerPlaylistItem item, MultiplayerRoomUser user
...
+        {
+            if (dbFactory == null) throw new InvalidOperationException($"Call {nameof(Initialise)} first.");
+
+            var item = room.Playlist.FirstOrDefault(item => item.ID == playlistItemId);
+
+            if (item == null)
+                throw new InvalidStateException("Item does not exist in the room.");
+
+            if (item == CurrentItem)
+                throw new InvalidStateException("The room's current item cannot be removed.");
```
peppy, line 169: Is there a technical reason for this limitation? Thinking we should allow it is there's nothing stopping us from doing so. If it's only here for the "host only" mode, we should limit it to that.

smoogipoo: Nothing handles this gracefully right now...

- `updateCurrentItem` assumes there's always one item in the room, expired or otherwise.
- `MultiplayerRoomSettings.PlaylistItemId` is non-nullable.
- "Adding" an item in host-only mode re-uses/updates the current item.
- Client side doesn't gracefully handle this either. Not only the same issues as above but some callbacks like `SettingsChanged` expect to have the `PlaylistItemId` given in the playlist.

I would like to fix that in a future PR, but I just don't feel safe with allowing the current item to be removed with the current state of things.

peppy: Okay, let's leave it for next time.

peppy (approved):

PR #94, opened 2021-12-09
Fix compile error in sample client

smoogipoo:
Prereqs:
- [x] https://github.com/ppy/osu/pull/16010
- [x] Nuget package update

Thought I'd committed this. Missed because I don't ever compile/use this project, and neither does CI apparently.


smoogipoo:
Actually, this is the one that's going to have a prereq on https://github.com/ppy/osu/pull/16010

PR #95, opened 2021-12-09
Add support for enqueueing items in host-only mode, and changing arbitrary playlist items

smoogipoo:
Prereqs:
- [x] https://github.com/ppy/osu-server-spectator/pull/94 (this PR includes it to not merge conflict)
- [x] https://github.com/ppy/osu/pull/16020

This adds support for:
- The host being able to add items to the queue, rather than only being able to change the existing one.
- The host being able to change items in all modes.
- Guests being able to change their own items in non-host-only mode. This is not yet supported client side (just lacking design implementation), but is a work item in https://github.com/ppy/osu/pull/15640.

Some tests have been moved from the host-only queue test class to the more broad queue tests, which accounts for a bulk of the test changes.

### DIFF
--- a/osu.Server.Spectator/Hubs/MultiplayerQueue.cs
+++ b/osu.Server.Spectator/Hubs/MultiplayerQueue.cs
@@ -106,10 +106,12 @@ namespace osu.Server.Spectator.Hubs
         {
             if (dbFactory == null) throw new InvalidOperationException($"Call {nameof(Initialise)} first.");
 
+            bool isNewAddition = item.ID == 0;
+
             if (room.Settings.QueueMode == QueueMode.HostOnly && !user.Equals(room.Host))
                 throw new NotHostException();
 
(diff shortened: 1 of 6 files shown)


osu.Server.Spectator/Hubs/MultiplayerQueue.cs
```diff
@@ -106,10 +106,12 @@ public async Task AddItem(MultiplayerPlaylistItem item, MultiplayerRoomUser user
         {
             if (dbFactory == null) throw new InvalidOperationException($"Call {nameof(Initialise)} first.");
 
+            bool isNewAddition = item.ID == 0;
```
smoogipoo, line 109: I did this because it leaves osu! backwards compatible with the currently-deployed osu-server-spectator, but I agree this may be confusing.

Let me know if it'd be preferred to add a separate `ChangeItem` path, at the expense of breaking this compatibility.

smoogipoo: Now separated into two methods.

peppy (approved):

PR #96, opened 2021-12-11
Unready all users on relevant playlist item changes

smoogipoo:
- Unready all users when the current item is edited.
- Prevent users readying up with all items expired.


peppy:
How did you find these ones? Reported somewhere?

smoogipoo:
Nope, I just thought about them.

peppy (approved):

PR #97, opened 2021-12-14
Require users to use AbortGameplay() to transition to idle during gameplay

smoogipoo:
Prereqs:
- [x] https://github.com/ppy/osu/pull/16075

### DIFF
--- a/osu.Server.Spectator/Hubs/MultiplayerHub.cs
+++ b/osu.Server.Spectator/Hubs/MultiplayerHub.cs
@@ -306,7 +306,7 @@ namespace osu.Server.Spectator.Hubs
 
                 Log(room, $"User changing state from {user.State} to {newState}");
 
-                if (!ensureValidStateSwitch(room, user.State, newState))
+                if (!ensureValidStateSwitch(room, user.State, newState, false))
                     return;
 
                 user.State = newState;
@@ -446,6 +446,9 @@ namespace osu.Server.Spectator.Hubs
                 var user = room.Users.FirstOrDefault(u => u.UserID == CurrentContextUserId);
                 if (user == null)
                     throw new InvalidOperationException("Local user was not found in the expected room");
+
+                if (!ensureValidStateSwitch(room, user.State, MultiplayerUserState.Idle, true))
+                    return;
 
                 await changeAndBroadcastUserState(room, user, MultiplayerUserState.Idle);
                 await updateRoomStateIfRequired(room);
@@ -713,14 +716,25 @@ namespace osu.Server.Spectator.Hubs
         /// <param name="room">The room.</param>
         /// <param name="oldState">The old state.</param>
         /// <param name="newState">The new state.</param>
-        private bool ensureValidStateSwitch(ServerMultiplayerRoom room, MultiplayerUserState oldState, MultiplayerUserState newState)
+        private bool ensureValidStateSwitch(ServerMultiplayerRoom room, MultiplayerUserState oldState, MultiplayerUserState newState, bool isAbortingGameplay)
         {
             switch (newState)
             {
                 case MultiplayerUserState.Idle:
-                    if (oldState == MultiplayerUserState.WaitingForLoad || oldState == MultiplayerUserState.Loaded || oldState == MultiplayerUserState.Playing)
+                    bool oldStateIsGameplay = oldState == MultiplayerUserState.WaitingForLoad || oldState == MultiplayerUserState.Loaded || oldState == MultiplayerUserState.Playing;
+
+                    if (isAbortingGameplay)
+                    {
+                        if (oldStateIsGameplay)
+                            break;
+
+                        // AbortGameplay() should not be used to transition into the idle state from non-gameplay states.
+                        throw new InvalidStateException("Cannot abort gameplay in a non-gameplay state.");
+                    }
+                    else if (oldStateIsGameplay)
                     {
                         // while playing, users can only transition to idle using AbortGameplay().
+                        // silently ignore an attempt to transition into the idle state via other means.
                         return false;
                     }
 
--- a/osu.Server.Spectator/Hubs/MultiplayerHub.cs
+++ b/osu.Server.Spectator/Hubs/MultiplayerHub.cs
@@ -305,7 +305,10 @@ namespace osu.Server.Spectator.Hubs
                     return;
 
                 Log(room, $"User changing state from {user.State} to {newState}");
-                ensureValidStateSwitch(room, user.State, newState);
+
+                if (!ensureValidStateSwitch(room, user.State, newState))
+                    return;
+
                 user.State = newState;
 
                 // handle whether this user should be receiving gameplay messages or not.
@@ -428,6 +431,24 @@ namespace osu.Server.Spectator.Hubs
                 await changeRoomState(room, MultiplayerRoomState.WaitingForLoad);
 
                 await Clients.Group(GetGroupId(room.RoomID, true)).LoadRequested();
+            }
+        }
+
+        public async Task AbortLoad()
+        {
+            using (var userUsage = await GetOrCreateLocalUserState())
+            using (var roomUsage = await getLocalUserRoom(userUsage.Item))
+            {
+                var room = roomUsage.Item;
+                if (room == null)
+                    throw new InvalidOperationException("Attempted to operate on a null room");
+
+                var user = room.Users.FirstOrDefault(u => u.UserID == CurrentContextUserId);
+                if (user == null)
+                    throw new InvalidOperationException("Local user was not found in the expected room");
+
+                await changeAndBroadcastUserState(room, user, MultiplayerUserState.Idle);
+                await updateRoomStateIfRequired(room);
             }
         }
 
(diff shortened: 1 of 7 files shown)


osu.Server.Spectator/Hubs/MultiplayerHub.cs
```diff
@@ -431,6 +434,24 @@ public async Task StartMatch()
...
+            {
+                var room = roomUsage.Item;
+                if (room == null)
+                    throw new InvalidOperationException("Attempted to operate on a null room");
+
+                var user = room.Users.FirstOrDefault(u => u.UserID == CurrentContextUserId);
+                if (user == null)
+                    throw new InvalidOperationException("Local user was not found in the expected room");
+
+                await changeAndBroadcastUserState(room, user, MultiplayerUserState.Idle);
```
peppy, line 450: Is this intentionally not doing any pre-checks? ie. the client could call this from a `Ready` or `Results` state and it would still set them to idle. Not sure that's the behaviour we're looking for.

It also means the following like `updateRoomStateIfRequired` is running even if there is no state change. Looks like that won't be an issue currently, but it's different to how we've done it until now so not 100%.

smoogipoo: I didn't think it was that big of a deal, but fair enough. I've added validation.

osu.Server.Spectator/Hubs/MultiplayerHub.cs
```diff
@@ -692,12 +716,29 @@ private async Task changeRoomState(ServerMultiplayerRoom room, MultiplayerRoomSt
         /// <param name="room">The room.</param>
         /// <param name="oldState">The old state.</param>
         /// <param name="newState">The new state.</param>
-        private void ensureValidStateSwitch(ServerMultiplayerRoom room, MultiplayerUserState oldState, MultiplayerUserState newState)
+        private bool ensureValidStateSwitch(ServerMultiplayerRoom room, MultiplayerUserState oldState, MultiplayerUserState newState, bool isAbortingGameplay)
```
peppy, line 719: This method is getting really weird, with a `bool` return, a potential exception throw, and an extra argument for a sub-state change. How does the following look instead:

```diff
diff --git a/osu.Server.Spectator.Tests/Multiplayer/UserStateManagementTests.cs b/osu.Server.Spectator.Tests/Multiplayer/UserStateManagementTests.cs
index 355e337..0d43174 100644
--- a/osu.Server.Spectator.Tests/Multiplayer/UserStateManagementTests.cs
+++ b/osu.Server.Spectator.Tests/Multiplayer/UserStateManagementTests.cs
@@ -262,13 +262,13 @@ namespace osu.Server.Spectator.Tests.Multiplayer
             await Hub.StartMatch();
 
             // Test during WaitingForLoad state.
-            await Hub.ChangeState(MultiplayerUserState.Idle);
+            await Assert.ThrowsAsync<InvalidStateException>(() => Hub.ChangeState(MultiplayerUserState.Idle));
             using (var room = await Rooms.GetForUse(ROOM_ID))
                 Assert.Equal(MultiplayerUserState.WaitingForLoad, room.Item?.Users[0].State);
 
             // Test during Playing state.
             await Hub.ChangeState(MultiplayerUserState.Loaded);
-            await Hub.ChangeState(MultiplayerUserState.Idle);
+            await Assert.ThrowsAsync<InvalidStateException>(() => Hub.ChangeState(MultiplayerUserState.Idle));
             using (var room = await Rooms.GetForUse(ROOM_ID))
                 Assert.Equal(MultiplayerUserState.Playing, room.Item?.Users[0].State);
 
diff --git a/osu.Server.Spectator/Hubs/MultiplayerHub.cs b/osu.Server.Spectator/Hubs/MultiplayerHub.cs
index f780a91..0aa17ae 100644
--- a/osu.Server.Spectator/Hubs/MultiplayerHub.cs
+++ b/osu.Server.Spectator/Hubs/MultiplayerHub.cs
@@ -306,8 +306,7 @@ namespace osu.Server.Spectator.Hubs
 
                 Log(room, $"User changing state from {user.State} to {newState}");
 
-                if (!ensureValidStateSwitch(room, user.State, newState, false))
-                    return;
+                ensureValidStateSwitch(room, user.State, newState);
 
                 user.State = newState;
 
@@ -447,8 +446,8 @@ namespace osu.Server.Spectator.Hubs
                 if (user == null)
                     throw new InvalidOperationException("Local user was not found in the expected room");
 
-                if (!ensureValidStateSwitch(room, user.State, MultiplayerUserState.Idle, true))
-                    return;
+                if (!isGameplayState(user.State))
+                    throw new InvalidStateException("Cannot abort gameplay in a non-gameplay state.");
 
                 await changeAndBroadcastUserState(room, user, MultiplayerUserState.Idle);
                 await updateRoomStateIfRequired(room);
@@ -716,29 +715,15 @@ namespace osu.Server.Spectator.Hubs
         /// <param name="room">The room.</param>
         /// <param name="oldState">The old state.</param>
         /// <param name="newState">The new state.</param>
-        private bool ensureValidStateSwitch(ServerMultiplayerRoom room, MultiplayerUserState oldState, MultiplayerUserState newState, bool isAbortingGameplay)
+        private void ensureValidStateSwitch(ServerMultiplayerRoom room, MultiplayerUserState oldState, MultiplayerUserState newState)
         {
             switch (newState)
             {
                 case MultiplayerUserState.Idle:
-                    bool oldStateIsGameplay = oldState == MultiplayerUserState.WaitingForLoad || oldState == MultiplayerUserState.Loaded || oldState == MultiplayerUserState.Playing;
-
-                    if (isAbortingGameplay)
-                    {
-                        if (oldStateIsGameplay)
-                            break;
-
-                        // AbortGameplay() should not be used to transition into the idle state from non-gameplay states.
+                    if (isGameplayState(oldState))
                         throw new InvalidStateException("Cannot abort gameplay in a non-gameplay state.");
-                    }
-                    else if (oldStateIsGameplay)
-                    {
-                        // while playing, users can only transition to idle using AbortGameplay().
-                        // silently ignore an attempt to transition into the idle state via other means.
-                        return false;
-                    }
 
-                    // any other state can return to idle.
+                    // any non-gameplay state can return to idle.
                     break;
 
                 case MultiplayerUserState.Ready:
@@ -783,8 +768,20 @@ namespace osu.Server.Spectator.Hubs
                 default:
                     throw new ArgumentOutOfRangeException(nameof(newState), newState, null);
             }
+        }
 
-            return true;
+        private static bool isGameplayState(MultiplayerUserState state)
+        {
+            switch (state)
+            {
+                default:
+                    return false;
+
+                case MultiplayerUserState.WaitingForLoad:
+                case MultiplayerUserState.Loaded:
+                case MultiplayerUserState.Playing:
+                    return true;
+            }
         }
 
         /// <summary>
```

smoogipoo: The first thing I noticed is that you're throwing the exception on `ChangeState` now, which should be silently ignored. It's not something that should result in a user error because it's entirely unpreventable.

peppy: That's the case for other state changes too, though. If we're going to make that change, none of these should be throwing.

One example is if the state is changed to `WaitingForLoad` but the user's client doesn't receive it in time, and sends a `NotReady` state change request.

I'm not heavily against removing the exception if it's a blocker on making the attached changes.

smoogipoo: > One example is if the state is changed to `WaitingForLoad` but the user's client doesn't receive it in time, and sends a `NotReady` state change request.

This is exactly the case being fixed in this PR, is it not? And, as far as I can tell it's the only (current) case, which is why I initially didn't handle every other gameplay state and only `WaitingForLoad`.

peppy: Another example would be a user pressing ready while they can still see a playlist item, but it's removed on the server already (exception thrown in `Ready` state change check).

The change in this PR is great because it establishes clear intent from the user to exit gameplay, meaning we're not in the future accidentally going to cause an exit from gameplay with a seemingly innocent `Idle` state change. I don't see it as just protecting from that state change alone.

smoogipoo: As far as I understand it, the exceptions are there to prevent completely invalid operations from being run if, for example, a malicious user sends arbitrary commands.
They're not there to tell the user "whoops! the laws on the universe prevented your request from arriving at the server in time, sorry about that". Is there a reason we should be doing that, which I'm missing?

In the same spirit, I would also agree to add the same silent ignore to the scenario you mentioned.

Is the bancho model of doing things (i.e. not having exceptions for multiplayer) an incorrect model to be using?

peppy: As I said above, I'd be fine with removing all exceptions from these state changes, except for the new `AbortGameplay` one, since the client is 100% on what it wants to do there, and sure of its current state. Although I could also agree with dropping that one to a silent ignore if you'd prefer.

smoogipoo: I agree with leaving exceptions in `AbortGameplay`, just not `ChangeState` from your above changes.

peppy: I've pushed some changes for now. I think we need to do some more refactoring around `ChangeState` but hopefully it's in an okay state for this PR? Please confirm you can live with it as is.

osu.Server.Spectator/Hubs/MultiplayerHub.cs
```diff
@@ -697,7 +725,10 @@ private void ensureValidStateSwitch(ServerMultiplayerRoom room, MultiplayerUserS
             switch (newState)
             {
                 case MultiplayerUserState.Idle:
-                    // any state can return to idle.
+                    if (isGameplayState(oldState))
+                        throw new InvalidStateException("Cannot abort gameplay in a non-gameplay state.");
```
smoogipoo, line 729: Message reads weird. Maybe something like `Cannot transition out of a gameplay state without aborting.`?

osu.Server.Spectator/Hubs/MultiplayerHub.cs
smoogipoo: Apart from the new comment I've left, the rest looks fine.

PR #99, opened 2022-02-14
Log total connections to a hub to statsd

peppy:
Had this lying around locally. Already using on production for months..

### DIFF
--- a/osu.Server.Spectator/Hubs/StatefulUserHub.cs
+++ b/osu.Server.Spectator/Hubs/StatefulUserHub.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Threading;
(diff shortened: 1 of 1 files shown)


osu.Server.Spectator/Hubs/StatefulUserHub.cs
```diff
@@ -24,6 +26,9 @@ public abstract class StatefulUserHub<TClient, TUserState> : Hub<TClient>, ILogT
 
         private readonly Logger logger;
 
+        // ReSharper disable once StaticMemberInGenericType
```
peppy, line 29: Open to a better solution if this is seen as something we don't want. Keep in mind that I intentionally want the `static` counter here to be local to each generic type.

smoogipoo (approved):

PR #103, opened 2022-02-18
Fix multiple tests not `await`ing on async actions

peppy:
These were showing up as warnings. Required a bit more effort than expected to repair (turns out they were silently failing until now with unobserved exceptions).


smoogipoo (approved):
Interesting, didn't realise/intend for this `Initialise` call to be done so many times...

PR #105, opened 2022-02-28
Add basics of replay storage support to `SpectatorHub`

peppy:
This currently stores all replays to the filesystem. The intention is to test this out for a while and make sure that things generally work, then extend this to upload the replays to a more persistent storage (probably S3).

- [x] Depends on https://github.com/ppy/osu/pull/17017.
- [x] Depends on game package bump

Note that I've intentionally left error handling missing from this. I kinda want to deploy and see how things fail to get a better understanding of exactly what cases we need to handle with SignalR (still not 100% on what guarantees we have).


osu.Server.Spectator/Hubs/SpectatorHub.cs
```diff
@@ -1,27 +1,78 @@
...
+                if (state.RulesetID == null) throw new ArgumentNullException(nameof(state.RulesetID));
+                if (state.BeatmapID == null) throw new ArgumentNullException(nameof(state.BeatmapID));
+
+                using (var db = databaseFactory.GetInstance())
+                {
+                    string? beatmapChecksum = await db.GetBeatmapChecksumAsync(state.BeatmapID.Value);
+                    string? username = await db.GetUsernameAsync(CurrentContextUserId);
+
+                    if (string.IsNullOrEmpty(beatmapChecksum))
+                        throw new ArgumentException(nameof(state.BeatmapID));
```
bdach, line 55: Probably would be a good idea to also send the checksum from the client and verify that it matches as well, but I guess it is not a huge priority for now?

osu.Server.Spectator/Hubs/SpectatorHub.cs
```diff
@@ -30,13 +81,44 @@ public async Task BeginPlaySession(SpectatorState state)
...
         {
-            await Clients.Group(GetGroupId(CurrentContextUserId)).UserSentFrames(CurrentContextUserId, data);
+            using (var usage = await GetOrCreateLocalUserState())
+            {
+                var score = usage.Item?.Score;
+                Debug.Assert(score != null);
+
+                score.ScoreInfo.Statistics = data.Header.Statistics;
+                score.ScoreInfo.MaxCombo = data.Header.MaxCombo;
+                score.ScoreInfo.Combo = data.Header.Combo;
```
bdach, line 91: Just going to note that the legacy encoder will also read `TotalScore` and write it to the score, and I'm not seeing it be set anywhere in this diff, which means that in some cases it will be 0 rather than the actual value. It may be an issue if the replay is to ever be imported into the client in some way, possibly irrelevant for now.

osu.Server.Spectator/Hubs/SpectatorHub.cs
```diff
@@ -30,13 +81,44 @@ public async Task BeginPlaySession(SpectatorState state)
...
+                var now = DateTimeOffset.UtcNow;
+
+                userScore.ScoreInfo.Date = now;
+                var legacyEncoder = new LegacyScoreEncoder(userScore, null);
+
+                string path = Path.Combine(REPLAYS_PATH, now.Year.ToString(), now.Month.ToString(), now.Day.ToString());
+
+                Directory.CreateDirectory(path);
+
+                string filename = $"{now.ToUnixTimeSeconds()}-{CurrentContextUserId}-{userScore.ScoreInfo.BeatmapInfo.OnlineID}.osr";
```
bdach, line 114: This filename construction is worrying me a bit there for several reasons:

* Primarily, no theoretically-unique component like a guid that would substantially decrease possibility of naming clashes/overwrites
* I'm also not sure about the second granularity of the timestamp (rather than something like msecs or even ticks), or the lack of a ruleset ID next to the beatmap ID (think converts)

peppy: It’s all just a temporary thing, but adding ruleset sounds okay.

For a clash to happen the same user would have to play and submit twice in a single second. I’m not sure that should be possible.

osu.Server.Spectator/Hubs/SpectatorHub.cs
peppy: I did notice this, but I think this is best left for a separate effort.

osu.Server.Spectator/Hubs/SpectatorHub.cs
peppy: Apparently this isn't possible right now because we need to know the max combo of a beatmap to calculate score. Could potentially be retrieved from the database on the fly here, but not sure about it yet. I'll add a TODO.

This is less relevant for lazer because it will calculate the total score post replay loading, and I'm not sure yet if we even want these replays to support loading into stable (where it would be used).

bdach (commented):
Tested this today, the happy path appears to work. One slight issue with the unhappy path - unsure if you want to address it here or not.

osu.Server.Spectator/Hubs/SpectatorHub.cs
```diff
@@ -30,13 +81,46 @@ public async Task BeginPlaySession(SpectatorState state)
 
         public async Task SendFrameData(FrameDataBundle data)
         {
-            await Clients.Group(GetGroupId(CurrentContextUserId)).UserSentFrames(CurrentContextUserId, data);
+            using (var usage = await GetOrCreateLocalUserState())
+            {
+                var score = usage.Item?.Score;
+                Debug.Assert(score != null);
```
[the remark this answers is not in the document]
peppy: Good catch. I've changed this to silently return rather than throw.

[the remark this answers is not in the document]
bdach: This is partially but not fully fixed with [1606f08](https://github.com/ppy/osu-server-spectator/pull/105/commits/1606f08584c4e07e5aec4e37823e9384b5a96834). As mentioned above `EndPlaySession()` also has a similar issue:

https://github.com/ppy/osu-server-spectator/blob/ce1bc3fc96d6f456793241058d537f8ab32c6d7f/osu.Server.Spectator/Hubs/SpectatorHub.cs#L109-L113

The date set will crash if one of the sanity checks in `BeginPlaySession()` fail.

smoogipoo:
Just to confirm, is the intention here to store _all_ replays? Since we now have the `SpectatedUserState` enum, we can only save those which have passed/failed.

peppy:
Yeah, as a starting point the thought is to store all. I'm in discussions with nanaya about how we want to go forward with replay storage, but nothing has been decided yet.

smoogipoo:
Hmm, okay. I'm more kinda worried about people holding down the retry button, but I'll let you worry about disk space usage and merge as is.

smoogipoo (commented):
Lgtm

smoogipoo (approved):

PR #106, opened 2022-03-17
Handle unplayable mods and invalid free mods during playlist item mutation

frenzibyte:
- [x] Depends on https://github.com/ppy/osu/pull/17270
- [x] Depends on game and rulesets bump

Have only relied on the test cases I wrote to confirm whether this is working properly or not, don't have the ability to run local servers and confirm this on my current machine yet.


smoogipoo (approved):

smoogipoo:
@nanaya We're going to need a way to validate mods (in osu-web) for their validity in multiplayer as both a room-wide "required" mod, or a per-user "free" mod. Correct me if I'm wrong, but I don't believe we have the functionality to do this yet?

nanaya:
There's already check for mods when creating multiplayer playlist item. Just need to be extended with whatever missing checks.

smoogipoo:
Roger, made an issue since it looks like a bit of an involved implementation: https://github.com/ppy/osu-web/issues/8945

PR #108, opened 2022-03-18
Fix users not removed from gameplay group after aborting gameplay

smoogipoo:
Fixes https://github.com/ppy/osu/issues/17321

When exiting gameplay via `AbortGameplay()`, users were not being removed from the gameplay group.

Testing is as follows:
1. Join two users.
2. Make both users ready.
3. Start match.
4. Exit match with both users.
5. Make only the host ready.
6. Start match. Both users will start even though the second was not ready. The second user will quit back to the subscreen and show an exception.

This is a chonky PR mainly because I've added support to be able to assert calls on individual users rather than a "general" receiver for the gameplay/non-gameplay groups. This is done by:
- Storing the movement of clients around calls to `AddToGroupAsync()`/`RemoveFromGroupAsync()`.
- Creating a new `DelegatingMultiplayerClient` that receives the clients for a particular group. This new object now becomes the target type of the `Receiver`/`GameplayReceiver`/`Receiver2` mock objects.
- Adding new `UserReceiver`/`User2Receiver` mock objects for assertions on the individual users.

### DIFF
--- a/osu.Server.Spectator.Tests/Multiplayer/MultiplayerTest.cs
+++ b/osu.Server.Spectator.Tests/Multiplayer/MultiplayerTest.cs
@@ -35,13 +35,41 @@ namespace osu.Server.Spectator.Tests.Multiplayer
         protected EntityStore<MultiplayerClientState> UserStates { get; }
 
         protected readonly Mock<IDatabaseFactory> DatabaseFactory;
-
         protected readonly Mock<IDatabaseAccess> Database;
 
-        protected readonly Mock<IMultiplayerClient> Receiver;
-        protected readonly Mock<IMultiplayerClient> GameplayReceiver;
+        /// <summary>
+        /// A general non-gameplay receiver for the room with ID <see cref="ROOM_ID"/>.
+        /// </summary>
+        protected readonly Mock<DelegatingMultiplayerClient> Receiver;
 
+        /// <summary>
+        /// A general gameplay receiver for the room with ID <see cref="ROOM_ID"/>.
+        /// </summary>
+        protected readonly Mock<DelegatingMultiplayerClient> GameplayReceiver;
+
+        /// <summary>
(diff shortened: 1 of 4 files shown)


osu.Server.Spectator.Tests/Multiplayer/MultiplayerTest.cs
```diff
@@ -35,26 +35,57 @@ public abstract class MultiplayerTest
...
         protected readonly Mock<IDatabaseFactory> DatabaseFactory;
-
         protected readonly Mock<IDatabaseAccess> Database;
 
-        protected readonly Mock<IMultiplayerClient> Receiver;
-        protected readonly Mock<IMultiplayerClient> GameplayReceiver;
+        /// <summary>
+        /// A general non-gameplay receiver for the room with ID <see cref="ROOM_ID"/>.
+        /// </summary>
+        protected readonly Mock<DelegatingMultiplayerClient> Receiver;
```
bdach, line 43: This is ballooning a bit now in size. I think I'd prefer to keep this more compact/potentially flexible via some dictionary usage:

```csharp
protected readonly Dictionary<int, DelegatingMultiplayerClient> RoomGeneralReceivers;
protected readonly Dictionary<int, DelegatingMultiplayerClient> RoomGameplayReceivers;
protected readonly Dictionary<int, IMultiplayerClient> UserReceivers;
```

Unsure if out of scope to change this here. Also tempted to fold `ContextUser`s in a similar way but that is definitely out of scope here.

bdach (commented):
Fix seems sane as far as I can tell, just 1 suggestion re: test

osu.Server.Spectator.Tests/Multiplayer/MultiplayerTest.cs
smoogipoo: Not going to do that in this PR because this isn't a refactoring PR.

bdach (approved):
fair enough, seems to work

peppy (approved):

PR #109, opened 2022-03-18
Implement multiplayer countdown timers

smoogipoo:
Prereqs:
- [x] https://github.com/ppy/osu/pull/17326

`CountdownImplementation` follows the same general idea as `MatchTypeImplementation`. The important part to keep in mind is that the countdown is run on a background thread, and interactions between it and the `MultiplayerHub` are synchronised through `hub.GetRoom()` (i.e. the `EntityStore`).

After discussion with @peppy, I've added one hack here for the time being. It is [incorrect](https://docs.microsoft.com/en-us/aspnet/core/signalr/hubs?view=aspnetcore-6.0#create-and-use-hubs) to store `MultiplayerHub` as it is a transient object, however we're doing exactly this through the `IMultiplayerServerMatchCallbacks` interface. It only becomes an issue if you use it off-thread, such as with countdown timers.

The hack is that, when the transient hub is disposed, its `Clients` object becomes inaccessible due to a disposal check in the base class. For the time being, I'm manually saving this `Clients` object to bypass this check since it doesn't seem to cause any issue among my test cases.

I will be working to refactor everything to use `HubContext` as described by https://docs.microsoft.com/en-us/aspnet/core/signalr/hubcontext?view=aspnetcore-6.0, which is the "proper" way of going about this. The implementation to do so is a little bit involved for now.


osu.Server.Spectator.Tests/Multiplayer/TestMultiplayerHub.cs
```diff
@@ -15,7 +15,7 @@ public TestMultiplayerHub(IDistributedCache cache, EntityStore<ServerMultiplayer
         {
         }
 
-        public ItemUsage<ServerMultiplayerRoom> GetRoom(long roomId) => Rooms.GetForUse(roomId).Result;
+        public new ItemUsage<ServerMultiplayerRoom> GetRoom(long roomId) => Rooms.GetForUse(roomId).Result;
```
peppy, line 18: Can this be renamed to `GetRoomSynchronous` to avoid any confusion? `new` hiding always has me on edge.

osu.Server.Spectator/Hubs/CountdownImplementation.cs
```diff
@@ -0,0 +1,119 @@
...
+            var finishSource = countdownFinishSource = new CancellationTokenSource();
+            var cancellationSource = CancellationTokenSource.CreateLinkedTokenSource(stopSource.Token, finishSource.Token);
+
+            Task lastCountdownTask = countdownTask;
+            countdownTask = start();
+
+            // Some sections in the following code require single-threaded execution mode. This is achieved by re-retrieving the room from the hub, forcing an exclusive lock on it.
+            async Task start()
+            {
+                // 1. Wait for the last countdown.
```
peppy, line 49: Not sure about the numbering here, probably remove.

peppy, line 49: Not sure about the numbering here, probably remove.

osu.Server.Spectator/Hubs/MultiplayerHub.cs
```diff
@@ -916,12 +926,55 @@ public async Task OnMatchSettingsChanged(ServerMultiplayerRoom room)
             await Clients.Group(GetGroupId(room.RoomID)).SettingsChanged(room.Settings);
         }
 
+        public async Task<ItemUsage<ServerMultiplayerRoom>> GetRoom(long roomId) => await Rooms.GetForUse(roomId);
+
+        public async Task InternalStartMatch(ServerMultiplayerRoom room)
```
peppy, line 931:
```suggestion
        internal async Task InternalStartMatch(ServerMultiplayerRoom room)
```

This will need to be marked `internal` else it can be called by a user, which we probably don't want. Same goes for `GetRoom` above. (will need to confirm that `internal` methods are not exposed via SignalR - i'm just assuming so here).

peppy, line 931:
```suggestion
        internal async Task InternalStartMatch(ServerMultiplayerRoom room)
```

This will need to be marked `internal` else it can be called by a user, which we probably don't want. Same goes for `GetRoom` above. (will need to confirm that `internal` methods are not exposed via SignalR - i'm just assuming so here).

osu.Server.Spectator/Hubs/CountdownImplementation.cs
```diff
@@ -0,0 +1,119 @@
...
+using System.Threading.Tasks;
+using osu.Game.Online.Multiplayer;
+using osu.Game.Online.Multiplayer.Countdown;
+
+namespace osu.Server.Spectator.Hubs
+{
+    /// <summary>
+    /// The implementation for a countdown timer.
+    /// </summary>
+    public class CountdownImplementation
```
peppy, line 15: I find it a touch weird that this is an `Implementation`, since that term is generally one we have used for the case where there are going to be more than one implementations.

I understand why this can't be a `ServerCountdown : Countdown` in a similar fashion to how `ServerMultiplayerRoom` is structured, but still not sure about the naming. `ServerCountdown` may fit better?

peppy, line 15: I find it a touch weird that this is an `Implementation`, not a `ServerCountdown : Countdown` in a similar fashion to how `ServerMultiplayerRoom` is structured. Is there reasoning behind this? ie. are there planned to be different types of countdowns in the future?

osu.Server.Spectator/Hubs/CountdownImplementation.cs
```diff
@@ -0,0 +1,119 @@
...
+                }
+
+                // 3. Run the countdown.
+                try
+                {
+                    await Task.Delay(countdown.EndTime - DateTimeOffset.Now, cancellationSource.Token).ConfigureAwait(false);
+                }
+                catch (OperationCanceledException)
+                {
+                }
```
peppy, line 80: How does this work? I would presume that if there's a cancellation we don't want the code following this step to run.

peppy, line 80: How does this work? I would presume that if there's a cancellation we don't want the code following this step to run.

osu.Server.Spectator.Tests/Multiplayer/MultiplayerCountdownTest.cs
```diff
@@ -0,0 +1,283 @@
...
+        [Fact]
+        public async Task GameplayDoesNotStartWhenCountdownCancelled()
+        {
+            await Hub.JoinRoom(ROOM_ID);
+            await Hub.ChangeState(MultiplayerUserState.Ready);
+
+            await Hub.SendMatchRequest(new StartMatchCountdownRequest { Delay = TimeSpan.FromMinutes(1) });
+            await Hub.SendMatchRequest(new StopCountdownRequest());
+
+            finishCountdown();
```
peppy, line 75: I don't think this should be here? Isn't the aim of this test to test the `StopCountdownRequest` part?

peppy, line 75: I don't think this should be here? Isn't the aim of this test to test the `StopCountdownRequest` part?

peppy (changes requested):

smoogipoo:
Above deleted comments were duped comments.

osu.Server.Spectator/Hubs/CountdownImplementation.cs
smoogipoo: Also, because it was mentioned in a dupe comment and I missed it, I don't really see how `Implementation` necessitates the need for "more than one implementation", and if you're going to bring up `TeamVersus` vs `MatchTypeImplementation` (base class) then I'd mention that I struggle with the naming and would prefer `TeamVersusImplementation`.

osu.Server.Spectator.Tests/Multiplayer/TestMultiplayerHub.cs
smoogipoo: I'd just removed this method since it's available in `MultiplayerHub` itself. Touches many files but I guess it's a simple change.

osu.Server.Spectator/Hubs/CountdownImplementation.cs
smoogipoo: I've added an inline comment here.

osu.Server.Spectator/Hubs/CountdownImplementation.cs
```diff
@@ -102,6 +110,20 @@ async Task start()
             }
         }
 
+        public async Task UpdateTimeRemaining()
+        {
+            using (var roomUsage = await hub.GetRoom(roomId))
```
peppy, line 115: I feel really weird about this reverse retrieval lock lookup thing.

Rather than having this `GetRoom` logic here (which feels pretty weird), how do you feel about making a method in `ServerMultiplayerRoom` like `UpdateTimings` which can perform this?

I almost feel like this whole class is only adding complexity and may be better to be integrated as part of `ServerMultiplayerRoom` itself..

osu.Server.Spectator/Hubs/MultiplayerHub.cs
```diff
@@ -140,6 +140,8 @@ public async Task<MultiplayerRoom> JoinRoomWithPassword(long roomId, string pass
                     }
                 }
 
+                await room.CountdownImplementation.UpdateTimeRemaining();
```
peppy, line 143: On a similar note to my above comment, I don't really like this call calling an `Update` methods on a property of the room, feels too deep?

osu.Server.Spectator.Tests/Multiplayer/TestMultiplayerHub.cs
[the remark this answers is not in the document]
smoogipoo: This doesn't exist anymore, I'm not sure how you commented on it.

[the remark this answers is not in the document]
peppy: Ignore this, must have been a lingering review comment I never submitted.

osu.Server.Spectator/Hubs/CountdownImplementation.cs
smoogipoo: Sure. I've merged `CountdownImplementation` into `ServerMultiplayerRoom` and changed it to `UpdateForRetrieval()` for if we want to use this in the future.

I originally didn't do it this way because:
1. I had more complex logic around the countdown `Task`.
2. The async method needs to re-retrieve the room, it can't access the local instance at all. This is a little less obvious now but I've commented it in the places where it's done so that hopefully it won't be missed if this is iterated on in the future. The other alternative is to make the method static but you'd have to pass in 8 parameters directly or via a struct.

osu.Server.Spectator/Hubs/ServerMultiplayerRoom.cs
```diff
@@ -74,6 +90,107 @@ private MatchTypeImplementation createTypeImplementation(MatchType type)
             }
         }
 
-        public readonly CountdownImplementation CountdownImplementation;
+        private CancellationTokenSource? countdownStopSource;
```
peppy, line 93: Probably `#region` this stuff

osu.Server.Spectator/Hubs/MultiplayerHub.cs
smoogipoo: Addressed in comment above.

osu.Server.Spectator/Hubs/ServerMultiplayerRoom.cs
```diff
@@ -72,5 +89,112 @@ private MatchTypeImplementation createTypeImplementation(MatchType type)
...
+            }
+        }
+
+        /// <summary>
+        /// Stops the current countdown. Its continuation will not run.
+        /// </summary>
+        public void StopCountdown() => countdownStopSource?.Cancel();
+
+        /// <summary>
+        /// Forces the current countdown to finish and run its continuation as soon as possible.
```
peppy, line 189: comment is still a big ambiguous (what does "finish" mean here? to me it means it's going to run the "start" call). How does this sound, if it matches the intention:

```suggestion
        /// Forces the current countdown to run its callback immediately, unless cancelled.
```

smoogipoo: It's not immediately. It's intended to read "as soon as possible". You can't assume the countdown is finished by the point this method returns.

peppy: That's fine to change then. important part to me is the "unless cancelled" and not using the word "finish".

smoogipoo: Reworded xmldoc and matched with osu-side changes.

peppy (approved):

PR #110, opened 2022-03-22
Add multiplayer auto-start countdown timer

smoogipoo:
Prereqs:
- [x] https://github.com/ppy/osu/pull/17402
- [x] https://github.com/ppy/osu-server-spectator/pull/109
- [x] https://github.com/ppy/osu-web/pull/8740


peppy:
I haven't looked through the code yet, but I'm a bit confused as to how this is intended to work. If the host isn't ready, should the countdown not start?

https://user-images.githubusercontent.com/191335/160071277-2f2d4178-a4da-451a-9fbc-ab34af94a0e1.mp4

I'd argue that it should probably start even when the host isn't ready, for cases like "host rotate" where the host may be afk?

smoogipoo:
Didn't notice the base PR wasn't merged yet, so I pushed some changes here that alter the behaviour of normal countdowns too as discussed in Discord. I can try moving them to the base PR upon request though it wouldn't be a pretty change.

peppy (approved):

peppy:
I've deployed this to the dev server for further testing.

PR #111, opened 2022-03-28
Fix incorrect group adjustments when unready-ing all users

smoogipoo:
Fixes https://github.com/ppy/osu/issues/17477

This is a request-for-comments PR because I don't know how to better do this given the limitations imposed by `EntityStore`.

The primary purpose is to fix users not being removed from groups when multiple users' states are adjusted at once. Once such way that this could happen is via the following process:
1. Join player 1 (host), player 2 (guest).
2. Ready up player 2.
3. Change room settings on player 1 (e.g. the queue mode). Player 2 will unready.
4. Start play _without_ ready-ing up player 2.
5. Player 2 will attempt to enter gameplay, but fail with the `AbortGameplay` exception.

The reason why this occurs is that `changeAndBroadcastUserState` only mutates groups for the current context's connection ID:
https://github.com/ppy/osu-server-spectator/blob/b18144df5da7c447051cf479ddc4cc02d489d9cc/osu.Server.Spectator/Hubs/MultiplayerHub.cs#L724-L736
Meaning that in the above sequence, player 2 never got removed from the gameplay group - only player 1 which triggered the change in match settings.

In order to fix this, I've passed a `MultiplayerClientState` to this `changeAndBroadcastUserState` method.
- This change ripples throughout the rest of the codebase, and makes it pretty unmaintainable.  
- Non-nullability is enforced at this point. The given `MultiplayerClientState` must match the given `MultiplayerRoomUser`.

Then, the `changeAndBroadcastAllUserStates` method has been added in order to ease the implementation of the former. This one serves to re-retrieve a `MultiplayerClientState` for the given users taking into account an optional existing `MultiplayerClientState` that has been retrieved from an upper level.

### DIFF
--- a/osu.Server.Spectator/Hubs/MultiplayerHub.cs
+++ b/osu.Server.Spectator/Hubs/MultiplayerHub.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Linq;
 using System.Threading.Tasks;
 using Microsoft.Extensions.Caching.Distributed;
@@ -314,7 +315,8 @@ namespace osu.Server.Spectator.Hubs
 
                 ensureValidStateSwitch(room, user.State, newState);
 
-                await changeAndBroadcastUserState(room, user, newState);
+                Debug.Assert(userUsage.Item != null);
+                await changeAndBroadcastUserState(room, user, userUsage.Item, newState);
 
                 // Signal newly-spectating users to load gameplay if currently in the middle of play.
                 if (newState == MultiplayerUserState.Spectating
@@ -323,7 +325,8 @@ namespace osu.Server.Spectator.Hubs
                     await Clients.Caller.LoadRequested();
                 }
 
-                await updateRoomStateIfRequired(room);
+                Debug.Assert(userUsage.Item != null);
+                await updateRoomStateIfRequired(room, userUsage.Item);
             }
         }
 
@@ -396,7 +399,7 @@ namespace osu.Server.Spectator.Hubs
                         if (room.Settings.AutoStartEnabled)
                             throw new InvalidStateException("Cannot start manual countdown if auto-start is enabled.");
 
-                        room.StartCountdown(new MatchStartCountdown { TimeRemaining = countdown.Duration }, InternalStartMatch);
+                        room.StartCountdown(new MatchStartCountdown { TimeRemaining = countdown.Duration }, r => InternalStartMatch(r, null));
 
                         break;
 
@@ -435,7 +438,8 @@ namespace osu.Server.Spectator.Hubs
                 if (readyUsers.Length == 0)
                     throw new InvalidStateException("Can't start match when no users are ready.");
 
-                await InternalStartMatch(room);
+                Debug.Assert(userUsage.Item != null);
+                await InternalStartMatch(room, userUsage.Item);
             }
         }
 
@@ -455,8 +459,9 @@ namespace osu.Server.Spectator.Hubs
                 if (!isGameplayState(user.State))
                     throw new InvalidStateException("Cannot abort gameplay while not in a gameplay state");
 
-                await changeAndBroadcastUserState(room, user, MultiplayerUserState.Idle);
-                await updateRoomStateIfRequired(room);
+                Debug.Assert(userUsage.Item != null);
+                await changeAndBroadcastUserState(room, user, userUsage.Item, MultiplayerUserState.Idle);
+                await updateRoomStateIfRequired(room, userUsage.Item);
             }
         }
 
@@ -474,10 +479,14 @@ namespace osu.Server.Spectator.Hubs
                     throw new InvalidOperationException("Local user was not found in the expected room");
 
                 Log(room, $"Adding playlist item for beatmap {item.BeatmapID}");
-                await room.Queue.AddItem(item, user);
+
+                Debug.Assert(userUsage.Item != null);
+                await room.Queue.AddItem(item, user, userUsage.Item);
+
                 Log(room, $"Item ID {item.ID} added at slot {room.Queue.UpcomingItems.TakeWhile(i => i != item).Count() + 1} (of {room.Playlist.Count})");
 
-                await updateRoomStateIfRequired(room);
+                Debug.Assert(userUsage.Item != null);
+                await updateRoomStateIfRequired(room, userUsage.Item);
             }
         }
 
@@ -495,7 +504,9 @@ namespace osu.Server.Spectator.Hubs
                     throw new InvalidOperationException("Local user was not found in the expected room");
 
                 Log(room, $"Editing playlist item {item.ID} for beatmap {item.BeatmapID}");
-                await room.Queue.EditItem(item, user);
+
+                Debug.Assert(userUsage.Item != null);
+                await room.Queue.EditItem(item, user, userUsage.Item);
             }
         }
 
@@ -513,9 +524,12 @@ namespace osu.Server.Spectator.Hubs
                     throw new InvalidOperationException("Local user was not found in the expected room");
 
                 Log(room, $"Removing playlist item {playlistItemId}");
-                await room.Queue.RemoveItem(playlistItemId, user);
 
-                await updateRoomStateIfRequired(room);
+                Debug.Assert(userUsage.Item != null);
+                await room.Queue.RemoveItem(playlistItemId, user, userUsage.Item);
+
+                Debug.Assert(userUsage.Item != null);
+                await updateRoomStateIfRequired(room, userUsage.Item);
             }
         }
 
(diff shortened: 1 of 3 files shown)


peppy:
To confirm, at a high level the main issue here is that to access the `userId,connectionId` mapping locking is required on each user?

I'm wondering if we could potentially move access of these mappings outside of a locked context. We might be making things more complex for ourselves by requiring this locking for the simple case of mapping lookup (where an incorrect lookup is not a critical failure – if a mapping is found but the user has already disconnected then i believe it would be a noop for adding/removing from a group, and connection IDs should be considered unique and distinct).

osu.Server.Spectator/Hubs/MultiplayerHub.cs
```diff
@@ -702,36 +717,73 @@ private async Task updateRoomStateIfRequired(ServerMultiplayerRoom room)
...
+        /// <param name="state">The new user state.</param>
+        /// <param name="existingClient">Any existing <see cref="MultiplayerClientState"/> usage.
+        /// Must not be <c>null</c> if a user has been retrieved via <see cref="StatefulUserHub{TClient,TUserState}.UserStates"/> or
+        /// <see cref="StatefulUserHub{TClient,TUserState}.GetOrCreateLocalUserState"/> prior to this call.</param>
+        private async Task changeAndBroadcastAllUserStates(ServerMultiplayerRoom room, MultiplayerRoomUser[] users, MultiplayerUserState state, MultiplayerClientState? existingClient)
         {
+            foreach (var u in users)
+            {
+                if (u.UserID == existingClient?.UserId)
+                    await changeAndBroadcastUserState(room, u, existingClient, state);
```
peppy, line 748: This is only required to avoid retrieving a nested usage and causing a deadlock, correct? I'd add an inline comment explaining this.

osu.Server.Spectator/Hubs/MultiplayerHub.cs
```diff
@@ -702,36 +717,73 @@ private async Task updateRoomStateIfRequired(ServerMultiplayerRoom room)
...
+                else
+                {
+                    using (var userUsage = await UserStates.GetForUse(u.UserID))
+                    {
+                        if (userUsage.Item != null)
+                            await changeAndBroadcastUserState(room, u, userUsage.Item, state);
+                    }
+                }
+            }
+        }
```
peppy, line 758:
```diff
diff --git a/osu.Server.Spectator/Hubs/MultiplayerHub.cs b/osu.Server.Spectator/Hubs/MultiplayerHub.cs
index c3fc519..1404816 100644
--- a/osu.Server.Spectator/Hubs/MultiplayerHub.cs
+++ b/osu.Server.Spectator/Hubs/MultiplayerHub.cs
@@ -737,15 +737,15 @@ namespace osu.Server.Spectator.Hubs
         /// <param name="room">The room.</param>
         /// <param name="users">The users whose states are to be changed.</param>
         /// <param name="state">The new user state.</param>
-        /// <param name="existingClient">Any existing <see cref="MultiplayerClientState"/> usage.
+        /// <param name="currentUser">The <see cref="MultiplayerClientState"/> for the current user context.
         /// Must not be <c>null</c> if a user has been retrieved via <see cref="StatefulUserHub{TClient,TUserState}.UserStates"/> or
         /// <see cref="StatefulUserHub{TClient,TUserState}.GetOrCreateLocalUserState"/> prior to this call.</param>
-        private async Task changeAndBroadcastAllUserStates(ServerMultiplayerRoom room, MultiplayerRoomUser[] users, MultiplayerUserState state, MultiplayerClientState? existingClient)
+        private async Task changeAndBroadcastAllUserStates(ServerMultiplayerRoom room, MultiplayerRoomUser[] users, MultiplayerUserState state, MultiplayerClientState? currentUser)
         {
             foreach (var u in users)
             {
-                if (u.UserID == existingClient?.UserId)
-                    await changeAndBroadcastUserState(room, u, existingClient, state);
+                if (u.UserID == currentUser?.UserId)
+                    await changeAndBroadcastUserState(room, u, currentUser, state);
                 else
                 {
                     using (var userUsage = await UserStates.GetForUse(u.UserID))

```

I think calling this the "current" user or "local" user reads better than "existing". I'd update all the other methods which have this parameter to match.

peppy (changes requested):

peppy:
I've left a review of changes required to use this approach, but in line with what i mentioned above, how about a method on `EntityStore` which can retrieve the `connectionId` specifically without the full usage overhead? `GetItemUnsafe` already exists and could potentially be used for testing feasability.

smoogipoo:
The problem is that most entry methods incoming from clients need to get the `MultiplayerClientState` entity first, in order to retrieve the room entity, e.g:
https://github.com/ppy/osu-server-spectator/blob/b18144df5da7c447051cf479ddc4cc02d489d9cc/osu.Server.Spectator/Hubs/MultiplayerHub.cs#L290-L295
https://github.com/ppy/osu-server-spectator/blob/b18144df5da7c447051cf479ddc4cc02d489d9cc/osu.Server.Spectator/Hubs/MultiplayerHub.cs#L838-L846
This effectively results in locking the the entire duration of the call, which can go through a pretty deep callstack before, at some point, we may need the connection IDs of every user in the room inclusive of the one which was locked at the very start of the callstack.

peppy:
Hmm `GetItemUnsafe` is one level too low.

I'd propose a `UserTrackingStore : EntityTrackingStore` which has a `GetConectionIdForUser` method which bypasses all locking requirements (so checking usages is not required).

peppy:
Was thinking something like this:

```csharp
    public class UserStore<TUserState> : EntityStore<TUserState>
        where TUserState : ClientState
    {
        public string? GetConnectionIdForUser(long id)
        {
            lock (EntityMapping)
            {
                if (!EntityMapping.TryGetValue(id, out var usage))
                    return null;

                return usage.Item?.ConnectionId;
            }
        }
    }
```

but that requires a few changes to DI injection, so testing could be done with:

```diff
diff --git a/osu.Server.Spectator/Entities/EntityStore.cs b/osu.Server.Spectator/Entities/EntityStore.cs
index f278591..eca2767 100644
--- a/osu.Server.Spectator/Entities/EntityStore.cs
+++ b/osu.Server.Spectator/Entities/EntityStore.cs
@@ -7,18 +7,39 @@ using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
 using osu.Framework.Extensions.ObjectExtensions;
+using osu.Server.Spectator.Hubs;
 using StatsdClient;
 
 namespace osu.Server.Spectator.Entities
 {
+    public static class EntityStoreExtensions
+    {
+        public static string? GetConnectionIdForUser<TUserState>(this EntityStore<TUserState> store, long id)
+            where TUserState : ClientState
+        {
+            lock (store.EntityMapping)
+            {
+                if (!store.EntityMapping.TryGetValue(id, out var usage))
+                    return null;
+
+                return usage.Item?.ConnectionId;
+            }
+        }
+    }
+
+    public class UserStore<TUserState> : EntityStore<TUserState>
+        where TUserState : ClientState
+    {
+    }
+
     /// <summary>
     /// Tracks and ensures consistency of a collection of entities that have a related permanent ID.
     /// </summary>
     /// <typeparam name="T">The type of the entity being tracked.</typeparam>
-    public sealed class EntityStore<T>
+    public class EntityStore<T>
         where T : class
     {
-        private readonly Dictionary<long, TrackedEntity> entityMapping = new Dictionary<long, TrackedEntity>();
+        public readonly Dictionary<long, TrackedEntity> EntityMapping = new Dictionary<long, TrackedEntity>();
 
         private const int lock_timeout = 5000;
 
@@ -39,9 +60,9 @@ namespace osu.Server.Spectator.Entities
             {
                 TrackedEntity? item;
 
-                lock (entityMapping)
+                lock (EntityMapping)
                 {
-                    if (!entityMapping.TryGetValue(id, out item))
+                    if (!EntityMapping.TryGetValue(id, out item))
                     {
                         if (!createOnMissing)
                         {
@@ -49,8 +70,8 @@ namespace osu.Server.Spectator.Entities
                             throw new KeyNotFoundException($"Attempted to get untracked entity {typeof(T)} id {id}");
                         }
 
-                        entityMapping[id] = item = new TrackedEntity(id, this);
-                        DogStatsd.Gauge($"{statsDPrefix}.total-tracked", entityMapping.Count);
+                        EntityMapping[id] = item = new TrackedEntity(id, this);
+                        DogStatsd.Gauge($"{statsDPrefix}.total-tracked", EntityMapping.Count);
                         DogStatsd.Increment($"{statsDPrefix}.create");
                     }
                 }
@@ -82,9 +103,9 @@ namespace osu.Server.Spectator.Entities
         {
             TrackedEntity? item;
 
-            lock (entityMapping)
+            lock (EntityMapping)
             {
-                if (!entityMapping.TryGetValue(id, out item))
+                if (!EntityMapping.TryGetValue(id, out item))
                     // was not tracking.
                     return;
             }
@@ -108,9 +129,9 @@ namespace osu.Server.Spectator.Entities
         /// </summary>
         public KeyValuePair<long, T>[] GetAllEntities()
         {
-            lock (entityMapping)
+            lock (EntityMapping)
             {
-                return entityMapping
+                return EntityMapping
                        .Where(kvp => kvp.Value.GetItemUnsafe() != null)
                        .Select(entity => new KeyValuePair<long, T>(entity.Key, entity.Value.GetItemUnsafe().AsNonNull()))
                        .ToArray();
@@ -122,20 +143,20 @@ namespace osu.Server.Spectator.Entities
         /// </summary>
         public void Clear()
         {
-            lock (entityMapping)
+            lock (EntityMapping)
             {
-                entityMapping.Clear();
+                EntityMapping.Clear();
             }
         }
 
         private void remove(long id)
         {
-            lock (entityMapping)
+            lock (EntityMapping)
             {
-                entityMapping.Remove(id);
+                EntityMapping.Remove(id);
 
                 DogStatsd.Increment($"{statsDPrefix}.removed");
-                DogStatsd.Gauge($"{statsDPrefix}.total-tracked", entityMapping.Count);
+                DogStatsd.Gauge($"{statsDPrefix}.total-tracked", EntityMapping.Count);
             }
         }
 

```

Haven't tested yet.

smoogipoo:
I was working along similar lines, and it passes tests + works in-game. I'll do that as a separate PR.

smoogipoo:
Superseded by https://github.com/ppy/osu-server-spectator/pull/111

PR #112, opened 2022-03-29
Cancel countdown when changing auto-start setting

smoogipoo:
As mentioned in https://github.com/ppy/osu/pull/17449#issuecomment-1079804704, except cancelling instead of restarting as it's easier to handle different cases.

### DIFF
--- a/osu.Server.Spectator/Hubs/MultiplayerHub.cs
+++ b/osu.Server.Spectator/Hubs/MultiplayerHub.cs
@@ -572,6 +572,12 @@ namespace osu.Server.Spectator.Hubs
                     Log(room, $"Switching queue mode to {settings.QueueMode}");
                 }
 
+                if (previousSettings.AutoStartDuration != settings.AutoStartDuration)
+                {
+                    room.StopCountdown();
+                    Log(room, $"Switching auto-start duration to {settings.AutoStartDuration}");
+                }
+
                 await OnMatchSettingsChanged(room);
 
                 await updateRoomStateIfRequired(room);
--- a/osu.Server.Spectator.Tests/Multiplayer/MultiplayerCountdownTest.cs
+++ b/osu.Server.Spectator.Tests/Multiplayer/MultiplayerCountdownTest.cs
@@ -336,6 +336,37 @@ namespace osu.Server.Spectator.Tests.Multiplayer
             }
         }
 
+        [Fact]
+        public async Task CountdownStoppedWhenAutoStartDurationChanged()
+        {
+            await Hub.JoinRoom(ROOM_ID);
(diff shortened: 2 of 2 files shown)


peppy (dismissed):

osu.Server.Spectator/Hubs/MultiplayerHub.cs
```diff
@@ -574,6 +574,8 @@ public async Task ChangeSettings(MultiplayerRoomSettings settings)
 
                 if (previousSettings.AutoStartDuration != settings.AutoStartDuration)
                 {
+                    // If autostart is still enabled, the countdown will be restarted in the `updateRoomStateIfRequired` call below.
+                    // Stopping the countdown here is important to ensure the new duration is applied over any previous autostart (or manual) countdown.
```
smoogipoo, line 578: It won't be restarted because all users are unreadied.

peppy: Ah, that's weird. Are we reading different PR descriptions?

Let me re-review this.

peppy: If that's the case, I think it may be more correct to put this call alongside the unready-all-users call? I'd probably expect any active countdown to restart if the host suddenly changes the match settings.

peppy (changes requested):
As proposed

osu.Server.Spectator/Hubs/MultiplayerHub.cs
smoogipoo: _Any_ match settings? Changing the room name should also cause the countdown to be cancelled? The other case that I might agree with is when the current playlist item changes, which also calls `unreadyAllUsers`, but I don't know about blanketing it on all match settings changes.

peppy: Fair, how about a `bool` argument to make it up to the caller?

smoogipoo: I guess you could apply the same thought behind calling `unreadyAllUsers` in the first place - should all users be unreadied even when it's only the room name that changes? Probably not.

Will move to `unreadyAllUsers`.

peppy: That's fine for now, but I think `NotifyMatchSettingsChanged(MultiplayerRoom room, bool resetReadyStates = true)` might work quite well.

smoogipoo: Sure, but I'll leave that to a separate change because there's other flags involved and more tests to be made.

peppy (approved):

PR #115, opened 2022-03-30
Fix mysql connection open not using `async` call

peppy:


osu.Server.Spectator/Database/DatabaseAccess.cs
```diff
@@ -234,12 +246,32 @@ public async Task EndMatchAsync(MultiplayerRoom room)
...
+
+            string host = Environment.GetEnvironmentVariable("DB_HOST") ?? "localhost";
+            string user = Environment.GetEnvironmentVariable("DB_USER") ?? "root";
+            string port = Environment.GetEnvironmentVariable("DB_PORT") ?? "3306";
+
+            DapperExtensions.InstallDateTimeOffsetMapper();
+
+            openConnection = new MySqlConnection($"Server={host};Port={port};Database=osu;User ID={user};ConnectionTimeout=5;ConnectionReset=false;Pooling=true;");
+
+            await openConnection.OpenAsync();
```
bdach, line 272: While this class is technically currently not used in a manner where it would need to be thread safe (fresh instances are created every time), this initialisation process is giving me the heebie-jeebies from a threading perspective. I don't think this is thread safe:

| Thread A | Thread B |
| :- | :- |
| `if (openConnection != null) // fails` | |
| | `if (openConnection != null) // fails` |
| `openConnection = new MySqlConnection(...)` | |
| `await openConnection.OpenAsync();` | |
| `return openConnection;` | |
| | `openConnection = new MySqlConnection(...)` |
| | `await openConnection.OpenAsync();` |
| | `return openConnection;` |

et voilà, there's a leaked DB connection. _Maybe_ the finaliser will close it, but this feels unnecessarily risky - although maybe it's just me being paranoid here?

Is it viable to keep `openConnection` readonly, create it in the ctor, and then only open it when needed? I'd feel much safer with that.

smoogipoo: Though I don't expect this class to ever be used in a situation like the above, I think the suggestion makes sense.

peppy: I didn't see this as important, but can convert to a `Lazy` to resolve the issue.

Note that the proposal above is hard to implement because it would require locking over the `OpenAsync` call.

smoogipoo (approved):
Let's go with this as-is for the time being, I'm not super interested in increasing complexity.

Issue #114, opened 2022-03-29
Disposed exception during playlist edit operation


peppy:
Opening for tracking purposes. Haven't figure out the cause yet..

```csharp
2022-03-27 14:16:58 [verbose]: [user:14613516] [room:162666] Editing playlist item 752460 for beatmap 1581930
2022-03-27 14:16:58 [verbose]: [user:27352453] [room:162666] User state changed from Ready to Idle
2022-03-27 14:16:58 [error]: [user:14613516] Failed to invoke hub method: EditPlaylistItem(osu.Game.Online.Rooms.MultiplayerPlaylistItem)
2022-03-27 14:16:58 [error]: System.ObjectDisposedException: Cannot access a disposed object.
2022-03-27 14:16:58 [error]: Object name: 'MultiplayerHub'.
2022-03-27 14:16:58 [error]: at Microsoft.AspNetCore.SignalR.Hub.CheckDisposed()
2022-03-27 14:16:58 [error]: at osu.Server.Spectator.Hubs.MultiplayerHub.changeAndBroadcastUserState(ServerMultiplayerRoom room, MultiplayerRoomUser user, MultiplayerUserState state) in /Users/dean/Projects/osu-server-spectator/osu.Server.Spectator/Hubs/MultiplayerHub.cs:line 739
2022-03-27 14:16:58 [error]: at osu.Server.Spectator.Hubs.MultiplayerHub.unreadyAllUsers(ServerMultiplayerRoom room) in /Users/dean/Projects/osu-server-spectator/osu.Server.Spectator/Hubs/MultiplayerHub.cs:line 982
2022-03-27 14:16:58 [error]: at osu.Server.Spectator.Hubs.MultiplayerHub.OnPlaylistItemChanged(ServerMultiplayerRoom room, MultiplayerPlaylistItem item) in /Users/dean/Projects/osu-server-spectator/osu.Server.Spectator/Hubs/MultiplayerHub.cs:line 939
2022-03-27 14:16:58 [error]: at osu.Server.Spectator.Hubs.MultiplayerQueue.EditItem(MultiplayerPlaylistItem item, MultiplayerRoomUser user) in /Users/dean/Projects/osu-server-spectator/osu.Server.Spectator/Hubs/MultiplayerQueue.cs:line 175
[...]

PR #116, opened 2022-03-30
Fix incorrect MultiplayerHub storage and usage

smoogipoo:
Fixes https://github.com/ppy/osu-server-spectator/issues/114

Hubs are transient objects (see bottom of https://docs.microsoft.com/en-us/aspnet/core/signalr/hubs?view=aspnetcore-6.0#configure-signalr-hubs), so it is incorrect to store them. The proper way to "store" them is to instead store an `IHubContext<THub>`, as explained in https://docs.microsoft.com/en-us/aspnet/core/signalr/hubcontext?view=aspnetcore-6.0.

Doing this incorrectly leads to multiple issues:
1. External calls require hacks, such as the hack I added for `Clients` in https://github.com/ppy/osu-server-spectator/pull/109:
https://github.com/ppy/osu-server-spectator/blob/be334d32a5fa44e7a8321a459f8ed2ce249fc4fc/osu.Server.Spectator/Hubs/MultiplayerHub.cs#L996-L1011
2. Direct-appearing calls on the `MultiplayerHub` are actually external calls, causing https://github.com/ppy/osu-server-spectator/issues/114.  
E.g.: `MultiplayerHub.EditPlaylistItem()` (new hub) -> `MultiplayerQueue.EditItem()` (references old hub) -> `MultiplayerHub.OnPlaylistItemChanged()` (old hub).

The `MultiplayerHubContext` class has been added to fix this:
- Instantiated by the `MultiplayerHub` and persisted to the `ServerMultiplayerRoom`.
- Communicates with clients via the `IHubContext<MultiplayerHub>` API.
- Contains the basic set of methods that are common to both `MultiplayerHub` and external usages (team types, queue, countdowns, etc).

Unfortunately, using `IHubContext` requires using the raw `SendAsync(string, params object[] parameters)` API to send messages to clients, similar to the osu!-side implementation.

I've also bunched in method renaming as discussed in Discord (e.g. `OnPlaylistItemChanged` -> `NotifyPlaylistItemChanged`) and xmldoc rewriting in this.

On top of fixing the above issue thread, it also removes the `Dispose()` hack for countdown timers. It could also probably remove https://github.com/ppy/osu-server-spectator/blob/be334d32a5fa44e7a8321a459f8ed2ce249fc4fc/osu.Server.Spectator/Hubs/StatefulUserHub.cs#L151-L158, but I haven't tested yet.


peppy:
Tests aren't passing for me..

smoogipoo:
Oops, for some reason those tests aren't under the `Multiplayer` namespace that I was running... Fixed.

peppy (approved):
Feels a lot better.

PR #118, opened 2022-04-05
Send UserLeft callback to the leaving user

smoogipoo:
If a new client logs into a user from an existing room, the existing user leaves the room as their connection is "cleaned up". They should receive a `UserLeft` callback to notify them that this is the case so that they can leave the room.

Note: The above scenario will throw an assert on older clients without the changes in https://github.com/ppy/osu/pull/17663.

Manually-tested scenarios:
- User leaves room.
- User is kicked.
- User is logged into by another client.

### DIFF
--- a/osu.Server.Spectator.Tests/Multiplayer/UserStateManagementTests.cs
+++ b/osu.Server.Spectator.Tests/Multiplayer/UserStateManagementTests.cs
@@ -347,7 +347,7 @@ namespace osu.Server.Spectator.Tests.Multiplayer
             await Hub.LeaveRoom();
 
             UserReceiver.Verify(r => r.UserLeft(It.IsAny<MultiplayerRoomUser>()), Times.Once);
-            User2Receiver.Verify(r => r.UserKicked(It.IsAny<MultiplayerRoomUser>()), Times.Never);
+            UserReceiver.Verify(r => r.UserKicked(It.IsAny<MultiplayerRoomUser>()), Times.Never);
         }
     }
 }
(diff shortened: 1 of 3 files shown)


osu.Server.Spectator.Tests/Multiplayer/UserStateManagementTests.cs
```diff
@@ -317,5 +317,37 @@ public async Task CanNotAbortGameplayInNonGameplayStates()
...
+
+            // Join another user to make sure the room doesn't close when the user leaves.
+            SetUserContext(ContextUser2);
+            await Hub.JoinRoom(ROOM_ID);
+
+            SetUserContext(ContextUser);
+            await Hub.LeaveRoom();
+
+            UserReceiver.Verify(r => r.UserLeft(It.IsAny<MultiplayerRoomUser>()), Times.Once);
+            User2Receiver.Verify(r => r.UserKicked(It.IsAny<MultiplayerRoomUser>()), Times.Never);
```
bdach, line 350: Not sure if this is a case of a copy-paste gone wrong - judging from the other test I'd expect to see

```suggestion
            UserReceiver.Verify(r => r.UserKicked(It.IsAny<MultiplayerRoomUser>()), Times.Never);
```

bdach:
Looks sensible aside from the above.

osu.Server.Spectator.Tests/Multiplayer/UserStateManagementTests.cs
smoogipoo: Indeed, I copy-pasted this from the line above :sweat_smile:

peppy:
Hmm, the thinking was that a `UserLeft` is always triggered by the user, so sending this to the user is meaningless (and potential added complexity). What would you say to changing the server to send a `UserKicked` in the scenario this is being used for instead of `UserLeft`?

smoogipoo:
Sure, using `UserKicked` would work.

peppy:
It should also be explicitly commented that `UserLeft` is only sent for *other* users and not the local user.

PR #121, opened 2022-04-20
Fix deleted beatmaps being allowed to be added to a multiplayer room

peppy:
Deleted beatmaps shouldn't be allowed to be used in multiplayer. As discoverd in https://github.com/ppy/osu/discussions/17106#discussioncomment-2599163.


smoogipoo:
Need to check whether osu-web also checks this (looks like it may not?): https://github.com/ppy/osu-web/blob/44103cdc243af886c2eda99856ef34b2696966a6/app/Models/Multiplayer/PlaylistItem.php#L41-L54 @nanaya

smoogipoo (approved):

nanaya:
There's a default scope on osu-web on the model to filter out deleted maps.

PR #122, opened 2022-04-22
Force start/abort multiplayer games after a timeout

smoogipoo:
Prereqs:
- [x] https://github.com/ppy/osu/pull/17933

The user state transition process for entering gameplay now looks like this:
```
1. Idle
2. Ready
3. WaitingForLoad (set by the server)
4. Loaded (set by the client, beatmap loaded but user is not yet ready to start gameplay)
5. ReadyForGameplay (set by the client, user is ready to start gameplay)
6. Playing (set by the server)
```

I've had to adjust existing tests to conform to the above new process for entering gameplay.

When the a match is started (clients receive `LoadRequested()`):
1. A 1-minute countdown is started.
2. Once the countdown fires, it starts gameplay for all users in the `Loaded` or `ReadyForGameplay` states, and aborts the loading process for all clients in the `WaitingForLoad` state (setting them to `Idle` and calling `LoadAborted()` in the process).  
- This may abort all users in the match if no users have finished loading.
- Countdowns are used since they provide an existing system and it would also be nice to eventually display a progress bar in the client.

Normal gameplay start:

https://user-images.githubusercontent.com/1329837/164751717-3a87e54b-1e21-4de5-8b74-78b054b48949.mp4

When one user is in a blocking overlay:

https://user-images.githubusercontent.com/1329837/164751764-c91b468d-113a-490c-8390-b4aa9cad7c41.mp4

When one user takes too long to load:

https://user-images.githubusercontent.com/1329837/164751811-263fbb31-f34c-403d-a2a7-33c6fb247ede.mp4

Notes on groups:
- The gameplay group is no longer used to transmit the `GameplayStarted` (previously `MatchStarted`) event. This is an intentional change because players in the `Ready` state are added to the gameplay group and it feels wrong to potentially send gameplay-related messages to those clients. The messages are invoked on individual clients instead.
- I plan to clean up group handling in a separate PR since I think there is worth to doing so, such as the automatic SignalR handling of connection lifetime.  
- As a result, I've had to update existing tests which were previously asserting on the gameplay/non-gameplay receivers to now assert on the individual clients.

### DIFF
--- a/SampleMultiplayerClient/MultiplayerClient.cs
+++ b/SampleMultiplayerClient/MultiplayerClient.cs
@@ -34,7 +34,8 @@ namespace SampleMultiplayerClient
             connection.On<int, MultiplayerUserState>(nameof(IMultiplayerClient.UserStateChanged), ((IMultiplayerClient)this).UserStateChanged);
             connection.On<int, BeatmapAvailability>(nameof(IMultiplayerClient.UserBeatmapAvailabilityChanged), ((IMultiplayerClient)this).UserBeatmapAvailabilityChanged);
             connection.On(nameof(IMultiplayerClient.LoadRequested), ((IMultiplayerClient)this).LoadRequested);
-            connection.On(nameof(IMultiplayerClient.MatchStarted), ((IMultiplayerClient)this).MatchStarted);
+            connection.On(nameof(IMultiplayerClient.LoadAborted), ((IMultiplayerClient)this).LoadAborted);
+            connection.On(nameof(IMultiplayerClient.GameplayStarted), ((IMultiplayerClient)this).GameplayStarted);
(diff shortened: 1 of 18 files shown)


peppy:
It's a pity that this causes the gameplay sync to be lost (previously it was mostly guaranteed). Maybe something to consider as a future effort though?

smoogipoo:
I'm not sure if that would be a breaking change if addressed further down the road. Perhaps something we may want to address in this PR.

peppy:
I guess the simplest solution would be to sent one extra server-to-client request to all playing users for beginning the transition to `Player`? Would probably involve a second countdown of 1-2 seconds to finally start the gameplay?

I think I'd prefer to do it as a follow-up, since it's only going to add to complexity here.

peppy (approved):

PR #123, opened 2022-04-26
Remove spinning loops in countdown tests

peppy:
While reviewing the new force gameplay PR I noticed that this is being added in more places, and doesn't feel great.

Addresses two points:
- The previous tests were waiting for a countdown to start as if it could be a delayed operation (but it can't be).
- The previous tests were using a loop to wait for a countdown to complete rather than checking the task completion direction.

### DIFF
--- a/osu.Server.Spectator.Tests/Multiplayer/MultiplayerCountdownTest.cs
+++ b/osu.Server.Spectator.Tests/Multiplayer/MultiplayerCountdownTest.cs
@@ -9,7 +9,6 @@ using System.Threading.Tasks;
 using Moq;
 using osu.Game.Online.Multiplayer;
 using osu.Game.Online.Multiplayer.Countdown;
-using osu.Server.Spectator.Hubs;
 using Xunit;
 
 namespace osu.Server.Spectator.Tests.Multiplayer
@@ -39,23 +38,21 @@ namespace osu.Server.Spectator.Tests.Multiplayer
             await Hub.ChangeState(MultiplayerUserState.Ready);
 
             await Hub.SendMatchRequest(new StartMatchCountdownRequest { Duration = TimeSpan.FromSeconds(3) });
-            await waitForCountingDown();
 
-            int attempts = 1000;
+            using (var usage = await Hub.GetRoom(ROOM_ID))
+                Assert.NotNull(usage.Item?.Countdown);
 
-            while (attempts-- > 0)
+            Task task;
+
+            using (var usage = await Hub.GetRoom(ROOM_ID))
             {
-                using (var usage = await Hub.GetRoom(ROOM_ID))
-                {
-                    var room = usage.Item;
-                    Debug.Assert(room != null);
+                var room = usage.Item;
+                Debug.Assert(room != null);
 
-                    if (!room.IsCountdownRunning)
-                        break;
-
-                    Thread.Sleep(10);
-                }
+                task = room.WaitForCountdown();
             }
+
+            await task;
 
             using (var usage = await Hub.GetRoom(ROOM_ID))
             {
@@ -74,7 +71,9 @@ namespace osu.Server.Spectator.Tests.Multiplayer
             await Hub.ChangeState(MultiplayerUserState.Ready);
 
             await Hub.SendMatchRequest(new StartMatchCountdownRequest { Duration = TimeSpan.FromMinutes(1) });
-            await waitForCountingDown();
+
+            using (var usage = await Hub.GetRoom(ROOM_ID))
+                Assert.NotNull(usage.Item?.Countdown);
 
             using (var usage = await Hub.GetRoom(ROOM_ID))
             {
@@ -86,7 +85,7 @@ namespace osu.Server.Spectator.Tests.Multiplayer
                 GameplayReceiver.Verify(r => r.LoadRequested(), Times.Never);
             }
 
-            await finishCountdown();
+            await skipToEndOfCountdown();
 
             using (var usage = await Hub.GetRoom(ROOM_ID))
             {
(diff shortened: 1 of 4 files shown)


osu.Server.Spectator.Tests/Multiplayer/MultiplayerCountdownTest.cs
```diff
@@ -248,24 +251,26 @@ public async Task AutoStartCountdownDoesNotStartWithZeroDuration()
...
         {
             await Hub.JoinRoom(ROOM_ID);
             await Hub.ChangeSettings(new MultiplayerRoomSettings { AutoStartDuration = TimeSpan.FromMinutes(1) });
 
             await Hub.ChangeState(MultiplayerUserState.Ready);
-            await waitForCountingDown();
 
-            await finishCountdown();
+            using (var usage = await Hub.GetRoom(ROOM_ID))
+                Assert.NotNull(usage.Item?.Countdown);
```
smoogipoo, line 267: The countdown object is set on the async thread. Removing the `waitForCountingDown()` makes this assert potentially intermittent and testable by adding a `Task.Delay()` at the start of the async operation.

smoogipoo (changes requested):

osu.Server.Spectator.Tests/Multiplayer/MultiplayerCountdownTest.cs
peppy: Thanks, I knew there was some reason it was done like this. I've done a bit of refactoring, see what you think (seems to work now with `Task.Delay`s scattered in the flow).

osu.Server.Spectator/Hubs/ServerMultiplayerRoom.cs
```diff
@@ -103,48 +103,24 @@ private MatchTypeImplementation createTypeImplementation(MatchType type)
...
-
-                    roomUsage.Item.Countdown = countdown;
-
-                    countdownStartTime = DateTimeOffset.Now;
-                    countdownDuration = countdown.TimeRemaining;
+            countdownTask = start();
 
-                    await hub.NotifyNewMatchEvent(roomUsage.Item, new CountdownChangedEvent { Countdown = countdown });
-                }
+            await hub.NotifyNewMatchEvent(this, new CountdownChangedEvent { Countdown = countdown });
```
smoogipoo, line 120: I'd prefer this to be sent before the task is started to avoid potential misordering of events.

osu.Server.Spectator/Hubs/ServerMultiplayerRoom.cs
```diff
@@ -165,13 +141,15 @@ async Task start()
                         if (roomUsage.Item == null)
                             return;
 
-                        roomUsage.Item.Countdown = null;
-
-                        await hub.NotifyNewMatchEvent(roomUsage.Item, new CountdownChangedEvent { Countdown = null });
+                        // A new countdown may have since replaced the currently tracked countdown.
+                        if (roomUsage.Item.Countdown != countdown)
+                            return;
```
smoogipoo, line 146: As far as I can tell, this is a newly-added check, however I logically can't follow the case where this would be required. Because any time where `StopCountdown()` is called, we must already be inside a usage of a room, which means `stopSource` should always be cancelled before this code is run and the check following this one should enough to cover this case.

Am I missing something?

peppy: I think it may no longer be required. Previously I didn't have the `StopCountdown` flow synchronous (as it is when I pushed the latest changes) and caused this requirement. Will re-test.

osu.Server.Spectator.Tests/Multiplayer/MultiplayerCountdownTest.cs
smoogipoo: I think it feels better with the rewrite, it follows how I originally envisioned this method a bit more.

osu.Server.Spectator/Hubs/ServerMultiplayerRoom.cs
peppy: Seems to work fine without it. I've added an assert in place as I think it makes it more obvious as to expectations.

osu.Server.Spectator.Tests/Multiplayer/MultiplayerCountdownTest.cs
```diff
@@ -9,14 +9,13 @@
...
 using osu.Game.Online.Multiplayer.Countdown;
-using osu.Server.Spectator.Hubs;
 using Xunit;
 
 namespace osu.Server.Spectator.Tests.Multiplayer
 {
     public class MultiplayerCountdownTest : MultiplayerTest
     {
-        [Fact]
+        [Fact(Timeout = 1000)]
```
smoogipoo, line 18: Not sure about these timeouts... I guess we can keep them until they cause intermittent CI failures, seem unnecessary otherwise.

smoogipoo (approved):

osu.Server.Spectator.Tests/Multiplayer/MultiplayerCountdownTest.cs
smoogipoo: Speak of the devil (CI failure) :smile:

peppy: The idea was to avoid runaway tests that never end (if we happen to break something). Should I set them to some higher number like 60s?

smoogipoo: Sure, I think that's reasonable.

smoogipoo (approved):

PR #125, opened 2022-05-09
👷 Update CI to tag both latest and using git tag

ThePooN:
Similarly to osu-web's current flow, tag Docker image using both latest and git tag the build corresponds to.

Also, this repository hasn't since a tag and thus a Docker build in 16 months, it would be great to push a tag after merging this.


peppy (approved):

ThePooN:
tyvm. would be great if you could tag the current build in production

peppy:
done

PR #127, opened 2022-05-13
Add CreateUser() test setup method + refactor setup

smoogipoo:
For writing tests with more than 2 users, I think it could be useful to have a `CreateUser()` setup method. This also reorders the setup method to read a bit more linearly.


smoogipoo:
Hang on a sec.

smoogipoo:
Good to go.

peppy (approved):

Issue #126, opened 2022-05-11
Investigate round-robin pinning items from new users to the top of the queue


smoogipoo:
Apparently this is still happening even though the intention behind https://github.com/ppy/osu-server-spectator/pull/89 was for new users to come after the first items from existing users. So if the queue looks like:
```
User1
User2
User1
```
When User3 adds their item, the queue would turn into
```
User1
User2
User3
User1
```

PR #128, opened 2022-05-13
Improve round robin queueing algorithm

smoogipoo:
Fixes https://github.com/ppy/osu-server-spectator/issues/126

Firstly, a very simple fix - the playlist wasn't reordered when items were removed.

As for round robin... The previous algorithm didn't really work at all, and after much trying/deliberation, this is what I've come up with. It's a rework of the algorithm in entirety.

Consider the queue as sets of items such that each user appears at most once in each set. The order in which items from users are picked into sets is by the order in which they were added to the playlist.

In the first set containing the single earliest item added by each user, items are sorted by their existing `PlaylistOrder` value. When an item is added by a _new_ user, it is added with `PlaylistOrder = ushort.Max`, so it will always be added to the end of this list.  
Preserving the existing `PlaylistOrder` is important for when items are played, in order to prevent the next item by the same user whose item was played from jumping to the front of the queue.

If there is a `PlaylistOrder` tie in the first set, which is currently not possible but may become possible in the future if we allow rooms to be created with more than 1 item, then the tie is broken by taking the order in which the items were added.

Any further sets of items are ordered such that they maintain the same ordering as the first set.

The sets are then combined to form the queue.

---

One edge case of this algorithm is during removal. Suppose the queue is
```
| item | user |
| ---- | ---- |
|  1   |  1   |
|  4   |  2   |
|  5   |  3   |
|  9   |  4   |
|  2   |  1   |
|  7   |  2   |
|  6   |  3   |
|  10  |  4   |
```
If the item 4 is removed, the movement of item 7 is as follows:
```
| item | user |
| ---- | ---- |
|  1   |  1   |
|  5   |  3   |
|  9   |  4   |   
|      |      |<---|
|  2   |  1   |    |
|  7   |  2   | ---|
|  6   |  3   |
|  10  |  4   |
```
Which is unexpected since item 7 was added earlier than item 9.

This is due to the preservation of order via the use of the existing `PlaylistOrder` value. There may be a simple solution to this, but I haven't figured it out yet, and from the view of not complicating this PR too much I'd say we should go with it.

### DIFF
--- a/osu.Server.Spectator.Tests/Multiplayer/MultiplayerAllPlayersRoundRobinQueueTests.cs
+++ b/osu.Server.Spectator.Tests/Multiplayer/MultiplayerAllPlayersRoundRobinQueueTests.cs
@@ -14,7 +14,7 @@ namespace osu.Server.Spectator.Tests.Multiplayer
     public class MultiplayerAllPlayersRoundRobinQueueTests : MultiplayerTest
     {
         [Fact]
-        public async Task RoundRobinOrdering()
+        public async Task RoundRobinOrderingWithGameplay()
         {
             Database.Setup(d => d.GetBeatmapChecksumAsync(3333)).ReturnsAsync("3333");
 
@@ -28,55 +28,30 @@ namespace osu.Server.Spectator.Tests.Multiplayer
             await checkCurrentItem(1);
 
             // User 1 adds an extra item
-            await Hub.AddPlaylistItem(new MultiplayerPlaylistItem
-            {
-                BeatmapID = 3333,
-                BeatmapChecksum = "3333"
-            });
+            await addItem();
 
             await checkCurrentItem(1);
-            await checkOrder(1, 0);
-            await checkOrder(2, 1);
+            await checkOrder(1, 2);
 
             // Item played.
             await runGameplay();
             await checkCurrentItem(2);
-            await checkOrder(2, 0);
+            await checkOrder(2);
 
             // User 2 adds two items.
             SetUserContext(ContextUser2);
-
-            await Hub.AddPlaylistItem(new MultiplayerPlaylistItem
-            {
-                BeatmapID = 3333,
-                BeatmapChecksum = "3333"
-            });
-
-            await Hub.AddPlaylistItem(new MultiplayerPlaylistItem
-            {
-                BeatmapID = 3333,
-                BeatmapChecksum = "3333"
-            });
+            await addItem();
+            await addItem();
 
             await checkCurrentItem(2);
-            await checkOrder(2, 0);
-            await checkOrder(3, 1);
-            await checkOrder(4, 2);
+            await checkOrder(2, 3, 4);
 
             // User 1 adds an item.
             SetUserContext(ContextUser);
-
-            await Hub.AddPlaylistItem(new MultiplayerPlaylistItem
-            {
-                BeatmapID = 3333,
-                BeatmapChecksum = "3333"
-            });
+            await addItem();
 
             await checkCurrentItem(2);
-            await checkOrder(2, 0);
-            await checkOrder(3, 1);
-            await checkOrder(5, 2);
-            await checkOrder(4, 3);
+            await checkOrder(2, 3, 5, 4);
 
             // Gameplay is now run to ensure the ordering doesn't change.
             await runGameplay();
@@ -111,6 +86,15 @@ namespace osu.Server.Spectator.Tests.Multiplayer
             SetUserContext(ContextUser);
         }
 
+        private async Task addItem()
+        {
+            await Hub.AddPlaylistItem(new MultiplayerPlaylistItem
+            {
+                BeatmapID = 3333,
+                BeatmapChecksum = "3333"
+            });
+        }
+
         private async Task checkCurrentItem(long expectedItemId)
         {
             using (var usage = await Hub.GetRoom(ROOM_ID))
@@ -122,14 +106,14 @@ namespace osu.Server.Spectator.Tests.Multiplayer
             }
         }
 
-        private async Task checkOrder(long itemId, ushort order)
+        private async Task checkOrder(params long[] itemIdsInOrder)
         {
             using (var usage = await Hub.GetRoom(ROOM_ID))
             {
                 var room = usage.Item;
                 Debug.Assert(room != null);
 
-                Assert.Equal(order, room.Playlist.Single(i => i.ID == itemId).PlaylistOrder);
+                Assert.Equal(itemIdsInOrder, room.Playlist.Where(item => !item.Expired).OrderBy(item => item.PlaylistOrder).Select(item => item.ID));
             }
         }
     }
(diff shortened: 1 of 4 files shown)


osu.Server.Spectator.Tests/Multiplayer/MultiplayerAllPlayersRoundRobinQueueTests.cs
```diff
@@ -28,55 +28,30 @@ public async Task RoundRobinOrdering()
...
-                BeatmapChecksum = "3333"
-            });
+            await addItem();
 
             await checkCurrentItem(2);
-            await checkOrder(2, 0);
-            await checkOrder(3, 1);
-            await checkOrder(5, 2);
-            await checkOrder(4, 3);
+            await checkOrder(2, 3, 5, 4);
```
smoogipoo, line 54: There's no actual changes to this test method, just extracting the first column here into the params array. The second column is redundant and will always be 0, 1, 2, 3, etc.., so it's been removed.

osu.Server.Spectator.Tests/Multiplayer/MultiplayerAllPlayersRoundRobinQueueTests.cs
```diff
@@ -91,6 +66,200 @@ public async Task RoundRobinOrdering()
...
+
+            // | item | user |
+            // | ---- | ---- |
+            // |  1   |  1   |
+            // |  4   |  2   |
+            // |  5   |  3   | (new)
+            // |  2   |  1   |
+            // |  3   |  1   |
+            await addItem();
+            await checkOrder(1, 4, 5, 2, 3);
```
smoogipoo, line 138: Of importance, this is the first test that fails on master, showing how https://github.com/ppy/osu-server-spectator/issues/126 was fixed in this PR.

The order on master is `1 5 4 2 3`, so item `5` by the new user was inserted towards the top of the queue, in front of an existing one of the same "priority" by another user.

osu.Server.Spectator/Hubs/MultiplayerQueue.cs
```diff
@@ -301,5 +313,39 @@ private async Task updatePlaylistOrder(IDatabaseAccess db)
                 await hub.NotifyPlaylistItemChanged(room, item);
             }
         }
+
+        private IEnumerable<IEnumerable<T>> interleaveMany<T>(IEnumerable<IEnumerable<T>> collections)
```
peppy, line 317: I'm having a very hard time understanding what this function is doing. Could you add some documentation explaining the overall purpose?

osu.Server.Spectator/Hubs/MultiplayerQueue.cs
```diff
@@ -263,28 +269,34 @@ private async Task updatePlaylistOrder(IDatabaseAccess db)
...
+
+                    bool isFirstSet = true;
+                    var firstSetOrderByUserId = new Dictionary<int, int>();
 
-                    // Assign a priority for items from each user, starting from 0 and increasing in order which the user added the items.
-                    foreach (var group in room.Playlist.Where(item => !item.Expired).OrderBy(item => item.ID).GroupBy(item => item.OwnerID))
+                    // Group each user's items in order of addition.
+                    var userItemGroups = room.Playlist.Where(item => !item.Expired).OrderBy(item => item.ID).GroupBy(item => item.OwnerID);
+
+                    foreach (IEnumerable<MultiplayerPlaylistItem> set in interleaveMany(userItemGroups))
```
peppy, line 280: Would rename "set" to "item". "set" to me reads like it's a "set of items" which makes it hard to mentally parse this function.

peppy:
The ordering in the tests looks same, but... I'm having a hard time following the implementation code-wise.

osu.Server.Spectator/Hubs/MultiplayerQueue.cs
bdach: From a read through of the code, I believe what this does is interleave a sequence of sequences in the order they're given. To give an actual example, let's say we have three enumerables:

```
A = [A1, A2, A3, A4]
B = [B1, B2]
C = [C1, C2, C3]
```

Then `interleaveMany(new[] { A, B, C })` would produce a single flattened enumerable containing

```
[A1, B1, C1, A2, B2, C2, A3, C3, A4]
```

I do agree that the implementation is pretty optimised, but... unorthodox, which isn't doing any favours wrt readability.

frenzibyte: I feel moving it to an extensions/utils class and providing a simplified explanation along with the example stated above would be enough to give this more sense IMO. The `RemoveAll` part in the method is a bit weird to read through as well, but if it's for optimisation, then I guess it's fine to be used as-is?

osu.Server.Spectator/Hubs/MultiplayerQueue.cs
bdach: The `set` here *is* an enumerable here, though, not a single item. In the first iteration `set` is a set composed of each individual player's oldest queued playlist items, in the second - the second oldest queued items, in the third - third oldest, etc.

osu.Server.Spectator/Hubs/MultiplayerQueue.cs
peppy: I can understand the intention from the name of the function, but can agree it should be an isolated method, probably with a test showing it does what we expect it to.

osu.Server.Spectator/Hubs/MultiplayerQueue.cs
smoogipoo: This is correct. It's also why I explicitly specified the type.

osu.Server.Spectator/Hubs/MultiplayerQueue.cs
smoogipoo: That's understandable, I can move it to an extension method with tests. The reason for the optimisation is because I'm wary of unlimited room sizes, but on another think over it it's not that big of a deal to just do `O(m*n)` iterations over the entire list, where `n` is the number of players and `m` is the maximum number of elements in any one player (capped at 3, currently). Thus I will be removing the optimisation.

osu.Server.Spectator/Hubs/MultiplayerQueue.cs
frenzibyte: Maybe separating it to a local may enhance the readability a bit? (not sure on the given name though)
```csharp
var groupsPerPriority = Interleave(userItemGroups);

foreach (var itemGroup in groupsPerPriority)
{

}
```

I was also confused into thinking that the foreach loop was iterating per item rater than per set mainly due to the method name including `Many` which makes me think it returns a flattened `IEnumerable<T>` type.

smoogipoo: I think that would confuse me more...

bdach (approved):

peppy (approved):

PR #131, opened 2022-05-24
Fix development startup exception due to sentry

smoogipoo:


osu.Server.Spectator/Startup.cs
```diff
@@ -94,7 +94,9 @@ public void Configure(IApplicationBuilder app, IWebHostEnvironment env)
 
             app.UseWebSockets();
 
+#if !DEBUG
```
peppy, line 97: What's the exception this is raising?

Rather than doing this, I'd prefer to see the DSN not configured similar to how it's done in osu! (and leaving this setting in place). Since you're now polluting `Startup` with DEBUG specific clauses, which is what `StartupDevelopment` is intended for.

peppy (changes requested):

osu.Server.Spectator/Startup.cs
smoogipoo: I didn't consider that. And technically it should probably be `if (env.IsDevelopment()`, which is already used in another place in this method.

Not sure what the solution is here...

```
crit: Microsoft.AspNetCore.Hosting.Diagnostics[6]
      Application startup exception
      System.InvalidOperationException: Unable to resolve service for type 'System.Func`1[Sentry.IHub]' while attempting to activate 'Sentry.AspNetCore.SentryTracingMiddleware'.
         at Microsoft.Extensions.Internal.ActivatorUtilities.ConstructorMatcher.CreateInstance(IServiceProvider provider)
         at Microsoft.Extensions.Internal.ActivatorUtilities.CreateInstance(IServiceProvider provider, Type instanceType, Object[] parameters)
         at Microsoft.AspNetCore.Builder.UseMiddlewareExtensions.<>c__DisplayClass5_0.<UseMiddleware>b__0(RequestDelegate next)
         at Microsoft.AspNetCore.Builder.ApplicationBuilder.Build()
         at Microsoft.AspNetCore.Hosting.GenericWebHostService.StartAsync(CancellationToken cancellationToken)
Unhandled exception. System.InvalidOperationException: Unable to resolve service for type 'System.Func`1[Sentry.IHub]' while attempting to activate 'Sentry.AspNetCore.SentryTracingMiddleware'.
   at Microsoft.Extensions.Internal.ActivatorUtilities.ConstructorMatcher.CreateInstance(IServiceProvider provider)
   at Microsoft.Extensions.Internal.ActivatorUtilities.CreateInstance(IServiceProvider provider, Type instanceType, Object[] parameters)
   at Microsoft.AspNetCore.Builder.UseMiddlewareExtensions.<>c__DisplayClass5_0.<UseMiddleware>b__0(RequestDelegate next)
   at Microsoft.AspNetCore.Builder.ApplicationBuilder.Build()
   at Microsoft.AspNetCore.Hosting.GenericWebHostService.StartAsync(CancellationToken cancellationToken)
   at Microsoft.Extensions.Hosting.Internal.Host.StartAsync(CancellationToken cancellationToken)
   at Microsoft.Extensions.Hosting.HostingAbstractionsHostExtensions.RunAsync(IHost host, CancellationToken token)
   at Microsoft.Extensions.Hosting.HostingAbstractionsHostExtensions.RunAsync(IHost host, CancellationToken token)
   at Microsoft.Extensions.Hosting.HostingAbstractionsHostExtensions.Run(IHost host)
   at osu.Server.Spectator.Program.Main(String[] args) in /home/smgi/Repos/osu-server-spectator/osu.Server.Spectator/Program.cs:line 27
```

peppy: In `Program`, move the sentry registration outside of the `#if` and specify an empty DSN if debug, should work fine.

smoogipoo: Works. I've force-pushed to not pollute commit history since the original code is completely gone.

PR #133, opened 2022-05-26
Add graceful shutdown support

peppy:
This will change shutdown behaviour to wait for all multiplayer rooms and spectator sessions to end before shutting down the server. Until now I've been manually managing this with two deploys. The new behaviour is required to deploy to kubernetes, where the application lifetime will be automatically managed.

This works well because websocket connections persist to existing server instances, but new connections will be directed to the fresher instance. Client side handling of the new `The server is shutting down.` error message allows clients which are attempting to start new spectator sessions or multiplayer rooms to first trigger a reconnect and find the newer server.

Eventually we may want to use multiple endpoints or sticky cookies/headers to persist server across connection loss (aka cloudflare support, since cloudflare arbitrarily drops websocket connections) but for now we are avoiding this by dropping cloudflare temporarily.

Not sure if tests can be added for this kind of coverage. You'll have to take my word that it works. And some videos.

The case where graceful shutdown works:

https://user-images.githubusercontent.com/191335/170455093-37688cbe-5106-45e3-8d60-c6ea03b22620.mp4

The case where it times out:

https://user-images.githubusercontent.com/191335/170455511-aa5c1e23-a0da-4ec9-a111-c991d7090c91.mp4

I've chosen a very high number for the non-graceful timeout as there's no harm in having old instances sitting around. We'll adjust as necessary (and potentially move to envvar if we want to tune it further).

### DIFF
--- a/osu.Server.Spectator/Entities/EntityStore.cs
+++ b/osu.Server.Spectator/Entities/EntityStore.cs
@@ -6,6 +6,8 @@ using System.Collections.Generic;
 using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
+using Microsoft.AspNetCore.SignalR;
+using Microsoft.Extensions.Hosting;
 using osu.Framework.Extensions.ObjectExtensions;
 using StatsdClient;
 
@@ -15,7 +17,7 @@ namespace osu.Server.Spectator.Entities
     /// Tracks and ensures consistency of a collection of entities that have a related permanent ID.
     /// </summary>
     /// <typeparam name="T">The type of the entity being tracked.</typeparam>
-    public sealed class EntityStore<T>
+    public sealed class EntityStore<T> : IEntityStore
         where T : class
     {
         private readonly Dictionary<long, TrackedEntity> entityMapping = new Dictionary<long, TrackedEntity>();
@@ -23,6 +25,25 @@ namespace osu.Server.Spectator.Entities
         private const int lock_timeout = 5000;
 
         private string statsDPrefix => $"entities.{typeof(T).Name}";
+
+        private bool acceptingNewEntities = true;
+
+        public EntityStore(GracefulShutdownManager shutdownManager, IHostApplicationLifetime applicationLifetime)
+            : this()
+        {
+            // Registration is required here to ensure `ApplicationStopping` triggers the shutdown wait before SignalR disconnects clients.
+            // Hope we can find a better way to make this work in the future.
+            applicationLifetime.ApplicationStopping.Register(shutdownManager.WaitForSafeShutdown);
+        }
+
+        public EntityStore()
+        {
+        }
+
+        /// <summary>
+        /// Inform this entity store that a server shutdown transition is in progress, and new entities should not be allowed.
+        /// </summary>
+        public void StopAcceptingEntities() => acceptingNewEntities = false;
 
         /// <summary>
         /// Retrieves an entity.
@@ -62,6 +83,9 @@ namespace osu.Server.Spectator.Entities
                             DogStatsd.Increment($"{statsDPrefix}.get-notfound");
                             throw new KeyNotFoundException($"Attempted to get untracked entity {typeof(T)} id {id}");
                         }
+
+                        if (!acceptingNewEntities)
+                            throw new HubException("Server is shutting down.");
 
                         entityMapping[id] = item = new TrackedEntity(id, this);
                         DogStatsd.Gauge($"{statsDPrefix}.total-tracked", entityMapping.Count);
@@ -237,6 +261,15 @@ namespace osu.Server.Spectator.Entities
             {
                 if (IsDestroyed) throw new InvalidOperationException("Attempted to use an item which has already been destroyed");
                 if (shouldBeLocked && !isLocked) throw new InvalidOperationException("Attempted to access a tracked entity without holding a lock");
+            }
+        }
+
+        public bool AnyRemainingUsage
+        {
+            get
+            {
+                lock (entityMapping)
+                    return entityMapping.Count > 0;
             }
         }
     }
--- a/osu.Server.Spectator/Entities/EntityStore.cs
+++ b/osu.Server.Spectator/Entities/EntityStore.cs
@@ -7,7 +7,6 @@ using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.SignalR;
-using Microsoft.Extensions.Hosting;
 using osu.Framework.Extensions.ObjectExtensions;
 using StatsdClient;
 
@@ -27,18 +26,6 @@ namespace osu.Server.Spectator.Entities
         private string statsDPrefix => $"entities.{typeof(T).Name}";
 
         private bool acceptingNewEntities = true;
-
-        public EntityStore(GracefulShutdownManager shutdownManager, IHostApplicationLifetime applicationLifetime)
-            : this()
-        {
-            // Registration is required here to ensure `ApplicationStopping` triggers the shutdown wait before SignalR disconnects clients.
-            // Hope we can find a better way to make this work in the future.
-            applicationLifetime.ApplicationStopping.Register(shutdownManager.WaitForSafeShutdown);
-        }
-
-        public EntityStore()
-        {
-        }
 
         public void StopAcceptingEntities() => acceptingNewEntities = false;
 
--- a/osu.Server.Spectator/Entities/EntityStore.cs
+++ b/osu.Server.Spectator/Entities/EntityStore.cs
@@ -40,10 +40,18 @@ namespace osu.Server.Spectator.Entities
         {
         }
 
-        /// <summary>
-        /// Inform this entity store that a server shutdown transition is in progress, and new entities should not be allowed.
-        /// </summary>
         public void StopAcceptingEntities() => acceptingNewEntities = false;
+
+        public int RemainingUsages
+        {
+            get
+            {
+                lock (entityMapping)
+                    return entityMapping.Count;
+            }
+        }
+
+        public string EntityName => typeof(T).Name;
 
         /// <summary>
         /// Retrieves an entity.
@@ -261,15 +269,6 @@ namespace osu.Server.Spectator.Entities
             {
                 if (IsDestroyed) throw new InvalidOperationException("Attempted to use an item which has already been destroyed");
                 if (shouldBeLocked && !isLocked) throw new InvalidOperationException("Attempted to access a tracked entity without holding a lock");
-            }
-        }
-
-        public bool AnyRemainingUsage
-        {
-            get
-            {
-                lock (entityMapping)
-                    return entityMapping.Count > 0;
             }
         }
     }
--- a/osu.Server.Spectator/GracefulShutdownManager.cs
+++ b/osu.Server.Spectator/GracefulShutdownManager.cs
@@ -26,22 +26,10 @@ public class GracefulShutdownManager
 
     public GracefulShutdownManager(EntityStore<ServerMultiplayerRoom> roomStore, EntityStore<SpectatorClientState> clientStateStore, IHostApplicationLifetime hostApplicationLifetime)
     {
-        addDependentStore(roomStore);
-        addDependentStore(clientStateStore);
+        dependentStores.Add(roomStore);
+        dependentStores.Add(clientStateStore);
 
         hostApplicationLifetime.ApplicationStopping.Register(shutdownSafely);
-    }
-
-    /// <summary>
-    /// Add an entity store which should be waited on for all usages to have finished.
-    /// </summary>
-    /// <param name="store"></param>
-    private void addDependentStore(IEntityStore? store)
-    {
-        if (store == null)
-            return;
-
-        dependentStores.Add(store);
     }
 
     private void shutdownSafely()
--- a/osu.Server.Spectator/GracefulShutdownManager.cs
+++ b/osu.Server.Spectator/GracefulShutdownManager.cs
@@ -1,6 +1,7 @@
 // Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
 // See the LICENCE file in the repository root for full licence text.
 
+using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Threading;
@@ -15,6 +16,8 @@ namespace osu.Server.Spectator;
 /// </summary>
 public class GracefulShutdownManager
 {
+    public static readonly TimeSpan TIME_BEFORE_FORCEFUL_SHUTDOWN = TimeSpan.FromHours(6);
+
     private readonly List<IEntityStore> dependentStores = new List<IEntityStore>();
 
     private Task? shutdownTask;
@@ -52,12 +55,11 @@ public class GracefulShutdownManager
             foreach (var store in dependentStores)
                 store.StopAcceptingEntities();
 
-            int timeWaited = 0;
+            TimeSpan timeWaited = new TimeSpan();
 
-            const int seconds_between_checks = 10;
-            const int max_hours_before_forceful_shutdown = 6;
+            TimeSpan timeBetweenChecks = TimeSpan.FromSeconds(10);
 
-            while (timeWaited < 3600 * max_hours_before_forceful_shutdown)
+            while (timeWaited < TIME_BEFORE_FORCEFUL_SHUTDOWN)
             {
                 var remaining = dependentStores.Select(store => (store.EntityName, store.RemainingUsages));
 
@@ -68,8 +70,8 @@ public class GracefulShutdownManager
                 foreach (var r in remaining)
                     Logger.Log($"{r.EntityName,10}: {r.RemainingUsages}");
 
-                Thread.Sleep(seconds_between_checks * 1000);
-                timeWaited += seconds_between_checks;
+                Thread.Sleep(timeBetweenChecks);
+                timeWaited = timeWaited.Add(timeBetweenChecks);
             }
 
             Logger.Log("All entities cleaned up. Server shutdown unblocking.");
--- a/osu.Server.Spectator/GracefulShutdownManager.cs
+++ b/osu.Server.Spectator/GracefulShutdownManager.cs
@@ -10,6 +10,9 @@ using osu.Server.Spectator.Entities;
 
 namespace osu.Server.Spectator;
 
+/// <summary>
+/// Ensures that shutdown is delayed until any existing usages have ceased.
+/// </summary>
 public class GracefulShutdownManager
 {
     private readonly List<IEntityStore> dependentStores = new List<IEntityStore>();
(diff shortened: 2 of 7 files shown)


osu.Server.Spectator/Startup.cs
```diff
@@ -63,6 +66,9 @@ public void ConfigureServices(IServiceCollection services)
                 // IdentityModelEventSource.ShowPII = true;
             });
 
+            // Allow a bit of extra time in addition to the graceful shutdown window for asp.net level forced shutdown.
+            services.Configure<HostOptions>(opts => opts.ShutdownTimeout = GracefulShutdownManager.TIME_BEFORE_FORCEFUL_SHUTDOWN.Add(TimeSpan.FromMinutes(1)));
```
bdach, line 70: What is the significance of the added extra minute here? Seems rather insignificant given `TIME_BEFORE_FORCEFUL_SHUTDOWN` is 6 hours...?

osu.Server.Spectator/Startup.cs
```diff
@@ -80,8 +86,21 @@ protected virtual void ConfigureAuthentication(IServiceCollection services)
         }
 
         // This method gets called by the runtime. Use this method to configure the HTTP request pipeline.
-        public void Configure(IApplicationBuilder app, IWebHostEnvironment env)
+        public void Configure(IApplicationBuilder app, IWebHostEnvironment env, IHostApplicationLifetime lifetime)
         {
+            var shutdownManager = app.ApplicationServices.GetService<GracefulShutdownManager>();
+
+            Debug.Assert(shutdownManager != null);
```
bdach, line 93: You can actually skip the assertion by using `GetRequiredService<T>()` as already done below:

```suggestion
            var shutdownManager = app.ApplicationServices.GetRequiredService<GracefulShutdownManager>();
```

osu.Server.Spectator/Startup.cs
```diff
@@ -80,8 +86,21 @@ protected virtual void ConfigureAuthentication(IServiceCollection services)
...
+
+            // This is done here, but it happens too late in shutdown.
```
bdach, line 95: What does "happens too late" mean here?

osu.Server.Spectator/GracefulShutdownManager.cs
```diff
@@ -0,0 +1,80 @@
...
+    private readonly List<IEntityStore> dependentStores = new List<IEntityStore>();
+
+    private Task? shutdownTask;
+
+    /// <summary>
+    /// Blocks until safe to continue with shutdown. Can be invoked from multiple locations.
+    /// </summary>
+    public void WaitForSafeShutdown()
+    {
+        lock (this)
```
bdach, line 30: Was `lock (this)` intentional in this context? Usually this type of thing is done on a dummy private variable (`private object syncRoot = new object();` or similar), to avoid external code from taking the lock on this instance and potentially causing deadlocks.

osu.Server.Spectator/Entities/EntityStore.cs
```diff
@@ -24,6 +26,33 @@ public sealed class EntityStore<T>
 
         private string statsDPrefix => $"entities.{typeof(T).Name}";
 
+        private bool acceptingNewEntities = true;
+
+        public EntityStore(GracefulShutdownManager shutdownManager, IHostApplicationLifetime applicationLifetime)
+            : this()
+        {
+            // Registration is required here to ensure `ApplicationStopping` triggers the shutdown wait before SignalR disconnects clients.
+            // Hope we can find a better way to make this work in the future.
```
bdach, line 35: Not sure I see anything too offensive about doing this registration where it is. Where were you looking to put it?

smoogipoo: I don't understand this. You can inject dependencies into dependencies. So you could add a `GracefulShutdownManager(IHostApplicationLifetime applicationLifetime)` ctor afaik, which seems more appropriate?

osu.Server.Spectator/Startup.cs
```diff
@@ -80,8 +86,21 @@ protected virtual void ConfigureAuthentication(IServiceCollection services)
...
+            // The more important registration is in EntityStore.
+            lifetime.ApplicationStopping.Register(shutdownManager.WaitForSafeShutdown);
```
smoogipoo, line 97: Also don't see why this is required if https://github.com/ppy/osu-server-spectator/pull/133/files/681bed1d646394408b7f83cf9b9923dc8bf0bd95#r883196847 were applied.

osu.Server.Spectator/Startup.cs
```diff
@@ -80,8 +86,21 @@ protected virtual void ConfigureAuthentication(IServiceCollection services)
...
+
+            // Importantly, we don't care to block `MultiplayerClientState` because they can only be created if a `ServerMultiplayerRoom` is first in existence.
+            // More so, we want to allow these states to be created so existing rooms can continue to function until they are disbanded.
+            shutdownManager.AddDependentStore(app.ApplicationServices.GetRequiredService<EntityStore<ServerMultiplayerRoom>>());
+            shutdownManager.AddDependentStore(app.ApplicationServices.GetRequiredService<EntityStore<SpectatorClientState>>());
```
smoogipoo, line 102: What about `EntityStore<MultiplayerClientState>`? Given 1 `SpectatorClientState` is basically the same as a `MultiplayerClientState`, I don't see an issue with having it, and _not_ having it would still new allow clients to connect.

Also, technically you can do this inside the `GracefulShutdownManager` ctor too.

smoogipoo (changes requested):
My suggestion is the following:

```diff
diff --git a/osu.Server.Spectator/Entities/EntityStore.cs b/osu.Server.Spectator/Entities/EntityStore.cs
index 65d763b..d2df263 100644
--- a/osu.Server.Spectator/Entities/EntityStore.cs
+++ b/osu.Server.Spectator/Entities/EntityStore.cs
@@ -7,7 +7,6 @@ using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.SignalR;
-using Microsoft.Extensions.Hosting;
 using osu.Framework.Extensions.ObjectExtensions;
 using StatsdClient;
 
@@ -28,18 +27,6 @@ namespace osu.Server.Spectator.Entities
 
         private bool acceptingNewEntities = true;
 
-        public EntityStore(GracefulShutdownManager shutdownManager, IHostApplicationLifetime applicationLifetime)
-            : this()
-        {
-            // Registration is required here to ensure `ApplicationStopping` triggers the shutdown wait before SignalR disconnects clients.
-            // Hope we can find a better way to make this work in the future.
-            applicationLifetime.ApplicationStopping.Register(shutdownManager.WaitForSafeShutdown);
-        }
-
-        public EntityStore()
-        {
-        }
-
         public void StopAcceptingEntities() => acceptingNewEntities = false;
 
         public int RemainingUsages
diff --git a/osu.Server.Spectator/GracefulShutdownManager.cs b/osu.Server.Spectator/GracefulShutdownManager.cs
index 1667a4f..09ae778 100644
--- a/osu.Server.Spectator/GracefulShutdownManager.cs
+++ b/osu.Server.Spectator/GracefulShutdownManager.cs
@@ -6,8 +6,10 @@ using System.Collections.Generic;
 using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
+using Microsoft.Extensions.Hosting;
 using osu.Framework.Logging;
 using osu.Server.Spectator.Entities;
+using osu.Server.Spectator.Hubs;
 
 namespace osu.Server.Spectator;
 
@@ -22,6 +24,16 @@ public class GracefulShutdownManager
 
     private Task? shutdownTask;
 
+    public GracefulShutdownManager(IHostApplicationLifetime applicationLifetime, EntityStore<MultiplayerClientState> multiplayerClients, EntityStore<SpectatorClientState> spectatorClients,
+                                   EntityStore<ServerMultiplayerRoom> rooms)
+    {
+        applicationLifetime.ApplicationStopping.Register(WaitForSafeShutdown);
+
+        AddDependentStore(multiplayerClients);
+        AddDependentStore(spectatorClients);
+        AddDependentStore(rooms);
+    }
+
     /// <summary>
     /// Blocks until safe to continue with shutdown. Can be invoked from multiple locations.
     /// </summary>
diff --git a/osu.Server.Spectator/Startup.cs b/osu.Server.Spectator/Startup.cs
index bc56bfe..3610365 100644
--- a/osu.Server.Spectator/Startup.cs
+++ b/osu.Server.Spectator/Startup.cs
@@ -3,7 +3,6 @@
 
 using System;
 using System.Collections.Generic;
-using System.Diagnostics;
 using Microsoft.AspNetCore.Authentication.JwtBearer;
 using Microsoft.AspNetCore.Builder;
 using Microsoft.AspNetCore.Hosting;
@@ -15,7 +14,6 @@ using Microsoft.Extensions.Options;
 using Newtonsoft.Json;
 using osu.Game.Online;
 using osu.Server.Spectator.Authentication;
-using osu.Server.Spectator.Entities;
 using osu.Server.Spectator.Extensions;
 using osu.Server.Spectator.Hubs;
 
@@ -86,25 +84,13 @@ namespace osu.Server.Spectator
         }
 
         // This method gets called by the runtime. Use this method to configure the HTTP request pipeline.
-        public void Configure(IApplicationBuilder app, IWebHostEnvironment env, IHostApplicationLifetime lifetime)
+        public void Configure(IApplicationBuilder app, IWebHostEnvironment env)
         {
-            var shutdownManager = app.ApplicationServices.GetService<GracefulShutdownManager>();
-
-            Debug.Assert(shutdownManager != null);
-
-            // This is done here, but it happens too late in shutdown.
-            // The more important registration is in EntityStore.
-            lifetime.ApplicationStopping.Register(shutdownManager.WaitForSafeShutdown);
-
-            // Importantly, we don't care to block `MultiplayerClientState` because they can only be created if a `ServerMultiplayerRoom` is first in existence.
-            // More so, we want to allow these states to be created so existing rooms can continue to function until they are disbanded.
-            shutdownManager.AddDependentStore(app.ApplicationServices.GetRequiredService<EntityStore<ServerMultiplayerRoom>>());
-            shutdownManager.AddDependentStore(app.ApplicationServices.GetRequiredService<EntityStore<SpectatorClientState>>());
+            // Warm up the shutdown manager.
+            app.ApplicationServices.GetRequiredService<GracefulShutdownManager>();
 
             if (env.IsDevelopment())
-            {
                 app.UseDeveloperExceptionPage();
-            }
 
             app.UseRouting();
 ```

**Comment by peppy** [comment]
Does it work? I had exactly that along with three other iterations and could not make it work. SignalR would always exit first and drop all connections.

File: osu.Server.Spectator/Startup.cs (hunk 1 above)
Comment (peppy) [reply osu.Server.Spectator/Startup.cs:70]: This is to allow the server to finish the exit process. ie. dispose and clean up entities, remove database multiplayer rooms, etc.

File: osu.Server.Spectator/Startup.cs (hunk 3 above)
Comment (peppy) [reply osu.Server.Spectator/Startup.cs:95]: SignalR is shut down first, making the whole thing fruitless.

File: osu.Server.Spectator/GracefulShutdownManager.cs (hunk 4 above)
Comment (peppy) [reply osu.Server.Spectator/GracefulShutdownManager.cs:30]: It was intentional because nothing else is going to be locking this component. I can change it if required.

File: osu.Server.Spectator/Entities/EntityStore.cs (hunk 5 above)
Comment (peppy) [reply osu.Server.Spectator/Entities/EntityStore.cs:35]: Give it a try and you'll see why I had to structure as it is.

File: osu.Server.Spectator/Startup.cs (hunk 7 above)
Comment (peppy) [reply osu.Server.Spectator/Startup.cs:102]: I'm 99.9% sure there's an inline comment mentioning this.

Adding it will cause new clients to be *unable to join existing rooms*.

**Comment by smoogipoo** [comment]
It seems all you need to do that is move the service warmup to after the hub endpoints are mapped.

Here's the relevant issue + workaround comment from a SignalR maintainer: https://github.com/dotnet/aspnetcore/issues/25069#issuecomment-912817907

**Comment by peppy** [comment]
I did test precisely that but it wasn't working. I'll give it another try to confirm if you haven't already.

File: osu.Server.Spectator/Startup.cs (hunk 7 above)
Comment (smoogipoo) [reply osu.Server.Spectator/Startup.cs:102]: How is this going to be routed? There's going to be two servers up simultaneously, and I don't think we have any way to identify the build ID via the connection, right?

So within a 6 hour timespan, _every new user_ that connects, regardless of being on a new build or not, will not be able to create rooms or spectate, or have their replays saved server-side?

I thought we were going down a different path of preserving existing connections and moving new connections to the new server.

Comment (peppy) [reply osu.Server.Spectator/Startup.cs:102]: This PR makes it so:

- New rooms cannot be created
- New spectator sessions cannot be started

It leaves in place:

- Existing rooms can continue to be played
- Users still on the server can still connect to existing rooms

I touched on the next steps in the OP here:

> Eventually we may want to use multiple endpoints or sticky cookies/headers to persist server across connection loss (aka cloudflare support, since cloudflare arbitrarily drops websocket connections) but for now we are avoiding this by dropping cloudflare temporarily.

And the inline comment here:

https://github.com/ppy/osu-server-spectator/pull/133/files#diff-90658690a1d84ef42d78f75a985211e9140da7fdadbcb3b5562eee9c5d894fd7R171-R174

Please let me know how I could have worded this more clearly or if something is missing which caused your confusion.

Comment (peppy) [reply osu.Server.Spectator/Startup.cs:102]: > So within a 6 hour timespan, _every new user_ that connects, regardless of being on a new build or not, will not be able to create rooms or spectate, or have their replays saved server-side?

New users connect to a new server. No new connections can reach a terminating server, only already connected ones.

Comment (smoogipoo) [reply osu.Server.Spectator/Startup.cs:102]: So based on that quote what you're trying to say is that the intention is if someone has a power outage then they're screwed, but if their connection drops momentarily then they can join back in to the old server?

And it's an intention because we don't actually have support for this yet.

It just seems weird and overly complex to me. I'm going to guess that your reasoning for this is that players on the server should still be able to join rooms so as to not shaft tournaments, however what about players that don't even have their game started up? I guess they're just shafted anyway?

Comment (peppy) [reply osu.Server.Spectator/Startup.cs:102]: Connection dropping momentarily *cannot reconnect to the old server*. Did you want to voice? There should be nothing complicated about it.

Also there is no "shafting" involved. We already discussed having a window where tournaments would know not to make rooms. This does not resolve that issue. The only thing it does is allow initial deployment of automatic release handling. It's equivalent to what I've been doing manually but with some new safeties.

**Comment by peppy** [comment]
To cover voice discussions:

- This is still an initial step forward, mostly to match the existing process I'm using for deployment
- @smoogipoo would prefer new users can't connect to existing games, to promote they move to the new server (aka adding the `MultiplayClientState` to blocked states during shutdown) but this means users on the same server potentially can't join rooms where they could previously. So for now I'd avoid doing this.

Follow up work to be done in near future:
- Users in multiplayer rooms should be aware of the server migration (countdown timer visible on their screen)
- Users with no server state should receive a message from the server to trigger a reconnect (moving them to the new server sooner rather than later). Note that this has the same effect of disallowing users to connect to existing rooms where they may have previously been able to, so requires further consideration for how to communicate this at the room listing or otherwise.
- This flow change doesn't handle old (aka incompatible) client versions at all, and doesn't intend to do so. Old versions should be disallowed connection to the server via a negotiation process, and force their client to upgrade instead. This is different to how osu! has worked historically but will allow for agile development from our end (aka not supporting 5 different versions) and is a user expectation already (*every* other game does this).

**Comment by smoogipoo** [comment]
My takeaways from the voice discussion:

Goals:
- Clients should be updated ASAP, don't want to support multiple builds.
- According to osu!stable, users naturally update their clients pretty quickly.
- The drain on user experience should be minimised.
- The server implementation is meant to support more agile deploys. Upgrade connections first and only force a client update if incompatible with the new server.

Issues raised:
- Spectator connections are immediately upgraded upon a new play session
  - Q: Old clients are incompatible with the new server.  
    A: If detected to be incompatible with the new server the client update will be forced.
- For multiplayer sessions the current functionality is only intended to match @peppy's manual labouring.
  - Q: This will delay updating clients ASAP.
    A: Doesn't affect new connections and existing clients are expected to naturally update pretty quickly. The 6hr timeout is a best-effort guess and could be lowered this to something like 10 minutes if needed.
  - Q: Old multiplayer connections is used but the upgraded spectator connection is incompatible with the client version.  
    A: Will need to trigger a client update, but also send a message from server to client to upgrade the connection immediately rather than on failure.
  - Q: Rooms are created (osu!web-side) but the multiplayer connection fails.
    A: Will need some process to rejoin the room after upgrading the connection.
  - Q: Rooms on the old client will be forcefully shut down after a while.  
    A: Will have some management of multiplayer rooms to handle them being managed by different servers, whether it's migrating them or having them not shown, having a countdown, or otherwise. This is a parallel issue and not relevant to my primary concerns.

File: osu.Server.Spectator/Startup.cs (hunk 7 above)
Comment (smoogipoo) [reply osu.Server.Spectator/Startup.cs:102]: See discussions later on in the thread.

File: osu.Server.Spectator/Startup.cs (hunk 1 above)
Comment (peppy) [reply osu.Server.Spectator/Startup.cs]: I've added an extra note regarding this.

File: osu.Server.Spectator/Startup.cs (hunk 2 above)
Comment (peppy) [reply osu.Server.Spectator/Startup.cs]: Code no longer exists 🎉

File: osu.Server.Spectator/Startup.cs (hunk 3 above)
Comment (peppy) [reply osu.Server.Spectator/Startup.cs]: This also no longer exists.

File: osu.Server.Spectator/GracefulShutdownManager.cs (hunk 4 above)
Comment (peppy) [reply osu.Server.Spectator/GracefulShutdownManager.cs]: Has been removed as no longer required.

File: osu.Server.Spectator/Entities/EntityStore.cs (hunk 5 above)
Comment (peppy) [reply osu.Server.Spectator/Entities/EntityStore.cs]: Has been resolved with new knowledge of how to make `ApplicationStopping` arrive in the correct order.

**Comment by peppy** [comment]
I've refactored things using the workaround explained at https://github.com/dotnet/aspnetcore/issues/25069#issuecomment-912817907. Have re-tested the two scenarios I was testing previously and both still work as expected. Looking a lot cleaner now.

File: osu.Server.Spectator/GracefulShutdownManager.cs [hunk 8]
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    /// <summary>
+    /// Add an entity store which should be waited on for all usages to have finished.
+    /// </summary>
+    /// <param name="store"></param>
+    private void addDependentStore(IEntityStore? store)
+    {
+        if (store == null)
+            return;
```
smoogipoo, line 42: Since the stores should never be null, might as well inline this function in the ctor? Calling `dependentStores.Add()` directly.

smoogipoo (approved):

osu.Server.Spectator/GracefulShutdownManager.cs
peppy: Sure.

smoogipoo (approved):

PR #136, opened 2022-06-13
Add sentry exception filter for `HubException`

peppy:
These exceptions are intended for the end-user, and shouldn't be logged to sentry.

https://sentry.ppy.sh/organizations/ppy/issues/2771/?project=8&query=is%3Aunresolved&statsPeriod=14d

### DIFF
--- a/osu.Server.Spectator/Entities/EntityStore.cs
+++ b/osu.Server.Spectator/Entities/EntityStore.cs
@@ -6,7 +6,6 @@ using System.Collections.Generic;
 using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
-using Microsoft.AspNetCore.SignalR;
 using osu.Framework.Extensions.ObjectExtensions;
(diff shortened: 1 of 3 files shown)


smoogipoo:
In general with new code, I've been very conscious of when `InvalidStateException` is used as opposed to `InvalidOperationException`. So I do believe that the end goal is to filter on `HubException`s.

However, for the time being, perhaps the path of least friction is to special type the shutdown exception.

bdach:
Approaching this from another angle: even if `HubException`s are supposed to be intended for the end user only, there is generally still some value in having a degree of observability of the volume of these exceptions as it may point at an underlying failure of associated systems - while in general a server shutdown is always a manual maintenance task, it is well understood what it means by infrastructure operators, and so it doesn't ever really need to show up on sentry.

peppy:
Have made a separate exception type.

smoogipoo (approved):

PR #142, opened 2022-09-05
Add support for multiple countdowns in multiplayer

smoogipoo:
Prereqs:
- [x] https://github.com/ppy/osu/pull/20145

The general structure is the same and the core "run-time" part of countdowns hasn't changed. The important changes are:
- The new `CountdownInfo` class stores supporting information about countdowns that was previously stored at a class-level.
- Countdowns are no longer exclusive by default. The previous behaviour is preferred for the match start and gameplay start countdowns by overriding `MultiplayerCountdown.IsExclusive`.

### DIFF
--- a/SampleMultiplayerClient/SampleMultiplayerClient.csproj
+++ b/SampleMultiplayerClient/SampleMultiplayerClient.csproj
@@ -7,11 +7,11 @@
     </PropertyGroup>
 
     <ItemGroup>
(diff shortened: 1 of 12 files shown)


peppy:
@smoogipoo made some minor naming changes, can you check you're okay with them?

I'm not going to test this locally because it doesn't seem like it needs it, but will test on dev/staging once this is merged.

peppy (approved):

smoogipoo:
Lgtm!

PR #143, opened 2022-09-16
Send shutting down countdown during graceful shutdown period

smoogipoo:
Prereqs:
- [x] https://github.com/ppy/osu/pull/20342

### DIFF
--- a/osu.Server.Spectator/GracefulShutdownManager.cs
+++ b/osu.Server.Spectator/GracefulShutdownManager.cs
@@ -5,8 +5,11 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Threading;
+using Microsoft.AspNetCore.SignalR;
 using Microsoft.Extensions.Hosting;
 using osu.Framework.Logging;
+using osu.Game.Online.Multiplayer;
+using osu.Game.Online.Multiplayer.Countdown;
 using osu.Server.Spectator.Entities;
 using osu.Server.Spectator.Hubs;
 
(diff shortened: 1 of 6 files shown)


osu.Server.Spectator/GracefulShutdownManager.cs
```diff
@@ -39,12 +44,36 @@ private void shutdownSafely()
...
+            });
+        }).Wait();
 
+        TimeSpan timeWaited = new TimeSpan();
         TimeSpan timeBetweenChecks = TimeSpan.FromSeconds(10);
+        bool finalNotificationSent = false;
 
         while (timeWaited < TIME_BEFORE_FORCEFUL_SHUTDOWN)
         {
+            TimeSpan timeRemaining = TIME_BEFORE_FORCEFUL_SHUTDOWN - timeWaited;
```
peppy, line 61: Was this implementation chose due to ease of implementation? I would have thought that one countdown would ve sent to all rooms when shutdown started, and one would be sent to any users joining a room on a server which is shutting down (at the point of joining the room).

If that sounds too hard to implement we can go with what you have to start with.

smoogipoo: What part? This is only done to send the other notification 5 minutes before shutdown. It is _also_ sent to all rooms at the start of the countdown and also when they join the room.

Should I remove the countdown in this loop? I'm unsure of it existing.

peppy: I think just one notification type is fine (removing the boolean), as per discord discussion.

Ignore the rest of what I said, I missed that countdowns were sent as part of the room.

peppy:
Will fix the warning level suggestions in a separate pull.

PR #144, opened 2022-09-20
Ensure `UserFinishedPlaying` is never called with a `Playing` state

smoogipoo:
Fixes https://github.com/ppy/osu/issues/19061

If a user forcefully quits their game, or the game crashes, the multiplayer server could send back a `UserFinishedPlaying(State = Playing)` message.


peppy (approved):
Looks sane enough.

PR #150, opened 2022-10-31
Update dependencies

peppy:
Mostly for "Freeze Frame" mod support in multiplayer.


peppy:
Going to push this out

PR #153, opened 2022-12-12
Add score token to `BeginPlaySession()` + store it to state

smoogipoo:
Prereqs:
- [x] https://github.com/ppy/osu/pull/21588
- [x] Game package bump

PR #154, opened 2022-12-12
Add base class structure for a "score uploader"

smoogipoo:
Prereqs:
- [x] https://github.com/ppy/osu-server-spectator/pull/153
- [x] https://github.com/ppy/osu/pull/21587
- [x] https://github.com/ppy/osu/pull/21588

Following the structure set by `MetadataBroadcaster`, this PR adds a `ScoreUploader` class which performs the background process of waiting for scores to receive ids and doing an upload. Right now it doesn't actually "upload" and instead stores to files as it previously did, however the file name has been renamed to account for the solo-score ID.

https://user-images.githubusercontent.com/1329837/206978833-60d1a4bf-325b-466a-bfd0-7637378be3a4.mp4

### DIFF
--- a/osu.Server.Spectator/Hubs/ScoreUploader.cs
+++ b/osu.Server.Spectator/Hubs/ScoreUploader.cs
@@ -57,26 +57,28 @@ namespace osu.Server.Spectator.Hubs
 
                 using (var db = databaseFactory.GetInstance())
                 {
-                    while (queue.TryDequeue(out var item))
+                    int countToTry = queue.Count;
+
+                    for (int i = 0; i < countToTry; i++)
                     {
+                        if (!queue.TryDequeue(out var item))
+                            continue;
+
                         long? scoreId = await db.GetScoreIdFromToken(item.Token);
 
-                        if (scoreId == null)
+                        if (scoreId == null && !item.Cancellation.IsCancellationRequested)
                         {
-                            if (item.Cancellation.IsCancellationRequested)
-                            {
-                                Console.WriteLine($"Score upload timed out for token: {item.Token}");
-                                item.Dispose();
-                                continue;
-                            }
-
-                            // Score is not ready yet.
+                            // Score is not ready yet - enqueue for the next attempt.
                             queue.Enqueue(item);
+                            continue;
                         }
-                        else
+
+                        using (item)
                         {
-                            await uploadScore(scoreId.Value, item);
-                            item.Dispose();
+                            if (scoreId != null)
+                                await uploadScore(scoreId.Value, item);
+                            else
+                                Console.WriteLine($"Score upload timed out for token: {item.Token}");
                         }
                     }
                 }
--- a/osu.Server.Spectator/Hubs/ScoreUploader.cs
+++ b/osu.Server.Spectator/Hubs/ScoreUploader.cs
@@ -16,10 +16,19 @@ namespace osu.Server.Spectator.Hubs
     public class ScoreUploader : IEntityStore, IDisposable
     {
         /// <summary>
-        /// A timeout to drop scores which haven't had IDs assigned to their tokens.
+        /// Amount of time (in milliseconds) between checks for pending score uploads.
+        /// </summary>
+        public double UploadInterval
+        {
+            get => timer.Interval;
+            set => timer.Interval = value;
+        }
+
+        /// <summary>
+        /// Amount of time (in milliseconds) before any individual score times out if a score ID hasn't been set.
         /// This can happen if the user forcefully terminated the game before the API score submission request is sent, but after EndPlaySession() has been invoked.
         /// </summary>
-        private const int timeout_seconds = 30;
+        public double TimeoutInterval = 30000;
 
         private bool shouldSaveReplays => Environment.GetEnvironmentVariable("SAVE_REPLAYS") == "1";
 
@@ -57,7 +66,7 @@ namespace osu.Server.Spectator.Hubs
             Interlocked.Increment(ref remainingUsages);
 
             var cancellation = new CancellationTokenSource();
-            cancellation.CancelAfter(TimeSpan.FromSeconds(timeout_seconds));
+            cancellation.CancelAfter(TimeSpan.FromMilliseconds(TimeoutInterval));
 
             queue.Enqueue(new UploadItem(token, score, cancellation));
         }
--- a/osu.Server.Spectator/Hubs/ScoreUploader.cs
+++ b/osu.Server.Spectator/Hubs/ScoreUploader.cs
@@ -80,13 +80,16 @@ namespace osu.Server.Spectator.Hubs
                             continue;
                         }
 
-                        using (item)
+                        try
                         {
                             if (scoreId != null)
                                 await uploadScore(scoreId.Value, item);
                             else
                                 Console.WriteLine($"Score upload timed out for token: {item.Token}");
-
+                        }
+                        finally
+                        {
+                            item.Dispose();
                             Interlocked.Decrement(ref remainingUsages);
                         }
                     }
--- a/osu.Server.Spectator/Hubs/ScoreUploader.cs
+++ b/osu.Server.Spectator/Hubs/ScoreUploader.cs
@@ -4,19 +4,27 @@
 using System;
 using System.Collections.Concurrent;
 using System.IO;
+using System.Threading;
 using System.Threading.Tasks;
 using System.Timers;
 using osu.Game.Scoring;
 using osu.Game.Scoring.Legacy;
 using osu.Server.Spectator.Database;
+using Timer = System.Timers.Timer;
 
 namespace osu.Server.Spectator.Hubs
 {
     public class ScoreUploader : IDisposable
     {
+        /// <summary>
+        /// A timeout to drop scores which haven't had IDs assigned to their tokens.
+        /// This can happen if the user forcefully terminated the game before the API score submission request is sent, but after EndPlaySession() has been invoked.
+        /// </summary>
+        private const int timeout_seconds = 30;
+
         // private bool shouldSaveReplays => Environment.GetEnvironmentVariable("SAVE_REPLAYS") == "1";
 
-        private readonly ConcurrentQueue<(long token, Score score)> queue = new ConcurrentQueue<(long token, Score score)>();
+        private readonly ConcurrentQueue<UploadItem> queue = new ConcurrentQueue<UploadItem>();
         private readonly IDatabaseFactory databaseFactory;
         private readonly Timer timer;
 
@@ -30,7 +38,13 @@ namespace osu.Server.Spectator.Hubs
             timer.Start();
         }
 
-        public void Enqueue(long token, Score score) => queue.Enqueue((token, score));
+        public void Enqueue(long token, Score score)
+        {
+            var cancellation = new CancellationTokenSource();
+            cancellation.CancelAfter(TimeSpan.FromSeconds(timeout_seconds));
+
+            queue.Enqueue(new UploadItem(token, score, cancellation));
+        }
 
         private async void update(object? sender, ElapsedEventArgs args)
         {
@@ -45,16 +59,24 @@ namespace osu.Server.Spectator.Hubs
                 {
                     while (queue.TryDequeue(out var item))
                     {
-                        long? scoreId = await db.GetScoreIdFromToken(item.token);
+                        long? scoreId = await db.GetScoreIdFromToken(item.Token);
 
                         if (scoreId == null)
                         {
+                            if (item.Cancellation.IsCancellationRequested)
+                            {
+                                Console.WriteLine($"Score upload timed out for token: {item.Token}");
+                                item.Dispose();
+                                continue;
+                            }
+
                             // Score is not ready yet.
                             queue.Enqueue(item);
                         }
                         else
                         {
-                            await uploadScore(scoreId.Value, item.score);
+                            await uploadScore(scoreId.Value, item);
+                            item.Dispose();
                         }
                     }
                 }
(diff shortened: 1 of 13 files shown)


osu.Server.Spectator/Hubs/ScoreUploader.cs
```diff
@@ -0,0 +1,130 @@
...
+                        long? scoreId = await db.GetScoreIdFromToken(item.Token);
+
+                        if (scoreId == null && !item.Cancellation.IsCancellationRequested)
+                        {
+                            // Score is not ready yet - enqueue for the next attempt.
+                            queue.Enqueue(item);
+                            continue;
+                        }
+
+                        using (item)
```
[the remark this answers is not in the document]
smoogipoo: I went through some 5 different versions of these 15 lines of code just because of the disposal. This is the best I've come up with.

osu.Server.Spectator/Hubs/ScoreUploader.cs
```diff
@@ -0,0 +1,130 @@
...
+
+            using (var outStream = File.Create(Path.Combine(path, filename)))
+                legacyEncoder.Encode(outStream);
+        }
+
+        public void Dispose()
+        {
+            timer.Stop();
+            timer.Dispose();
+        }
```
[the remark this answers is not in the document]
smoogipoo: Although these components are not being disposed at all... I've changed it to use a cancellation token for safety.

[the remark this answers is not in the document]
bdach: They're not getting disposed? Weird, I checked manually and I did see a hit after I ctrl-c'd a running instance...

bdach (dismissed):

osu.Server.Spectator/Hubs/ScoreUploader.cs
```diff
@@ -0,0 +1,153 @@
...
+        {
+            public void Dispose()
+            {
+                Cancellation.Dispose();
+            }
+        }
+
+        private int remainingUsages;
+
+        public int RemainingUsages => remainingUsages;
```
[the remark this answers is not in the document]
smoogipoo: Exactly. However now that I say this, if an error occurred during upload then the count will be incorrect. I'll add a comment + fix.

peppy:
@smoogipoo can you check on tests here?

smoogipoo:
Hold off on this one for a while longer, I want to add more tests.

smoogipoo:
Should be ready to go, depending on test failure (can't repro locally - adjusting timing in efforts to avoid on CI).

osu.Server.Spectator/Hubs/ScoreUploader.cs
```diff
@@ -0,0 +1,163 @@
...
+            set => timer.Interval = value;
+        }
+
+        /// <summary>
+        /// Amount of time (in milliseconds) before any individual score times out if a score ID hasn't been set.
+        /// This can happen if the user forcefully terminated the game before the API score submission request is sent, but after EndPlaySession() has been invoked.
+        /// </summary>
+        public double TimeoutInterval = 30000;
+
+        private bool shouldSaveReplays => Environment.GetEnvironmentVariable("SAVE_REPLAYS") == "1";
```
peppy, line 33: Unless you're intending for this to change at runtime (seems weird?)

```suggestion
        private bool shouldSaveReplays { get; } = Environment.GetEnvironmentVariable("SAVE_REPLAYS") == "1";
```

osu.Server.Spectator/Hubs/ScoreUploader.cs
```diff
@@ -0,0 +1,163 @@
...
+
+        private readonly ConcurrentQueue<UploadItem> queue = new ConcurrentQueue<UploadItem>();
+        private readonly IDatabaseFactory databaseFactory;
+        private readonly IScoreStorage scoreStorage;
+        private readonly Timer timer;
+        private readonly CancellationTokenSource timerCancellationSource;
+        private readonly CancellationToken timerCancellationToken;
```
peppy, line 40: Probably just remove this for simplicity, as `timerCancellationSource` is `readonly`.

osu.Server.Spectator/Hubs/ScoreUploader.cs
```diff
@@ -0,0 +1,163 @@
...
+            public void Dispose()
+            {
+                Cancellation.Dispose();
+            }
+        }
+
+        private int remainingUsages;
+
+        // Using the count of items in the queue isn't correct since items are dequeued for processing.
+        public int RemainingUsages => remainingUsages;
```
peppy, line 154: Probably use `TotalInFlight` to match `osu-queue-processor` terminology.

osu.Server.Spectator/Hubs/ScoreUploader.cs
```diff
@@ -0,0 +1,163 @@
...
+        }
+
+        /// <summary>
+        /// Flushes all pending uploads.
+        /// </summary>
+        public async Task Flush()
+        {
+            try
+            {
+                timer.Stop();
```
peppy, line 81: I don't believe this is required.

osu.Server.Spectator/Hubs/ScoreUploader.cs
```diff
@@ -0,0 +1,163 @@
...
+
+                if (queue.IsEmpty)
+                    return;
+
+                Console.WriteLine($"Beginning upload of {queue.Count} scores");
+
+                using (var db = databaseFactory.GetInstance())
+                {
+                    int countToTry = queue.Count;
```
peppy, line 90: I'm not sure about using a queue here rather than a `Bucket` or `HashSet`, feels a bit weird to dequeue then requeue potentially hundreds or thousands of items each loop.

osu.Server.Spectator/Hubs/ScoreUploader.cs
```diff
@@ -0,0 +1,163 @@
...
+            timerCancellationSource.Cancel();
+            timerCancellationSource.Dispose();
+        }
+
+        private record UploadItem(long Token, Score Score, CancellationTokenSource Cancellation) : IDisposable
+        {
+            public void Dispose()
+            {
+                Cancellation.Dispose();
+            }
```
peppy, line 148: I'm not sure I understand this disposal flow. Is there a valid reason to require this? The only case I can see this cancellation used is much before disposal. Is it just to clean up `CancellationTokenSource` because it's `IDisposable`?

osu.Server.Spectator/Hubs/ScoreUploader.cs
```diff
@@ -0,0 +1,163 @@
...
+            if (!shouldSaveReplays)
+                return;
+
+            Interlocked.Increment(ref remainingUsages);
+
+            var cancellation = new CancellationTokenSource();
+            cancellation.CancelAfter(TimeSpan.FromMilliseconds(TimeoutInterval));
+
+            queue.Enqueue(new UploadItem(token, score, cancellation));
+        }
```
peppy, line 72: I get that this is meant to handle the case where the user hasn't completed score submission, but if the server had a large backlog (or was waiting on slow s3 operations), it feels like this could cause unexpected failures. ie. there's a chance the upload is never attempted even once, if other uploads in the queue take over 30 seconds in total.

osu.Server.Spectator/Hubs/ScoreUploader.cs
```diff
@@ -0,0 +1,163 @@
...
+
+        public ScoreUploader(IDatabaseFactory databaseFactory, IScoreStorage scoreStorage)
+        {
+            this.databaseFactory = databaseFactory;
+            this.scoreStorage = scoreStorage;
+
+            timerCancellationSource = new CancellationTokenSource();
+            timerCancellationToken = timerCancellationSource.Token;
+
+            timer = new Timer(5000);
```
peppy, line 50: I think this needs to run more like every 50 or 100 ms. We'd want scores to be uploaded ASAP, else users may visit the score page and have no replay available for download.

osu.Server.Spectator/Hubs/ScoreUploader.cs
```diff
@@ -0,0 +1,163 @@
...
+                    }
+                }
+            }
+            catch (Exception e)
+            {
+                Console.WriteLine($"Error during score upload: {e}");
+            }
+            finally
+            {
+                if (timerCancellationToken.IsCancellationRequested)
```
peppy, line 130: Doesn't this mean the server could shut down with items still in the queue?

peppy (commented):
This PR feels pretty scary to me, because it diverges from how we're doing queue processing elsewhere.

Part of me wants to see `osu-queue-processor` used in place here, but not sure how well that would slot in either. The logic there is tried and proven to be correct to ensure items are processed correctly and not lost in the nether.

osu.Server.Spectator/Hubs/ScoreUploader.cs
smoogipoo: Accessing `CancellationTokenSource.Token` after disposal is an invalid operation.

osu.Server.Spectator/Hubs/ScoreUploader.cs
smoogipoo: It's not, but I thought it best to have for tests.

osu.Server.Spectator/Hubs/ScoreUploader.cs
smoogipoo: It is structured so that the score ID is fetched from the database first, and _only if_ it is `null` does the cancellation get considered.

osu.Server.Spectator/Hubs/ScoreUploader.cs
smoogipoo: That's what `RemainingUsages` is for. It behaves like every other entity store in waiting for the count to reach 0 before shutdown.

osu.Server.Spectator/Hubs/ScoreUploader.cs
smoogipoo: Hmmm, this is just because it implements `IEntityStore`. Perhaps I should use a different interface?

smoogipoo:
> Part of me wants to see `osu-queue-processor` used in place here

I've suggested using redis more probably about 50 times at this point between us, so I'm not against it. I envision it like `server-spectator` -> pushes scores + frames to queue -> `replay-processor` -> pulls items and does basically the same process as here.

Is that what you're envisioning also? Not sure how complicated that's going to be from the perspective of serialisation.

smoogipoo:
Re: queue processor: It kinda works? https://github.com/smoogipoo/osu-server-spectator/tree/replay-processor-queue

Only thing I'm unsure about is how to handle retries for the score ID retrieval.

osu.Server.Spectator/Hubs/ScoreUploader.cs
smoogipoo: In practice, this is running on two threads - the signalr thread and the timer thread. The simplest I can come up with this this:

```diff
diff --git a/osu.Server.Spectator/Hubs/ScoreUploader.cs b/osu.Server.Spectator/Hubs/ScoreUploader.cs
index f45f332..1b20417 100644
--- a/osu.Server.Spectator/Hubs/ScoreUploader.cs
+++ b/osu.Server.Spectator/Hubs/ScoreUploader.cs
@@ -3,6 +3,7 @@

 using System;
 using System.Collections.Concurrent;
+using System.Collections.Generic;
 using System.Threading;
 using System.Threading.Tasks;
 using osu.Game.Scoring;
@@ -32,7 +33,7 @@ namespace osu.Server.Spectator.Hubs

         private bool shouldSaveReplays => Environment.GetEnvironmentVariable("SAVE_REPLAYS") == "1";

-        private readonly ConcurrentQueue<UploadItem> queue = new ConcurrentQueue<UploadItem>();
+        private readonly List<UploadItem> queue = new List<UploadItem>();
         private readonly IDatabaseFactory databaseFactory;
         private readonly IScoreStorage scoreStorage;
         private readonly Timer timer;
@@ -68,7 +69,7 @@ namespace osu.Server.Spectator.Hubs
             var cancellation = new CancellationTokenSource();
             cancellation.CancelAfter(TimeSpan.FromMilliseconds(TimeoutInterval));

-            queue.Enqueue(new UploadItem(token, score, cancellation));
+            enqueue(new UploadItem(token, score, cancellation));
         }

         /// <summary>
@@ -80,26 +81,26 @@ namespace osu.Server.Spectator.Hubs
             {
                 timer.Stop();

-                if (queue.IsEmpty)
+                UploadItem[] items;
+
+                lock (queue)
+                    items = queue.ToArray();
+
+                if (items.Length == 0)
                     return;

-                Console.WriteLine($"Beginning upload of {queue.Count} scores");
+                Console.WriteLine($"Beginning upload of {items.Length} scores");

                 using (var db = databaseFactory.GetInstance())
                 {
-                    int countToTry = queue.Count;
-
-                    for (int i = 0; i < countToTry; i++)
+                    foreach (var item in items)
                     {
-                        if (!queue.TryDequeue(out var item))
-                            continue;
-

```

Is that what you're expecting? If not, perhaps it's easier if you implement it yourself.

osu.Server.Spectator/Hubs/ScoreUploader.cs
peppy: Let's leave it for now.

osu.Server.Spectator/Hubs/ScoreUploader.cs
peppy: Nah, that's not really any better. I'm just going to turn a blind eye here because it's likely going to change in the future and what you have, while isn't pretty/efficient, does work.

peppy (approved):

PR #155, opened 2022-12-13
Add S3 upload for replays

smoogipoo:
Prereqs:
- [x] https://github.com/ppy/osu-server-spectator/pull/154
- [x] https://github.com/ppy/osu/pull/21627

Environment variables:
`SAVE_REPLAYS`: Needs to be 1.
`S3_KEY`: AWS key.
`S3_SECRET`: AWS secret.
`REPLAYS_BUCKET`: Bucket name.

https://user-images.githubusercontent.com/1329837/207252484-34490c37-3437-401a-92ab-f6eb4e010501.mp4

### DIFF
--- a/osu.Server.Spectator/S3.cs
+++ b/osu.Server.Spectator/S3.cs
@@ -1,0 +1,105 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using Amazon;
+using Amazon.Runtime;
+using Amazon.S3;
+using Amazon.S3.Model;
+
+namespace osu.Server.Spectator
+{
+    public static class S3
+    {
+        public static byte[]? Retrieve(string bucket, string key, RegionEndpoint? endpoint = null)
+        {
+            using (var client = getClient(endpoint))
+            {
+                try
+                {
+                    var obj = client.GetObjectAsync(bucket, key).Result;
+
+                    using (var memory = new MemoryStream())
+                    {
+                        obj.ResponseStream.CopyTo(memory);
+                        return memory.ToArray();
+                    }
+                }
+                catch (Exception)
+                {
+                    return null;
+                }
+            }
+        }
+
+        public static bool CheckExists(string bucket, string key)
+        {
+            using (var client = getClient())
+            {
+                try
+                {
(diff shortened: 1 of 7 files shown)


osu.Server.Spectator/S3.cs
```diff
@@ -0,0 +1,105 @@
...
+using Amazon;
+using Amazon.Runtime;
+using Amazon.S3;
+using Amazon.S3.Model;
+
+namespace osu.Server.Spectator
+{
+    public static class S3
+    {
+        public static byte[]? Retrieve(string bucket, string key, RegionEndpoint? endpoint = null)
```
bdach, line 17: Bit peculiar that this is the only method that accepts a `RegionEndpoint` and the rest just defaults to `USWest1`.

osu.Server.Spectator/S3.cs
```diff
@@ -0,0 +1,105 @@
...
+                    BucketName = bucket,
+                    Key = key,
+                    CannedACL = S3CannedACL.PublicRead,
+                    Headers =
+                    {
+                        ContentLength = contentLength,
+                        ContentType = contentType,
+                    },
+                    InputStream = stream
+                }).Wait();
```
bdach, line 101: Is there a reason why all of these methods are fully blocking and synchronous? It's especially interesting given that `uploadScore()` - the only caller of any of these so far - is async, so it *looks* like it would support await perfectly fine with zero changes required.

smoogipoo: Because these were copied from another one of our projects :D will fix

osu.Server.Spectator/S3.cs
smoogipoo: I’ll remove the other methods, this entire class is a carbon copy from another project that needed these.

osu.Server.Spectator/Storage/S3ScoreStorage.cs
```diff
@@ -0,0 +1,29 @@
...
+    {
+        public async Task WriteAsync(Score score)
+        {
+            using (var outStream = new MemoryStream())
+            {
+                new LegacyScoreEncoder(score, null).Encode(outStream, true);
+
+                outStream.Seek(0, SeekOrigin.Begin);
+
+                Console.WriteLine($"Uploading replay for score {score.ScoreInfo.OnlineID}");
```
peppy, line 23: Not sure we want this level of console output, but will leave it for now and can remove after deploy as required.

peppy (approved):

PR #157, opened 2022-12-20
Send notifications about score processing completion to clients

bdach:
- Continuation of https://github.com/ppy/osu/issues/18877
- [x] Depends on https://github.com/ppy/osu/pull/21738
- [x] Depends on game package bump
- [x] Soft-depends on https://github.com/ppy/osu/pull/21753 (for correct delivery of notifications for aborted scores)

**Note:** This is an RFC. It has been tested _very_ lightly. The happy path seems to work, but I'm not sure about the multitude of sad paths. I'm PRing this now to get it out for comments and see how many WTFs I get, as I'm not confident about almost everything in this diff, so I wouldn't necessarily advise to hurry this one up too much.

Testing procedure, for whoever is inclined:

- Begin with a working full-stack `osu` + `osu-web` + `osu-server-spectator` setup
- In `osu`, check out pull mentioned above
- In `osu-server-spectator`, switch to local checkout
- Set up & launch `osu-queue-score-statistics` so that it can pick up scores placed in redis by `osu-web`
- At this point messages will flow, but there will be no visible effect game-side. To make one happen, the following crude patch can be applied:

<details>
<summary>patch to apply to game</summary>

```diff
diff --git a/osu.Game/OsuGame.cs b/osu.Game/OsuGame.cs
index a0a45e18a8..45c94b73d2 100644
--- a/osu.Game/OsuGame.cs
+++ b/osu.Game/OsuGame.cs
@@ -990,6 +990,11 @@ protected override void LoadComplete()
 
             // Importantly, this should be run after binding PostNotification to the import handlers so they can present the import after game startup.
             handleStartupImport();
+
+            SpectatorClient.OnUserScoreProcessed += (userId, scoreId) =>
+            {
+                Notifications.Post(new SimpleNotification { Text = $"OnUserScoreProcessed (userId: {userId}, scoreId: {scoreId})" });
+            };
         }
 
         private void handleStartupImport()
diff --git a/osu.Game/OsuGameBase.cs b/osu.Game/OsuGameBase.cs
index 7586dc6407..468bc61696 100644
--- a/osu.Game/OsuGameBase.cs
+++ b/osu.Game/OsuGameBase.cs
@@ -188,7 +188,7 @@ public virtual string Version
 
         private RulesetConfigCache rulesetConfigCache;
 
-        private SpectatorClient spectatorClient;
+        protected SpectatorClient SpectatorClient { get; private set; }
 
         protected MultiplayerClient MultiplayerClient { get; private set; }
 
@@ -298,7 +298,7 @@ private void load(ReadableKeyCombinationProvider keyCombinationProvider)
 
             // TODO: OsuGame or OsuGameBase?
             dependencies.CacheAs(beatmapUpdater = new BeatmapUpdater(BeatmapManager, difficultyCache, API, Storage));
-            dependencies.CacheAs(spectatorClient = new OnlineSpectatorClient(endpoints));
+            dependencies.CacheAs(SpectatorClient = new OnlineSpectatorClient(endpoints));
             dependencies.CacheAs(MultiplayerClient = new OnlineMultiplayerClient(endpoints));
             dependencies.CacheAs(metadataClient = new OnlineMetadataClient(endpoints));
 
@@ -343,7 +343,7 @@ private void load(ReadableKeyCombinationProvider keyCombinationProvider)
             if (API is APIAccess apiAccess)
                 AddInternal(apiAccess);
 
-            AddInternal(spectatorClient);
+            AddInternal(SpectatorClient);
             AddInternal(MultiplayerClient);
             AddInternal(metadataClient);
 
```

</details>

At this point, every completed play should result in a notification being fired. Aborted plays won't always fire the notification. I'm not sure why yet, I'm going to look into that a bit more still tomorrow if this approach doesn't just get dumpstered immediately.

I'll be adding some self-commentary review comments shortly.

### DIFF
--- a/osu.Server.Spectator/Hubs/ScoreProcessedSubscriber.cs
+++ b/osu.Server.Spectator/Hubs/ScoreProcessedSubscriber.cs
@@ -1,0 +1,176 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+using System;
+using System.Collections.Concurrent;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+using Newtonsoft.Json;
+using osu.Framework.Logging;
+using osu.Server.Spectator.Database;
+using StackExchange.Redis;
+using StatsdClient;
+using Timer = System.Timers.Timer;
+
+namespace osu.Server.Spectator.Hubs;
+
+public sealed class ScoreProcessedSubscriber : IScoreProcessedSubscriber, IDisposable
+{
+    /// <summary>
+    /// The maximum amount of time to wait for a <see cref="ScoreProcessed"/> message for a given score in milliseconds.
+    /// </summary>
+    private const int timeout_interval_ms = 30_000;
+
+    private const string statsd_prefix = "subscribers.score-processed";
+
+    private readonly IDatabaseFactory databaseFactory;
+    private readonly ISubscriber? subscriber;
+
+    private readonly ConcurrentDictionary<long, ScoreProcessedSubscription> subscriptions = new ConcurrentDictionary<long, ScoreProcessedSubscription>();
+    private readonly Timer timer;
+    private readonly Logger logger;
+
+    public ScoreProcessedSubscriber(
+        IDatabaseFactory databaseFactory,
+        IConnectionMultiplexer redis)
+    {
+        this.databaseFactory = databaseFactory;
+
+        timer = new Timer(1000);
+        timer.AutoReset = true;
+        timer.Elapsed += (_, _) => Task.Run(purgeTimedOutSubscriptions);
+        timer.Start();
+
+        subscriber = redis.GetSubscriber();
+        subscriber.Subscribe("osu-channel:score:processed", (_, message) => onMessageReceived(message));
+
+        logger = Logger.GetLogger(nameof(ScoreProcessedSubscriber));
+    }
+
+    private void onMessageReceived(string message)
+    {
+        try
+        {
+            var scoreProcessed = JsonConvert.DeserializeObject<ScoreProcessed>(message);
+
+            if (scoreProcessed == null)
+                return;
+
+            if (subscriptions.TryRemove(scoreProcessed.ScoreId, out var subscription))
+            {
+                using (subscription)
+                    subscription.InvokeAsync().Wait();
+            }
+
+            DogStatsd.Increment($"{statsd_prefix}.messages.delivered");
+        }
+        catch (Exception ex)
+        {
+            logger.Add($"Failed to process message {message}", LogLevel.Error, ex);
+            DogStatsd.Increment($"{statsd_prefix}.messages.dropped");
+        }
+    }
+
+    public async Task RegisterForNotificationAsync(string receiverConnectionId, int userId, long scoreToken, ScoreProcessedAsyncCallback callback)
+    {
+        try
+        {
+            using var db = databaseFactory.GetInstance();
+
+            long? scoreId = await db.GetScoreIdFromToken(scoreToken);
+
+            if (scoreId == null)
+            {
+                DogStatsd.Increment($"{statsd_prefix}.subscriptions.dropped");
+                return;
+            }
+
+            var subscription = new ScoreProcessedSubscription(receiverConnectionId, userId, scoreId.Value, callback);
+
+            // because the score submission flow happens concurrently with the spectator play finished flow,
+            // it is theoretically possible for the score processing to complete before the spectator hub had a chance to register for notifications.
+            // to cover off this possibility, check the database directly once.
+            if (await db.IsScoreProcessedAsync(scoreId.Value))
+            {
+                using (subscription)
+                    await subscription.InvokeAsync();
+                DogStatsd.Increment($"{statsd_prefix}.messages.delivered-immediately");
+                return;
+            }
+
+            subscriptions.TryAdd(scoreId.Value, subscription);
+            DogStatsd.Gauge($"{statsd_prefix}.subscriptions.total", subscriptions.Count);
+        }
+        catch (Exception ex)
+        {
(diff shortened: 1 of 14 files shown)


osu.Server.Spectator/Hubs/ScoreProcessedSubscriber.cs
```diff
@@ -0,0 +1,176 @@
...
+            DogStatsd.Gauge($"{statsd_prefix}.subscriptions.total", subscriptions.Count);
+        }
+        catch (Exception ex)
+        {
+            logger.Add($"Failed to register connection {receiverConnectionId} for info about score {userId}:{scoreToken}", LogLevel.Error, ex);
+            DogStatsd.Increment($"{statsd_prefix}.subscriptions.failed");
+        }
+    }
+
+    private void purgeTimedOutSubscriptions()
```
bdach, line 112: I've added this to avoid a situation where the subscriber gets backlogged due to the score processor not being able to catch up to incoming scores, causing a cascading failure. I'm not sure how realistic of a scenario this is in practice.

osu.Server.Spectator/Hubs/ScoreProcessedSubscriber.cs
```diff
@@ -0,0 +1,176 @@
...
+        {
+            var scoreProcessed = JsonConvert.DeserializeObject<ScoreProcessed>(message);
+
+            if (scoreProcessed == null)
+                return;
+
+            if (subscriptions.TryRemove(scoreProcessed.ScoreId, out var subscription))
+            {
+                using (subscription)
+                    subscription.InvokeAsync().Wait();
```
bdach, line 63: The `.Wait()` is unfortunate, but I'm not sure how to avoid it.

osu.Server.Spectator/Hubs/SpectatorHub.cs
```diff
@@ -119,6 +126,8 @@ public async Task EndPlaySession(SpectatorState state)
                     score.ScoreInfo.Date = DateTimeOffset.UtcNow;
 
                     scoreUploader.Enqueue(scoreToken.Value, score);
+                    await scoreProcessedSubscriber.RegisterForNotificationAsync(Context.ConnectionId, CurrentContextUserId, scoreToken.Value,
+                        (connectionId, userId, scoreId) => Clients.Client(connectionId).UserScoreProcessed(userId, scoreId));
```
bdach, line 130: This callback has a closure to the hub instance. I'm not sure if I'm allowed to do this, but I'm not sure how *else* I'm supposed to be delivering the async message otherwise. I guess I could have some sort of memory-based polling wherein the hub queries the subscriber if it has anything new to notify about? But can you even have a recurring timer in a hub? No idea.

Do note though that I am attempting to be extra careful to capture all of the ambient state of the hub (`Context.ConnectionId`, `CurrentContextUserId`) at the point of call. Not doing so would risk misfiring the message to a completely random user at some point in the future.

bdach: Never mind I am dumb and didn't think hard enough what to google. Googling "hub lifetime" spells out [pretty clearly](https://github.com/dotnet/AspNetCore.Docs/issues/6888) that I was doing Bad Things. Thankfully there's also https://learn.microsoft.com/en-us/aspnet/core/signalr/hubcontext?view=aspnetcore-2.1 and it works and I fixed it.

bdach:
> Aborted plays won't always fire the notification. I'm not sure why yet

I've figured out what this was; see https://github.com/ppy/osu/pull/21753 for explanation and proposed fix.

osu.Server.Spectator/Hubs/ScoreProcessedSubscriber.cs
```diff
@@ -0,0 +1,182 @@
...
+using Timer = System.Timers.Timer;
+
+namespace osu.Server.Spectator.Hubs;
+
+public sealed class ScoreProcessedSubscriber : IScoreProcessedSubscriber, IDisposable
+{
+    /// <summary>
+    /// The maximum amount of time to wait for a <see cref="ScoreProcessed"/> message for a given score in milliseconds.
+    /// </summary>
+    private const int timeout_interval_ms = 30_000;
```
peppy, line 25: We'll see how this plays out. I have a feeling going higher than this will be what we want eventually (minutes at least) to allow for potential server-side issues/outages.

osu.Server.Spectator/Hubs/ScoreProcessedSubscriber.cs
peppy: Probably fine.

osu.Server.Spectator/Hubs/ScoreProcessedSubscriber.cs
peppy: Going to have to see how this plays out post deploy really.

peppy (approved):

osu.Server.Spectator/Hubs/ScoreProcessedSubscriber.cs
```diff
@@ -0,0 +1,182 @@
...
+    }
+
+    private void purgeTimedOutSubscriptions()
+    {
+        var scoreIds = subscriptions.Keys.ToArray();
+        int purgedCount = 0;
+
+        foreach (var scoreId in scoreIds)
+        {
+            if (subscriptions.TryGetValue(scoreId, out var subscription) && subscription.TimedOut)
```
smoogipoo, line 124: One thing to watch out for is that `CancellationTokenSource.Token` throws an exception if the CTS has been disposed. There could theoretically be a small moment where this `TryGetValue()` retrieves the value on this thread, but then the `using (subscription)` above runs and only after that does the `subscription.TimedOut` check happen.

bdach: I was aware of this, but when I checked the implementation, then I didn't expect an `IsCancellationRequested` check to throw after disposal. Maybe bad form relying on implementation details.

smoogipoo: No you're correct, I didn't know there was such a discrepancy between `Token` and `IsCancellationRequested`

PR #160, opened 2022-12-28
Increase limit of queued beatmaps a host can add in host-only mode to 50

### LINKED ISSUE #159: Host only mode limits host to 3 beatmaps maximum
It shouldn't have any limit. The 3 beatmap limit is intended for other queue modes.

peppy:
Closes #159.

### DIFF
--- a/osu.Server.Spectator.Tests/Multiplayer/MultiplayerAllPlayersQueueTests.cs
+++ b/osu.Server.Spectator.Tests/Multiplayer/MultiplayerAllPlayersQueueTests.cs
@@ -179,24 +179,6 @@ namespace osu.Server.Spectator.Tests.Multiplayer
         }
 
         [Fact]
-        public async Task HostMayAddManyItems()
-        {
-            Database.Setup(d => d.GetBeatmapChecksumAsync(3333)).ReturnsAsync("3333");
-
-            await Hub.JoinRoom(ROOM_ID);
-            await Hub.ChangeSettings(new MultiplayerRoomSettings { QueueMode = QueueMode.HostOnly });
-
-            for (int i = 1; i < MultiplayerQueue.PER_USER_LIMIT + 1; i++)
-                await addItem();
-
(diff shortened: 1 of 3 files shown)


osu.Server.Spectator.Tests/Multiplayer/MultiplayerAllPlayersQueueTests.cs
```diff
@@ -179,7 +179,25 @@ public async Task OneNonExpiredItemExistsWhenChangingToHostOnlyMode()
...
         [Fact]
-        public async Task UserMayOnlyContainLimitedNumberOfItems()
+        public async Task HostMayAddManyItems()
+        {
+            Database.Setup(d => d.GetBeatmapChecksumAsync(3333)).ReturnsAsync("3333");
+
+            await Hub.JoinRoom(ROOM_ID);
+            await Hub.ChangeSettings(new MultiplayerRoomSettings { QueueMode = QueueMode.HostOnly });
+
+            for (int i = 1; i < MultiplayerQueue.PER_USER_LIMIT + 1; i++)
```
bdach, line 189: I actually got slightly confused at this and thought it may have been a copy-paste error, but it this number of repeats exceeds `PER_USER_LIMIT` so a new behaviour _is_ being actually tested here. Although the `HOST_LIMIT` isn't being exercised by this test but maybe that's beside the point.

bdach (approved):

osu.Server.Spectator.Tests/Multiplayer/MultiplayerAllPlayersQueueTests.cs
bdach: Actually one second, this test is in a bit of a wrong place...

bdach: Moved from `MultiplayerAllPlayersQueueTests` to `MultiplayerHostOnlyQueueTests` in baf6f95dec3f72271076b101904f984467ed1544.

PR #161, opened 2023-01-12
Don't store replays for unranked beatmaps

smoogipoo:
Resolves https://github.com/ppy/osu-queue-score-statistics/issues/128

### DIFF
--- a/osu.Server.Spectator/Hubs/SpectatorHub.cs
+++ b/osu.Server.Spectator/Hubs/SpectatorHub.cs
@@ -10,6 +10,7 @@ using osu.Game.Beatmaps;
 using osu.Game.Online.Spectator;
 using osu.Game.Scoring;
 using osu.Server.Spectator.Database;
+using osu.Server.Spectator.Database.Models;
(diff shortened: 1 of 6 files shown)


osu.Server.Spectator/Hubs/SpectatorHub.cs
```diff
@@ -123,6 +125,11 @@ public async Task EndPlaySession(SpectatorState state)
                     if (score == null || scoreToken == null)
                         return;
 
+                    // Do nothing with scores on unranked beatmaps.
+                    var status = score.ScoreInfo.BeatmapInfo.Status;
+                    if (status != BeatmapOnlineStatus.Ranked && status != BeatmapOnlineStatus.Approved)
```
peppy, line 130: Should loved also have replays?

stanriders: Qualified as well

peppy:
Tests look to be failing specifically on this branch.

peppy (approved):

PR #168, opened 2023-05-12
Switch BouncyCastle to BouncyCastle.NetCore

timiimit:
Currently a deprecated [BouncyCastle](https://www.nuget.org/packages/BouncyCastle) package is being used which only supports .NET Framework, so it can only be used on Windows.

This PR makes a switch to [BouncyCastle.NetCore](https://www.nuget.org/packages/BouncyCastle.NetCore) which makes this project cross-platform.

### DIFF
--- a/osu.Server.Spectator/osu.Server.Spectator.csproj
+++ b/osu.Server.Spectator/osu.Server.Spectator.csproj
@@ -7,7 +7,7 @@
 
     <ItemGroup>
         <PackageReference Include="AWSSDK.S3" Version="3.7.103.36" />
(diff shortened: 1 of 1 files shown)


peppy (approved):

smoogipoo:
I... have no idea what `BouncyCastle.NetCore` is. It looks like a fork (https://github.com/chrishaly/bc-csharp) that hasn't had its master branch updated in 8-10 years, and there's no branch for the 1.9.0 version.

Shouldn't we be using https://www.nuget.org/packages/BouncyCastle.Cryptography instead?

peppy:
Does look like a better choice.

PR #169, opened 2023-05-14
Assign multiplayer star rating

timiimit:
- [x] Depends on ppy/osu/pull/23551 & a game package with said PR
- [x] Depends on https://github.com/ppy/osu-server-spectator/pull/174 by necessity

This retrieves beatmap's `difficultyrating` from the db and uses it to populate `MultiplayerPlaylistItem` which is then used by the client to show latest information about room's star rating .

### DIFF
--- a/osu.Server.Spectator/Database/DatabaseAccess.cs
+++ b/osu.Server.Spectator/Database/DatabaseAccess.cs
@@ -67,11 +67,6 @@ namespace osu.Server.Spectator.Database
             });
         }
 
-        public async Task<string?> GetBeatmapChecksumAsync(int beatmapId)
-        {
-            return (await GetBeatmapAsync(beatmapId))?.checksum;
-        }
-
         public async Task MarkRoomActiveAsync(MultiplayerRoom room)
         {
             var connection = await getConnectionAsync();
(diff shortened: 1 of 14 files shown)


osu.Server.Spectator/Database/DatabaseAccess.cs
```diff
@@ -67,11 +67,6 @@ public async Task<multiplayer_room> GetRoomAsync(long roomId)
             });
         }
 
-        public async Task<string?> GetBeatmapChecksumAsync(int beatmapId)
-        {
-            return (await GetBeatmapAsync(beatmapId))?.checksum;
-        }
```
bdach, line 73: This method is extensively mocked in unit tests, so removing it and not adjusting the tests that mock `GetBeatmapChecksumAsync()` to mock `GetBeatmapAsync()` instead makes the unit tests not compile and also not pass.

timiimit: I replaced all uses of `GetBeatmapChecksumAsync()`.

bdach: You missed one, but I fixed it for you in ed21209d1f8a650c38d76d2c9695e19745a76d0c.

bdach:
Seems fine, but given the fact that this needs a whole lot of things to happen before merge (merge of https://github.com/ppy/osu-server-spectator/pull/174 first, then a game package with https://github.com/ppy/osu/pull/23551, which is going to be in the release-after-next), I'm just going to let this hang unapproved until CI can be actually unblocked.

bdach (approved):

peppy (approved):

PR #170, opened 2023-05-14
Force all non-spectators to play even when not explicitly "Ready"

timiimit:
Implements ppy/osu/issues/22090

It has been previously decided that players who don't press the ready button in multiplayer should still be pushed into gameplay when match starts. This is what this PR achieves. The actual implementation was fairly simple, but i had to handle a lot of changes in tests. I also wrote a new one in order to ensure that `MultiplayerHubContext.ChangeAndBroadcastUserState` doesn't add user to or remove them from gameplay group multiple times which would cause a memory leak & trigger certain events multiple times.

- [x] Depends on https://github.com/ppy/osu/pull/24071
- [x] Depends on `osu.Game` package bump

### DIFF
--- a/osu.Server.Spectator/Hubs/MultiplayerHubContext.cs
+++ b/osu.Server.Spectator/Hubs/MultiplayerHubContext.cs
@@ -128,25 +128,26 @@ namespace osu.Server.Spectator.Hubs
         {
             log(room, user, $"User state changed from {user.State} to {state}");
 
-            user.State = state;
-
             string? connectionId = users.GetConnectionIdForUser(user.UserID);
 
 			if (connectionId != null)
 			{
-                switch (state)
+                if (state == MultiplayerUserState.FinishedPlay)
                 {
-                    case MultiplayerUserState.FinishedPlay:
-                        await context.Groups.RemoveFromGroupAsync(connectionId, MultiplayerHub.GetGroupId(room.RoomID, true));
-                        break;
-
-                    case MultiplayerUserState.Ready:
-                    case MultiplayerUserState.Spectating:
-                    case MultiplayerUserState.Idle:
+                    // when entering FinishedPlay state, we can be confident that player can be removed from group
+                    await context.Groups.RemoveFromGroupAsync(connectionId, MultiplayerHub.GetGroupId(room.RoomID, true));
+                }
+                else
+                {
+                    // when leaving Results state, we can be confident that player can be removed from group
+                    if (user.State == MultiplayerUserState.Results)
+                    {
                         await context.Groups.AddToGroupAsync(connectionId, MultiplayerHub.GetGroupId(room.RoomID, true));
-                        break;
+                    }
                 }
 			}
+
+            user.State = state;
 
 			await context.Clients.Group(MultiplayerHub.GetGroupId(room.RoomID)).SendAsync(nameof(IMultiplayerClient.UserStateChanged), user.UserID, user.State);
         }
--- a/osu.Server.Spectator/Hubs/MultiplayerHubContext.cs
+++ b/osu.Server.Spectator/Hubs/MultiplayerHubContext.cs
@@ -132,8 +132,8 @@ namespace osu.Server.Spectator.Hubs
 
             string? connectionId = users.GetConnectionIdForUser(user.UserID);
 
-            if (connectionId != null)
-            {
+			if (connectionId != null)
+			{
                 switch (state)
                 {
                     case MultiplayerUserState.FinishedPlay:
@@ -146,9 +146,9 @@ namespace osu.Server.Spectator.Hubs
                         await context.Groups.AddToGroupAsync(connectionId, MultiplayerHub.GetGroupId(room.RoomID, true));
                         break;
                 }
-            }
+			}
 
-            await context.Clients.Group(MultiplayerHub.GetGroupId(room.RoomID)).SendAsync(nameof(IMultiplayerClient.UserStateChanged), user.UserID, user.State);
+			await context.Clients.Group(MultiplayerHub.GetGroupId(room.RoomID)).SendAsync(nameof(IMultiplayerClient.UserStateChanged), user.UserID, user.State);
         }
 
         public async Task ChangeRoomState(ServerMultiplayerRoom room, MultiplayerRoomState newState)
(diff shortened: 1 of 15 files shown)


osu.Server.Spectator/Hubs/MultiplayerHubContext.cs
```diff
@@ -128,26 +128,27 @@ public async Task ChangeAndBroadcastUserState(ServerMultiplayerRoom room, Multip
...
             log(room, user, $"User state changed from {user.State} to {state}");
 
-            user.State = state;
-
             string? connectionId = users.GetConnectionIdForUser(user.UserID);
 
             if (connectionId != null)
             {
-                switch (state)
+                if (state == MultiplayerUserState.FinishedPlay)
```
peppy, line 135: Please change this back to a switch statement.

osu.Server.Spectator/Hubs/MultiplayerHubContext.cs
```diff
@@ -128,26 +128,27 @@ public async Task ChangeAndBroadcastUserState(ServerMultiplayerRoom room, Multip
...
-                        break;
-
-                    case MultiplayerUserState.Ready:
-                    case MultiplayerUserState.Spectating:
+                    // when entering FinishedPlay state, we can be confident that player can be removed from group
+                    await context.Groups.RemoveFromGroupAsync(connectionId, MultiplayerHub.GetGroupId(room.RoomID, true));
+                }
+                else
+                {
+                    // when leaving Results state, we can be confident that player can be removed from group
```
peppy, line 142: Huh?

osu.Server.Spectator/Hubs/MultiplayerHubContext.cs
```diff
@@ -128,26 +128,27 @@ public async Task ChangeAndBroadcastUserState(ServerMultiplayerRoom room, Multip
...
+                if (state == MultiplayerUserState.FinishedPlay)
                 {
-                    case MultiplayerUserState.FinishedPlay:
-                    case MultiplayerUserState.Idle:
-                        await context.Groups.RemoveFromGroupAsync(connectionId, MultiplayerHub.GetGroupId(room.RoomID, true));
-                        break;
-
-                    case MultiplayerUserState.Ready:
-                    case MultiplayerUserState.Spectating:
+                    // when entering FinishedPlay state, we can be confident that player can be removed from group
```
peppy, line 137: I don't understand this comment.

Why can `FinishedPlay` be removed but not `Idle`, like in the old code? I guess this is a technical limitation from osu!'s side regarding transition from the results screen, which while if so, is probably fine for an initial implementation, the comment here just does nothing to explain anything.

osu.Server.Spectator/Hubs/MultiplayerHubContext.cs
```diff
@@ -128,26 +128,27 @@ public async Task ChangeAndBroadcastUserState(ServerMultiplayerRoom room, Multip
...
+                    // when leaving Results state, we can be confident that player can be removed from group
+                    if (user.State == MultiplayerUserState.Results)
+                    {
                         await context.Groups.AddToGroupAsync(connectionId, MultiplayerHub.GetGroupId(room.RoomID, true));
-                        break;
+                    }
                 }
             }
 
+            user.State = state;
```
peppy, line 150: If you *need* to use the previous state, make a variable called `previousState`. Then put this back where it was.

osu.Server.Spectator/Hubs/MultiplayerHubContext.cs
```diff
@@ -166,15 +167,15 @@ public async Task StartMatch(ServerMultiplayerRoom room)
             if (room.Queue.CurrentItem.Expired)
                 throw new InvalidStateException("Cannot start an expired playlist item.");
 
-            var readyUsers = room.Users.Where(u => u.State == MultiplayerUserState.Ready).ToArray();
-
             // If no users are ready, skip the current item in the queue.
-            if (readyUsers.Length == 0)
```
peppy, line 172: What's the reasoning behind this?

peppy (changes requested):

osu.Server.Spectator/Hubs/MultiplayerHubContext.cs
timiimit: This condition is (at the moment) only relevant when countdown is used (if match is started by host, host will be ready).

The reason for the change is that we want `readyUsers` to contain both `Ready` and `Idle` players because gameplay will start for both. But we only want to consider `Ready` players when deciding whether gameplay will start or not. When everyone is idle and countdown reaches 0, it seems foolish to actually start gameplay. In my opinion at least one person should be ready.

osu.Server.Spectator/Hubs/MultiplayerHubContext.cs
```diff
@@ -166,15 +167,15 @@ public async Task StartMatch(ServerMultiplayerRoom room)
...
+            if (!room.Users.Any(u => u.State == MultiplayerUserState.Ready))
             {
                 await room.Queue.FinishCurrentItem();
```
timiimit, line 173:
```suggestion
```
Also, according to `MultiplayerMatchStartCountdownTest.AutoStartCountdownContinuesWhileAllUsersNotReady` test we want to skip current item if no one is ready.

I think it would make more sense to keep item in the queue after countdown because no one was ready for the item to begin. After countdown, host can either restart countdown or remove the item manually if it is not desired. Perhaps this change would be better suited for a separate PR?

osu.Server.Spectator/Hubs/MultiplayerHubContext.cs
timiimit: `Idle` player needs to be in gameplay group so that he will join gameplay when next match starts. The only time when player is not in a state fit for gameplay group is in `FinishedPlay` or `Results` states.

I will remove the comment.

osu.Server.Spectator/Hubs/MultiplayerHubContext.cs
timiimit: using the previous state was the only way I could find to keep `EnsureThereIsNoGameplayGroupMemoryLeak` test passing. `Spectating` state throws everything off.

osu.Server.Spectator/Hubs/MultiplayerHubContext.cs
timiimit: Does it make sense to have a switch with only 1 case? There needs to be one check for previous state and one for new state.

osu.Server.Spectator/Hubs/MultiplayerHubContext.cs
timiimit: Oops, comment meant to say "can be added to group". Nevertheless, I will remove it.

osu.Server.Spectator/Hubs/MultiplayerHubContext.cs
peppy: I believe this is incorrect. I don't understand what you mean. The progression between states is sequential.

peppy:
I've pushed simplifications to this PR.

I have NOT yet tested this, and believe there may be a very big oversight: What happens if a user doesn't have the beatmap available locally? This needs test coverage at very least.

peppy:
Fails precisely as expected:

osu! 2023-06-27 at 10 34 16

peppy:
I've fixed this up and tested quite extensively.

Unfortunately there are now a lot of line changes. But most are fixing tests.

I'm not going to attempt to split out the beatmap availability changes into their own PR as they are very intertwined with the behavioural changes in this PR.

peppy (approved):

bdach:
I've piled on some changes of my own. They're mostly menial cleanup in tests, but do have a second look.

Other than that this looks good. I haven't tested that much, but in the testing I've done, I can't seem to break this.

bdach (approved):

PR #172, opened 2023-05-21
Fix build warnings

timiimit:
Not sure if your build process returns warnings, but I always get these two.
image

This PR fixes both.

### DIFF
--- a/osu.Server.Spectator/StartupDevelopment.cs
+++ b/osu.Server.Spectator/StartupDevelopment.cs
@@ -45,7 +45,8 @@ namespace osu.Server.Spectator
         /// </summary>
         protected override Task<AuthenticateResult> HandleAuthenticateAsync()
         {
-            if (!Context.Request.Headers.TryGetValue("user_id", out var userIdString))
+            string? userIdString;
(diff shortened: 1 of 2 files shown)


osu.Server.Spectator/StartupDevelopment.cs
```diff
@@ -45,7 +45,8 @@ public class LocalAuthenticationHandler : AuthenticationHandler<AuthenticationSc
         /// </summary>
         protected override Task<AuthenticateResult> HandleAuthenticateAsync()
         {
-            if (!Context.Request.Headers.TryGetValue("user_id", out var userIdString))
+            string? userIdString;
+            if (!Context.Request.Headers.TryGetValue("user_id", out var userIdValues) || (userIdString = userIdValues) == null)
```
bdach, line 49: This compound-assignment-and-nullcheck is dreadful readability-wise. If you ask me this should read

```csharp
string? userIdString = null;

if (Context.Request.Headers.TryGetValue("user_id", out var userIdValue))
    userIdString = userIdValue;

userIdString ??= Interlocked.Increment(ref userIDCounter).ToString();
```

timiimit: I agree, way better.

bdach (approved):

smoogipoo (approved):

PR #176, opened 2023-06-25
Fix issues with score upload logic

peppy:
-  Rewrite `ScoreUploader` timing logic to avoid multiple unexpected concurrent flush runs
- Don't specify ACL when uploading to S3

Would probably be nice to have test logic covering uploads arriving during shutdown (aka disposal) but a bit hard to make competent tests timing-wise.

### DIFF
--- a/osu.Server.Spectator/Hubs/ScoreUploader.cs
+++ b/osu.Server.Spectator/Hubs/ScoreUploader.cs
@@ -39,7 +39,7 @@ namespace osu.Server.Spectator.Hubs
             cancellationSource = new CancellationTokenSource();
             cancellationToken = cancellationSource.Token;
 
-            Task.Run(runFlushLoop);
+            Task.Factory.StartNew(runFlushLoop, TaskCreationOptions.LongRunning);
         }
 
         private void runFlushLoop()
--- a/osu.Server.Spectator/Hubs/ScoreUploader.cs
+++ b/osu.Server.Spectator/Hubs/ScoreUploader.cs
@@ -9,7 +9,6 @@ using osu.Game.Scoring;
 using osu.Server.Spectator.Database;
 using osu.Server.Spectator.Entities;
 using osu.Server.Spectator.Storage;
-using Timer = System.Timers.Timer;
 
 namespace osu.Server.Spectator.Hubs
 {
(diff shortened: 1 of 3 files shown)


osu.Server.Spectator/Hubs/ScoreUploader.cs
```diff
@@ -33,22 +28,29 @@ public double UploadInterval
...
-            timer = new Timer(50);
-            timer.AutoReset = false;
-            timer.Elapsed += (_, _) => Task.Run(Flush);
-            timer.Start();
+            Task.Run(runFlushLoop);
+        }
+
+        private void runFlushLoop()
+        {
+            while (!queue.IsEmpty || !cancellationToken.IsCancellationRequested)
```
bdach, line 47: Doesn't checking for empty here mean that if the queue ever becomes empty then the uploader will stop flushing forevermore?

peppy: I'm open to rewriting the conditional to be less confusing, but the logic currently there does check out as far as i can tell.

bdach: Oh wait, it's an `||` not an `&&`. Misread.

Misled me because the `queue.IsEmpty` is basically a dead condition _until_ cancellation happens.

peppy:
Not sure about the test failure. I ran this locally 20+ times without any failure, will need to revisit.

bdach:
No bright ideas here either I'm afraid. One thing that is slightly suspect is that the score uploader is never disposed in the tests. But trying to force threadpool saturation via `SetMaxThreads(1, 1)` got me precisely anywhere.

osu.Server.Spectator/Hubs/ScoreUploader.cs
```diff
@@ -33,22 +28,29 @@ public double UploadInterval
...
-            timerCancellationSource = new CancellationTokenSource();
-            timerCancellationToken = timerCancellationSource.Token;
+            cancellationSource = new CancellationTokenSource();
+            cancellationToken = cancellationSource.Token;
 
-            timer = new Timer(50);
-            timer.AutoReset = false;
-            timer.Elapsed += (_, _) => Task.Run(Flush);
-            timer.Start();
+            Task.Run(runFlushLoop);
```
smoogipoo, line 42: Am I reading this correct, that this is infinite-looping a TPL thread? This can't be good. At the very least this should be a long running thread.

smoogipoo (changes requested):

osu.Server.Spectator/Hubs/ScoreUploader.cs
peppy: Oh yep, for sure.

smoogipoo (approved):
Looks okay to me

PR #178, opened 2023-07-16
Update game packages

peppy:


bdach:
I mean sure, but does this mean we're not doing https://github.com/ppy/osu-server-spectator/pull/170 this release? I would have expected this bump to happen there, normally.

bdach (approved):

peppy:
oops

peppy:
we can work with this though.

PR #185, opened 2023-09-01
Fix disparate score token types being used as one type

bdach:
Server-side counterpart to https://github.com/ppy/osu/pull/24697. See description of said pull for rationale.

- [x] Depends on https://github.com/ppy/osu-web/pull/10437 (merge & deployment) for correct operation
- [x] Depends on https://github.com/ppy/osu/pull/24697 (merge & nuget package) to compile


peppy:
Closing for now to keep things clean. See https://github.com/ppy/osu-infrastructure/issues/24 for rationale.

PR #188, opened 2023-10-03
Add ability to invite players to multiplayer rooms

minetoblend:
Companion to ppy/osu#25005.


osu.Server.Spectator/Hubs/Multiplayer/MultiplayerHub.cs
```diff
@@ -233,6 +233,46 @@ public async Task LeaveRoom()
...
+                var inverseRelation = await db.GetUserRelation(userId, CurrentContextUserId);
+
+                if (inverseRelation == null)
+                {
+                    bool userAllowsPMs = await db.GetUserAllowsPMs(userId);
+                    if (!userAllowsPMs)
+                        throw new UserBlocksPMsException();
+                }
+                else if (inverseRelation.foe)
+                    throw new UserBlockedException();
```
peppy, line 257: Not sure about the same exception being used for blocking in both directions. Maybe okay?

peppy (commented):
Looks pretty good overall 👍

peppy (approved):

bdach (approved):

Issue #187, opened 2023-10-01
Non-passing scores should not upload replays


peppy:
I don't think we're tracking this anywhere, and it's quite important. Likely a simple change.

smoogipoo:
Added to worklist item https://github.com/ppy/osu-infrastructure/issues/19, removing this issue from project (mentioned issue is in project).

PR #190, opened 2023-10-10
Don't upload replays for failed scores

smoogipoo:
Resolves https://github.com/ppy/osu-server-spectator/issues/187
Closes https://github.com/ppy/osu-infrastructure/issues/19

`SoloScore` model taken from [osu-queue-score-statistics](https://github.com/ppy/osu-queue-score-statistics/blob/3cc58af62a4805ca51c1ef13ed85385b4179cf84/osu.Server.Queues.ScoreStatisticsProcessor/Models/SoloScore.cs)

### DIFF
--- a/osu.Server.Spectator/Database/DatabaseAccess.cs
+++ b/osu.Server.Spectator/Database/DatabaseAccess.cs
@@ -295,14 +295,15 @@ namespace osu.Server.Spectator.Database
             });
         }
 
-        public async Task<long?> GetScoreIdFromToken(long token)
+        public async Task<SoloScore?> GetScoreFromToken(long token)
         {
             var connection = await getConnectionAsync();
 
-            return await connection.QuerySingleOrDefaultAsync<long?>("SELECT `score_id` FROM `solo_score_tokens` WHERE `id` = @Id", new
-            {
-                Id = token
-            });
+            return await connection.QuerySingleOrDefaultAsync<SoloScore?>(
+                "SELECT * FROM `solo_scores` WHERE `id` = (SELECT `score_id` FROM `solo_score_tokens` WHERE `id` = @Id)", new
+                {
+                    Id = token
+                });
         }
 
         public async Task<bool> IsScoreProcessedAsync(long scoreId)
--- a/osu.Server.Spectator/Hubs/ScoreUploader.cs
+++ b/osu.Server.Spectator/Hubs/ScoreUploader.cs
@@ -7,6 +7,7 @@ using System.Threading;
 using System.Threading.Tasks;
 using osu.Game.Scoring;
 using osu.Server.Spectator.Database;
+using osu.Server.Spectator.Database.Models;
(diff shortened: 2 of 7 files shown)


osu.Server.Spectator/Database/DatabaseAccess.cs
```diff
@@ -295,14 +295,15 @@ public async Task MarkScoreHasReplay(Score score)
...
 
-            return await connection.QuerySingleOrDefaultAsync<long?>("SELECT `score_id` FROM `solo_score_tokens` WHERE `id` = @Id", new
-            {
-                Id = token
-            });
+            return await connection.QuerySingleOrDefaultAsync<SoloScore?>(
+                "SELECT * FROM `solo_scores` WHERE `id` = (SELECT `score_id` FROM `solo_score_tokens` WHERE `id` = @Id)", new
+                {
+                    Id = token
+                });
```
bdach, line 306: I'll just note that this is changing the same method that my series of pulls concerning replay treatment is changing (see https://github.com/ppy/osu-server-spectator/pull/185).

This query is still - same as `master` - correct only for solo scores and wrong for multiplayer scores (see https://github.com/ppy/osu/pull/24697 / https://github.com/ppy/osu-infrastructure/blob/master/score-submission.md#multiplayer for full explanation), but I suppose it is no *more* wrong than `master`.

Make of that what you will wrt which pull should go first etc.

osu.Server.Spectator/Hubs/ScoreUploader.cs
```diff
@@ -101,14 +102,26 @@ public async Task Flush()
...
+                                return;
+                            }
+
+                            if (!dbScore.ScoreInfo.Passed)
+                                return;
+
+                            var updatedScore = new Score
+                            {
+                                Replay = item.Score.Replay,
+                                ScoreInfo = dbScore.ScoreInfo.ToScoreInfo(item.Score.ScoreInfo.Mods, item.Score.ScoreInfo.BeatmapInfo)
```
bdach, line 117: Is there a reason why this is preferring the db-retrieved `ScoreInfo` over the already-existing local one? Seems generally arbitrary.

smoogipoo: The alternative is to manually copy over the `Passed` value from the databased model. Does that sit better with you?

For me, if we have to do that for `Passed` then it seems as something we would have to do over again in the future for other properties, and I feel uneasy about that.

osu.Server.Spectator/Database/DatabaseAccess.cs
smoogipoo: I'm okay waiting, I don't see this as a higher priority than the other PR.

bdach: disregard above, see https://github.com/ppy/osu-infrastructure/issues/24

peppy (approved):

PR #197, opened 2023-11-22
Update users' last visit time on connecting to spectator server

bdach:
Addresses the core part of ppy/osu#12635.

The functionality is purposefully placed in `MetadataHub` due to an IRL discussion with @peppy (and also briefly @nanaya), whose conclusion is that we probably don't want a new hub for this to avoid introducing even more websocket connections into the mix.

The updates are performed in batches of 100. The updates in a batch are staggered, and there is a delay between batches as well. The delay is partially arbitrary, but matches web [in a sense](https://github.com/ppy/osu-web/blob/f858ec2d1c2c61f4229c40bc01a943e9df2376e0/app/Http/Middleware/UpdateUserLastvisit.php#L37).

I intend to extend the metadata hub further with reporting the users' activities and broadcasting them to other clients, as well as tracking per-build information.

This change does not depend on https://github.com/ppy/osu-server-spectator/pull/196, but the two will inevitably conflict. I will deal with that when it will be required to do so.

### DIFF
--- a/osu.Server.Spectator.Tests/BatchMetadataUpdaterTests.cs
+++ b/osu.Server.Spectator.Tests/BatchMetadataUpdaterTests.cs
@@ -1,0 +1,73 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+using Moq;
+using osu.Server.Spectator.Database;
+using osu.Server.Spectator.Entities;
+using osu.Server.Spectator.Hubs.Metadata;
+using Xunit;
+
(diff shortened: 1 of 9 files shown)


nanaya:
note that in osu-web the process also updates elasticsearch for the online user search.

bdach:
Well that is going to be a bit of a problem isn't it...

I guess this leaves two options: either do it in web instead, or have spectator server touch elastic, which it has never done before? Dunno...

peppy:
> note that in osu-web the process also updates elasticsearch for the online user search.

I guess we may need a LIO endpoint for this? The other path we can take is having the last activity updated based on incoming API requests from the lazer client.. @nanaya thoughts?

nanaya:
I think updating on API requests is easier. It may get out of date if the user doesn't hit api at all for a long time but it should be rare?

bdach:
Closing, pending osu-web implementation.

PR #201, opened 2023-12-01
Add support for aborting the multiplayer match

smoogipoo:
Prereqs:
- [x] https://github.com/ppy/osu/pull/25637
- [x] nuget w/ above

### DIFF
--- a/SampleMultiplayerClient/MultiplayerClient.cs
+++ b/SampleMultiplayerClient/MultiplayerClient.cs
@@ -35,7 +35,7 @@ namespace SampleMultiplayerClient
             connection.On<int, MultiplayerUserState>(nameof(IMultiplayerClient.UserStateChanged), ((IMultiplayerClient)this).UserStateChanged);
             connection.On<int, BeatmapAvailability>(nameof(IMultiplayerClient.UserBeatmapAvailabilityChanged), ((IMultiplayerClient)this).UserBeatmapAvailabilityChanged);
             connection.On(nameof(IMultiplayerClient.LoadRequested), ((IMultiplayerClient)this).LoadRequested);
-            connection.On(nameof(IMultiplayerClient.LoadAborted), ((IMultiplayerClient)this).LoadAborted);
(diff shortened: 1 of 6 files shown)


SampleMultiplayerClient/MultiplayerClient.cs
```diff
@@ -217,9 +220,9 @@ Task IMultiplayerClient.LoadRequested()
             return Task.CompletedTask;
         }
 
-        Task IMultiplayerClient.LoadAborted()
+        Task IMultiplayerClient.GameplayAborted(GameplayAbortReason reason)
         {
-            Console.WriteLine($"User {UserID} gameplay load was aborted");
+            Console.WriteLine($"User {UserID} gameplay load was aborted ({reason})");
```
bdach, line 225:
```suggestion
            Console.WriteLine($"User {UserID} gameplay was aborted ({reason})");
```

osu.Server.Spectator/Hubs/Multiplayer/MultiplayerHub.cs
```diff
@@ -507,6 +507,27 @@ public async Task StartMatch()
...
+                var room = roomUsage.Item;
+                if (room == null)
+                    throw new InvalidOperationException("Attempted to operate on a null room");
+
+                ensureIsHost(room);
+
+                foreach (var user in room.Users.Where(u => isGameplayState(u.State)))
+                {
+                    await HubContext.ChangeAndBroadcastUserState(room, user, MultiplayerUserState.Idle);
+                    await Clients.User(user.UserID.ToString()).GameplayAborted(GameplayAbortReason.HostAbortedTheMatch);
```
bdach, line 524: I'm not sure how much it matters but rather than messaging each individual client you could also do `Clients.Group(GetGroupId(room.RoomId)).GameplayAborted(...)`?

peppy: I think sending to all users is probably better (so we can show it as an event in the room lobby for those that are refereeing or whatever).

Issue #138, opened 2022-06-18
Add tracking and reporting of users-per-build


peppy:
As brought up in https://github.com/ppy/osu/discussions/18590#discussioncomment-2932477, we need to report counts of users per release so it can be shown in places like the [changelog header](https://osu.ppy.sh/home/changelog).

Here's bancho's implementation of this for reference. The important part is the `UPDATE` statement (the `INSERT` one is not relevant for the newer implemetation) but I've included the whole stats method for context.

```csharp
        private static bool isUpdatingStats;

        private static void updateStats()
        {
            if (isUpdatingStats || !ReportClientCounts) return;

            var countOsu = UserManager.CountOsu;

            // avoid reporting low stats by accident
            if (countOsu < 32) return;

            isUpdatingStats = true;

            RunThread(() =>
            {
                try
                {
                    Database.RunNonQueryLowPriority(
                        "INSERT INTO osu_banchostats (users_irc, users_osu, multiplayer_games) VALUES (@irc, @osu, @mp)",
                        new MySqlParameter("irc", UserManager.CountIrc),
                        new MySqlParameter("osu", countOsu),
                        new MySqlParameter("mp", Lobby.Matches.Count));

                    SortedDictionary<string, int> versionCounts = new SortedDictionary<string, int>();

                    foreach (Client c in UserManager.Clients)
                    {
                        NetClient nc = c as NetClient;
                        if (nc == null) continue;

                        string version = Regex.Replace(nc.Version, "[^0-9.]", string.Empty);

                        if (versionCounts.ContainsKey(version))
                            versionCounts[version]++;
                        else
                            versionCounts[version] = 1;
                    }

                    foreach (var v in versionCounts)
                    {
                        Database.RunNonQueryLowPriority("UPDATE osu_builds SET users = @users WHERE version = @version",
                            new MySqlParameter("users", v.Value),
                            new MySqlParameter("version", v.Key)
                        );
                    }
                }
                catch (Exception e)
                {
                    Bacon.WriteLine("Database connection failed!");
                    Bacon.WriteLine(e.ToString());
                }

                isUpdatingStats = false;
            });
        }
```

bdach:
Implementation needs more tweaks (see https://discord.com/channels/188630481301012481/188630652340404224/1184782370105200753)

PR #203, opened 2023-12-08
Track user counts per build

bdach:
- [x] Depends on https://github.com/ppy/osu-server-spectator/pull/202
- Closes https://github.com/ppy/osu-server-spectator/issues/138

No osu!-side changes necessary for this. [The client already sends its version hash to the spectator server on connecting, in headers](https://github.com/ppy/osu/blob/07da9d95a90c330b4bf90e26fb68ee24a2e3e803/osu.Game/Online/HubClientConnector.cs#L72) - the server just hasn't been doing anything with this data thus far.

The tracking is disabled by default, and must be enabled by setting the envvar `TRACK_BUILD_USER_COUNTS` to `1`. I imagine we may want to keep this disabled on dev server just to have less noise there as it's pretty much useless over there.

The user counts are updated every 5 minutes by default.

### DIFF
--- a/osu.Server.Spectator/Database/DatabaseAccess.cs
+++ b/osu.Server.Spectator/Database/DatabaseAccess.cs
@@ -2,6 +2,7 @@
 // See the LICENCE file in the repository root for full licence text.
 
 using System;
+using System.Collections.Generic;
 using System.Diagnostics;
 using System.IdentityModel.Tokens.Jwt;
 using System.Linq;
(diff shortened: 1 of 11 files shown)


osu.Server.Spectator/Database/DatabaseAccess.cs
```diff
@@ -335,6 +336,24 @@ public async Task<bool> GetUserAllowsPMs(int userId)
...
+        public async Task<IEnumerable<osu_build>> GetAllLazerBuildsAsync()
+        {
+            var connection = await getConnectionAsync();
+
+            return await connection.QueryAsync<osu_build>(
+                "SELECT `build_id`, `version`, `hash`, `users` "
+                + "FROM `osu_builds` `b` "
+                + "JOIN `osu_updates`.`streams` `s` ON `b`.`stream_id` = `s`.`stream_id` "
+                + "WHERE `s`.`name` = 'lazer'");
+        }
```
bdach, line 348: If there's anything that needs scrutinising in this diff, it is this query. I am  _reasonably_ convinced that it will work in production, but it would be wise to check, especially that we have [weird carve-outs for android still](https://github.com/ppy/osu/blob/07da9d95a90c330b4bf90e26fb68ee24a2e3e803/osu.Game/OsuGameBase.cs#L256-L266) and I'm not sure how that works in production.

Generally the reason this is a thing is to decrease blast radius, first of all, as I generally wouldn't want to be spraying updates on production based from a client-supplied version hash without any checking. However it has a correctness purpose as well - see the implementation of `BuildUserCountUpdater`. If the thing was to only rely on the user counts as reported by clients to the server, it would stop being aware of a build once the last client that was connected to it on that build disconnects.

peppy: Modified slightly, seems fine otherwise.

Obviously won't hold out if we horizontally scale, but we'll deal with that later.

peppy (approved):

osu.Server.Spectator/Database/DatabaseAccess.cs
bdach: Yeah, I have done 0 consideration for scaling here given that the rest of the server doesn't scale.

PR #210, opened 2024-01-29
Upgrade to .NET 8

peppy:
The majority of the overhead we're seeing is system%, in sockets and native code. Hardly any of what we are doing is causing overhead. Given the amount of optimisation done between net6 and net8, this should give us some headroom.

Willing to take a gamble on this one under the premise that we can rollback if required. Will test on staging for efficiency.

https://github.com/dotnet/AspNetCore.Docs/issues/17880

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -16,10 +16,10 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v2
 
-      - name: Install .NET 6.0.x
+      - name: Install .NET 8.0.x
         uses: actions/setup-dotnet@v1
         with:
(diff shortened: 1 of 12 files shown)


bdach:
I had once [attempted this](https://github.com/ppy/osu-server-spectator/compare/master...bdach:osu-server-spectator:dotnet-7) with dotnet 7 and the notes more or less check out so sure.

I have not tested this because due to a whole mess of ubuntu reasons I needed to upgrade and upgrading got me net8 but broke net6 so now I can launch server spectator but cannot launch client. (Easily fixable but annoying, so in order to not delay this any further...)

bdach (approved):

PR #212, opened 2024-01-30
Fix incorrect client hash

smoogipoo:
The client will eventually send the correct hash as it does for osu-web, for this server too.

### DIFF
--- a/osu.Server.Spectator/Hubs/Metadata/MetadataHub.cs
+++ b/osu.Server.Spectator/Hubs/Metadata/MetadataHub.cs
@@ -38,7 +38,14 @@ namespace osu.Server.Spectator.Hubs.Metadata
                 string? versionHash = null;
 
                 if (Context.GetHttpContext()?.Request.Headers.TryGetValue("OsuVersionHash", out StringValues headerValue) == true)
+                {
                     versionHash = headerValue;
+
+                    // The token is 82 chars long, and the clientHash is the first 32 of those.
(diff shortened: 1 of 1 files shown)


osu.Server.Spectator/Hubs/Metadata/MetadataHub.cs
```diff
@@ -38,8 +38,15 @@ public override async Task OnConnectedAsync()
                 string? versionHash = null;
 
                 if (Context.GetHttpContext()?.Request.Headers.TryGetValue("OsuVersionHash", out StringValues headerValue) == true)
+                {
                     versionHash = headerValue;
 
+                    // The token is 82 chars long, and the clientHash is the first 32 of those.
+                    // See: https://github.com/ppy/osu-web/blob/7be19a0fe0c9fa2f686e4bb686dbc8e9bf7bcf84/app/Libraries/ClientCheck.php#L92
+                    if (versionHash?.Length >= 82)
+                        versionHash = versionHash.Substring(versionHash.Length - 82, 32);
```
bdach, line 47: What's with the subtraction? If the client hash is the first 32 chars of this, then why is the first argument not... just 0?

smoogipoo: The 82-length portion is actually at the end of a much longer string.

Actually now that I mention it, I should probably make sure this is only sent on the first request otherwise this is quite a lot of data to send in a realtime system.

smoogipoo: Should be fine so long as SignalR doesn't fallback to one of the polling transports. If it does we can fix that when we come to it.

peppy (approved):

PR #216, opened 2024-02-01
Change type of some exceptions so they're logged

smoogipoo:


smoogipoo:
🤔 I can't repro the test failure locally, and the exception is expected

bdach (approved):

Issue #220, opened 2024-02-22
Investigate high rate of replay upload timeouts


peppy:
![Image](https://github.com/ppy/osu-server-spectator/assets/20418176/2cfb985f-135f-4f16-9423-cc305fc912a7)

https://sentry.ppy.sh/organizations/ppy/issues/34904/

peppy:
@bdach to follow up on this, we figured that the number being reported on datadog was much higher than actual, but I forget the reason why, or how i confirmed that this was the case. or if we actually decided this was still an issue..

```sql
root@db-master.osu.io:osu> select has_replay, count(*) from scores where preserve = 1 and passed = 1 and legacy_score_id is null and id between 2386002880 and 238700288
                        -> 0 group by has_replay;
+------------+----------+
| has_replay | count(*) |
+------------+----------+
| 1          | 22345    |
| 0          | 3669     |
+------------+----------+
```

bdach:
That query is missing an `approved in (1, 2)`.

Running the query did look significantly better but we never got to explain why the numbers were being overestimated like that. I believe we did rule out aborted submission.

Next step would be to take a sample of those timeout messages from logs or from sentry and look for possible patterns in database. Or just blindly try increasing timeout I guess.

peppy:
> That query is missing an `approved in (1, 2)`.

aha, that's the one.

```sql
root@db-master.osu.io:osu> select has_replay, count(*) from scores join osu_beatmaps using (beatmap_id) where preserve = 1 and passed = 1 and legacy_score_id is null an
                        -> d id between 2386002880 and 2387002880 and approved in (1,2) group by has_replay;
+------------+----------+
| has_replay | count(*) |
+------------+----------+
| 1          | 21148    |
| 0          | 36       |
+------------+----------+
```

peppy:
Let's close this one for the time being. As we discovered, the failure rate is actually very low.

bdach:
I'd still want to investigate the metrics themselves and why they're overreporting the actual count. Maybe we can save on some unnecessary work in the server.

PR #224, opened 2024-03-14
Skip score uploads when nothing was hit in score

bdach:
RFC.

The client [will not submit a play if hit statistics indicate that they never hit anything.](https://github.com/ppy/osu/blob/a47ccb8edd2392258b6b7e176b222a9ecd511fc0/osu.Game/Screens/Play/SubmittingPlayer.cs#L281)

This check was not exercised spectator server-side, which meant that scores with nothing hit would get completed and enqueued for upload that would never succeed (because the score would never be submitted anyway). To save on some processing, just port the same check server-side to skip upload if nothing was hit.

I'm hoping the effects of this is going to be the decrease of the score timeouts monitoring on sentry & datadog closer to real levels (see discussion in https://github.com/ppy/osu-server-spectator/issues/220).

### DIFF
--- a/osu.Server.Spectator/Hubs/Spectator/SpectatorHub.cs
+++ b/osu.Server.Spectator/Hubs/Spectator/SpectatorHub.cs
@@ -8,6 +8,7 @@ using System.Threading.Tasks;
 using Microsoft.Extensions.Caching.Distributed;
 using osu.Game.Beatmaps;
 using osu.Game.Online.Spectator;
+using osu.Game.Rulesets.Scoring;
 using osu.Game.Scoring;
 using osu.Server.Spectator.Database;
 using osu.Server.Spectator.Database.Models;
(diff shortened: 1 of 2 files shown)


osu.Server.Spectator/Hubs/Spectator/SpectatorHub.cs
```diff
@@ -145,10 +146,16 @@ public async Task EndPlaySession(SpectatorState state)
                     if (status < min_beatmap_status_for_replays || status > max_beatmap_status_for_replays)
                         return;
 
-                    score.ScoreInfo.Date = DateTimeOffset.UtcNow;
+                    // if the user never hit anything, further processing that depends on the score existing can be waived because the client won't have submitted the score anyway.
+                    // note that this isn't an early return as we still want to end the play session.
+                    // see: https://github.com/ppy/osu/blob/a47ccb8edd2392258b6b7e176b222a9ecd511fc0/osu.Game/Screens/Play/SubmittingPlayer.cs#L281
```
smoogipoo, line 151: I don't quite understand what this comment means by "end the play session" - there's nothing running after the block other than the finally, right?

bdach: There is actually:

https://github.com/ppy/osu-server-spectator/blob/cc1d36da977b363d2ac21d93a3b9690f40c26146/osu.Server.Spectator/Hubs/Spectator/SpectatorHub.cs#L159

Skipping that causes a test to fail. I considered moving that call inside `finally` but wasn't convinced that was correct.

smoogipoo: Oh, it was hidden. How come that isn't an issue for the above `return`?

bdach: I can understand this one skipping it:

https://github.com/ppy/osu-server-spectator/blob/cc1d36da977b363d2ac21d93a3b9690f40c26146/osu.Server.Spectator/Hubs/Spectator/SpectatorHub.cs#L138-L141

because it assumes that `BeginPlaySession()` failed. But I don't understand this one:

https://github.com/ppy/osu-server-spectator/blob/cc1d36da977b363d2ac21d93a3b9690f40c26146/osu.Server.Spectator/Hubs/Spectator/SpectatorHub.cs#L143-L146

I think that may be an actual bug?

smoogipoo: Hmmm okay. That second one is the one I was referring to specifically. I think this is fine to get in as-is but that should probably be investigated.

smoogipoo (approved):

osu.Server.Spectator/Hubs/Spectator/SpectatorHub.cs
bdach: Yeah I'm looking at it.

PR #227, opened 2024-04-24
Apply chat filters to room name when editing multiplayer room settings

bdach:
Covers off another avenue of people potentially misusing text boxes after https://github.com/ppy/osu-web/pull/11175.

The chat filters are retrieved once-ever and never refetched until the instance gets restarted. I think web also does this? And also I didn't wanna complicate things unless necessary, but if deemed required I can add some crude local expiration/refetch logic.


peppy (approved):

peppy:
This looks good. Let's merge this in and look at the whitespace / block support separately.

PR #229, opened 2024-05-01
Add extended capabilities for chat filters

bdach:
Port of https://github.com/ppy/osu-web/pull/11181. Shares the same test cases and end behaviour.


osu.Server.Spectator/ChatFilters.cs
```diff
@@ -21,18 +28,57 @@ public ChatFilters(IDatabaseFactory factory)
 
         public async Task<string> FilterAsync(string input)
         {
-            if (filters == null)
+            if (!filtersInitialised)
+                await initialiseFilters();
+
+            if (blockRegex?.Match(input).Success == true)
+                throw new InvalidStateException(string.Empty);
```
bdach, line 35: I didn't see much benefit to providing an actual exception message here.

This is how this looks client-side:

osu_2024-05-01_09-26-15

peppy: How does it show if there is a message? Just as part of the exception description?

bdach: Yeah it just gets appended at the end.

peppy: Maybe better than nothing.. You just know users are going to report as a bug otherwise 😅

bdach: Yeah sure. Not sure if you want me to apply something or are going to do so yourself, given the review self-request below.

peppy: I'll grab it

peppy (approved):

PR #231, opened 2024-05-06
Ensure ranks are properly populated in spectated scores

bdach:
> [!WARNING]
> This - or an alternative PR fixing this same issue - **must** be merged before the next bump of game packages in spectator server. Without fixing this issue one way or another, **all** online replays after next game bump & deploy thereof will have rank D **stored to the replay and read client-side**.

RFC.

This begins to matter after https://github.com/ppy/osu/pull/28058, as with that pull the rank is encoded to the replay, so it must be correctly set on the server side. Otherwise spectator server will default to storing D rank, which the client will happily read and use raw without checking.

This is purposefully computed server-side rather than sent in replay headers for the reasons of simplicity and backwards compatibility. You could add score rank to the replay headers, but to ensure that replays recorded with old clients don't have D rank set, recomputing server-side would have to be done at least for the old clients. And you could argue recomputing ranks always server-side removes a tamper vector too (although that's not a very strong argument).

---

Also includes a sentry package bump in https://github.com/ppy/osu-server-spectator/commit/e1603205f57e0cd98058cb71b7f8b904d579488a - without this the server dies at runtime with newest game packages.

### DIFF
--- a/osu.Server.Spectator.Tests/SpectatorHubTest.cs
+++ b/osu.Server.Spectator.Tests/SpectatorHubTest.cs
@@ -2,6 +2,7 @@
 // See the LICENCE file in the repository root for full licence text.
 
 using System;
+using System.Collections.Generic;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.SignalR;
 using Microsoft.Extensions.Caching.Distributed;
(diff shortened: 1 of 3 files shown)


peppy:
@bdach i'm quite confident it is, but just to confirm, this is fine to deploy immediately yes?

peppy (approved):

bdach:
As far as I'm aware yes, this change is basically inert until the next game package bump.

peppy:
Inert assuming no funny business, sure.

I ask because I was planning a new spectator deploy today, will include this 👍

PR #237, opened 2024-06-27
Allow clients to receive realtime updates for a given playlist

bdach:
- Continuation of https://github.com/ppy/osu/issues/28136
- [x] Depends on https://github.com/ppy/osu/pull/28636
- [x] Depends on game package w/ above

This pull is intended to be the thing powering recently-added game-side components like https://github.com/ppy/osu/pull/28561 or https://github.com/ppy/osu/pull/28468.

The general design is not intended to explicitly be specific to daily challenge; the API of the feature is supposed to - in theory - apply to any playlist, so if we wanted to we could add these displays to any old playlist room. The one exception is multiplayer; I kinda forget the reason since I wrote the code for this a *while* back, but I believe it had something to do with the fact that in realtime multi the playlist items are not fixed in time. Now I'm not sure it's an issue anymore but to be honest I don't think realtime rooms are ever going to need this sort of feedback either, so I'm leaving that limitation be there.

Few open questions in my mind remain, which is why I'm PRing this early and not PRing the final piece that integrates this into the game yet

- Not sure of the general placement of this (considered metadata hub and spectator hub, decided on the former in the end)
- The general data structure design is a bit ad-hoc to easily support what I want to show client-side and may need adjusting / generalising
- Not sure how the db queries will perform on production scale, things may need to be shifted depending on that

so before all of that is set in stone I'd wanna refrain from getting too deep into it so that I don't have to change things five times over.

### DIFF
--- a/osu.Server.Spectator/Database/DatabaseAccess.cs
+++ b/osu.Server.Spectator/Database/DatabaseAccess.cs
@@ -48,6 +48,16 @@ namespace osu.Server.Spectator.Database
             }) != 0;
         }
 
+        public async Task<multiplayer_room?> GetRoomAsync(long roomId)
+        {
+            var connection = await getConnectionAsync();
+
+            return await connection.QueryFirstOrDefaultAsync<multiplayer_room>("SELECT * FROM multiplayer_rooms WHERE id = @RoomID", new
+            {
+                RoomID = roomId
+            });
+        }
+
         public async Task<multiplayer_room?> GetRealtimeRoomAsync(long roomId)
         {
             var connection = await getConnectionAsync();
@@ -294,7 +304,7 @@ namespace osu.Server.Spectator.Database
             });
         }
 
-        public async Task<SoloScore?> GetScoreFromToken(long token)
+        public async Task<SoloScore?> GetScoreFromTokenAsync(long token)
         {
             var connection = await getConnectionAsync();
 
@@ -303,6 +313,16 @@ namespace osu.Server.Spectator.Database
                 {
                     Id = token
                 });
+        }
+
+        public async Task<SoloScore?> GetScoreAsync(long id)
+        {
+            var connection = await getConnectionAsync();
+
+            return await connection.QuerySingleOrDefaultAsync<SoloScore?>("SELECT * FROM `scores` WHERE `id` = @Id", new
+            {
+                Id = id
+            });
         }
 
         public async Task<bool> IsScoreProcessedAsync(long scoreId)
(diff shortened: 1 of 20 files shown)


osu.Server.Spectator/Database/DatabaseAccess.cs
```diff
@@ -382,6 +402,78 @@ public async Task<IEnumerable<multiplayer_room>> GetActiveDailyChallengeRoomsAsy
...
+        public async Task<MultiplayerPlaylistItemStats[]> GetMultiplayerRoomStatsAsync(long roomId)
+        {
+            var connection = await getConnectionAsync();
+
+            long[] playlistItemIds = (await GetAllPlaylistItemsAsync(roomId)).Select(item => item.id).ToArray();
+            var result = new MultiplayerPlaylistItemStats[playlistItemIds.Length];
+
+            for (int i = 0; i < playlistItemIds.Length; ++i)
+            {
+                long[] totalScores = (await connection.QueryAsync<long>(
```
peppy, line 427: I'm not sure how well this will scale, but I think we can go life with it and deal with that later.

Non-representative tests as we'll likely see a lot more usage than most existing playlists:

```sql
select playlist_item_id, count(score_id) from multiplayer_score_links group by playlist_item_id order by
                 -> count(score_id) desc limit 10;
+------------------+-----------------+
| playlist_item_id | count(score_id) |
+------------------+-----------------+
| 5415805          | 1090            |
| 7091206          | 353             |
| 5415807          | 321             |
| 5827890          | 291             |
| 4350065          | 280             |
| 4296355          | 279             |
| 5415798          | 253             |
| 5344648          | 250             |
| 7091692          | 249             |
| 4296362          | 242             |
+------------------+-----------------+

-- cold query (includes japan-us roundtrip)
select sum(total_score) from multiplayer_score_links join scores on (score_id = scores.id) where playlist_item_id = 5415805;
+------------------+
| sum(total_score) |
+------------------+
| 32986707         |
+------------------+

1 row in set
Time: 0.504s

-- warm query (includes japan-us roundtrip)
select sum(total_score) from multiplayer_score_links join scores on (score_id = scores.id) where playlist_item_id = 5415805;
+------------------+
| sum(total_score) |
+------------------+
| 32986707         |
+------------------+

1 row in set
Time: 0.130s
```

It shouldn't be a huge reach to cache these values if we need to, as well.

bdach: If it's any solace the intention is that this will only be requested whenever someone enters the daily challenge screen.

Caching can 100% be done even now. It's non-critical data, nobody is going to care about perfect accuracy of this.

peppy: Yep, aware of that. I'm also against caching until we need to.

peppy (approved):

PR #238, opened 2024-07-17
Read & use client-provided session GUID for concurrency control

bdach:
- [x] Depends on https://github.com/ppy/osu/pull/28892
- Second part of https://github.com/ppy/osu/issues/26338

Read description of aforementioned PR for explanation what is going on here.

This change is designed to be backwards-compatible, in the sense that old client will continue to use the `jti` cialm (with all of the deficiencies which that incurs), and new clients with the aforementioned change will use the new one which is hopefully rid of the deficiencies in question. We can probably shutter the old path in a few months - or I can just delete it now if that is considered acceptable.

### DIFF
--- a/osu.Server.Spectator/Entities/ConnectionState.cs
+++ b/osu.Server.Spectator/Entities/ConnectionState.cs
@@ -59,13 +59,6 @@ namespace osu.Server.Spectator.Entities
         public void RegisterConnectionId(HubLifetimeContext context)
             => ConnectionIds[context.Hub.GetType()] = context.Context.ConnectionId;
 
-        private bool tryGetClientSessionID(HubLifetimeContext context, out Guid clientSessionId)
-        {
-            clientSessionId = Guid.Empty;
-            return context.Context.GetHttpContext()?.Request.Headers.TryGetValue(HubClientConnector.CLIENT_SESSION_ID_HEADER, out var value) == true
-                   && Guid.TryParse(value, out clientSessionId);
-        }
-
         public bool IsConnectionFromSameClient(HubLifetimeContext context)
         {
             if (tryGetClientSessionID(context, out var clientSessionId))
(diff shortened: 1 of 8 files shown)


osu.Server.Spectator/Entities/ConnectionState.cs
```diff
@@ -42,5 +58,36 @@ public ConnectionState(HubLifetimeContext context)
...
+            return hubRegistered && connectionIdMatches;
+        }
+
+        public bool CanCleanUpConnection(HubLifetimeContext context)
+        {
+            bool hubRegistered = ConnectionIds.TryGetValue(context.Hub.GetType(), out string? registeredConnectionId);
+            bool connectionIdMatches = registeredConnectionId == context.Context.ConnectionId;
+
+            return hubRegistered && connectionIdMatches;
+        }
```
bdach, line 91: Readers will note that the content of this method is mostly moved from `ConcurrentConnectionLimiter`, except for the fact that the `tokenIdMatches` check is gone.

As far as I can tell that check was mostly dead / bogus anyway; the signalr communication protocol is such that the JWT is only sent at the start of a connection, after which a protocol switch to webscokets happens, and headers are not sent anymore. Which, in turn, means that the JWT is not being sent over the wire, and the previous check would just keep checking the same stored JWT from the initial connection, probably fetching it from a dictionary keyed by connection ID or something. Thus, the connection ID should be more than enough to enforce concurrency at this point.

peppy:
> We can probably shutter the old path in a few months - or I can just delete it now if that is considered acceptable.

I'd keep it available until next release at least (as in the one that doesn't happen this week).

smoogipoo (approved):

osu.Server.Spectator/Entities/ConnectionState.cs
```diff
@@ -42,5 +58,37 @@ public ConnectionState(HubLifetimeContext context)
...
+
+        public bool CanCleanUpConnection(HubLifetimeContext context)
+        {
+            bool hubRegistered = ConnectionIds.TryGetValue(context.Hub.GetType(), out string? registeredConnectionId);
+            bool connectionIdMatches = registeredConnectionId == context.Context.ConnectionId;
+
+            return hubRegistered && connectionIdMatches;
```
peppy, line 84: Is there some special reason for having these two methods exist rather than a single method? They are doing the exact same thing now.

bdach: `HubLifetimeContext` and `HubInvocationContext` do not have a common superclass despite the methods in question indeed being mirror images of each other.

peppy: right, missed that. having the method name the same might help in that case, if you can think of a name which works for both.

peppy: nevermind, what you PR'd works fine.

peppy (approved):

PR #243, opened 2024-09-17
Exclude restricted users from daily challenge-related queries

bdach:
This is a port of https://github.com/ppy/osu-web/pull/11488 for spectator server.

Probably worth examining from a query performance angle. I haven't really even attempted to do so because I don't have a representative database to get realistic query plans from.


peppy (approved):
Should be fine, no need for profiling.

PR #249, opened 2024-11-22
Update databased room status depending on gameplay state

smoogipoo:
Prereqs:
- [ ] https://github.com/ppy/osu-web/pull/11681


bdach:
Adding a TODO item to do a web deploy with the change linked; there is a tag from Nov 22, but it doesn't contain the relevant db migration I believe.

bdach (approved):

peppy (approved):

PR #251, opened 2024-12-24
Add support for "freestyle" in multiplayer

smoogipoo:
- [x] Depends on https://github.com/ppy/osu/pull/31260 for model changes


peppy:
@smoogipoo I've pushed some changes here, can you please review them? Important one is 00aa32a which optimises the validity check for cases with large player counts in rooms.

peppy (approved):

smoogipoo:
The code reads correctly to me.

PR #252, opened 2025-01-02
Signal clients on friend presence state changes

smoogipoo:
Prereqs:
- [ ] See structural discussion at https://github.com/ppy/osu/pull/31389
- [ ] Add signalling from osu-web when friends are added and removed. How should this be done?

### DIFF
--- a/osu.Server.Spectator/Hubs/Metadata/MetadataHub_Friends.cs
+++ b/osu.Server.Spectator/Hubs/Metadata/MetadataHub_Friends.cs
@@ -1,0 +1,34 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+using System.Threading.Tasks;
+
+namespace osu.Server.Spectator.Hubs.Metadata
+{
+    public partial class MetadataHub
+    {
+        private async Task registerFriends(MetadataClientState state)
+        {
+            using (var db = databaseFactory.GetInstance())
+            {
+                foreach (var friend in await db.GetUserFriends(state.UserId))
+                    await Groups.AddToGroupAsync(Context.ConnectionId, friend_presence_watchers(friend.zebra_id));
+            }
+
+            await Clients.Group(friend_presence_watchers(state.UserId)).FriendConnected(state.UserId);
+        }
+
+        private async Task unregisterFriends(MetadataClientState state)
+        {
+            using (var db = databaseFactory.GetInstance())
+            {
+                foreach (var friend in await db.GetUserFriends(state.UserId))
+                    await Groups.RemoveFromGroupAsync(Context.ConnectionId, friend_presence_watchers(friend.zebra_id));
+            }
+
+            await Clients.Group(friend_presence_watchers(state.UserId)).FriendDisconnected(state.UserId);
+        }
+
+        private static string friend_presence_watchers(int userId) => $"metadata:online-presence-watchers:{userId}";
+    }
+}
--- a/osu.Server.Spectator/Hubs/Metadata/MetadataHub_Friends.cs
+++ b/osu.Server.Spectator/Hubs/Metadata/MetadataHub_Friends.cs
@@ -11,7 +11,7 @@ namespace osu.Server.Spectator.Hubs.Metadata
         {
             using (var db = databaseFactory.GetInstance())
             {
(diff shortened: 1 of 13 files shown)


osu.Server.Spectator/Hubs/Metadata/MetadataHub_Friends.cs
```diff
@@ -0,0 +1,34 @@
...
+namespace osu.Server.Spectator.Hubs.Metadata
+{
+    public partial class MetadataHub
+    {
+        private async Task registerFriends(MetadataClientState state)
+        {
+            using (var db = databaseFactory.GetInstance())
+            {
+                foreach (var friend in await db.GetUserFriends(state.UserId))
+                    await Groups.AddToGroupAsync(Context.ConnectionId, friend_presence_watchers(friend.zebra_id));
```
peppy, line 15: Hmm, as discussed IRL there's two other ways we could go about this, and I'm not sure which is best.

- The way this is implemented, 1000 users connecting with 1000 friends each will be creating 1M channels. The overhead of broadcast friend presence updates is low, but the initial connect overhead is high.
- An alternative would be doing the inverse – connecting user registers against a "notify-me" channel and when a friend connects, a ping would be sent to each friend. This could reduce the channel overhead to 1000 in the above scenario, but on connect we would need to iterate all friends and send messages out (slightly higher db overhead, but probably file). If we use this method, broadcasting further updates (ie. if we want to convey activity state changes) would not be possible using the new channel subscriptions.
- The way bancho did this was to send the friends list to the client, and then broadcast all online/offline changes and leave everything to the client. I'd lean towards this method – because users are wanting to see the dashboard show all online users like stable does – but I'm not sure about the performance overheads of having such a large number of SignalR broadcasts in the background (especially since it won't stop during gameplay).

I think I'm leaning towards what you have here as an initial implementation since it's simple to get out and if it's going to fall over, we'll know pretty quickly (`server-spectator` will probably hit memory or performance issues on a sizable number of users connecting).

peppy: @bdach this one is still in early discussion but you may have something to say here.

peppy: > I'm confused by this part however, because we already have the dashboard working for all online users, and it works because users implicitly opt into the broadcasts by opening the dashboard

in stable, users are shown regardless of whether they are playing a beatmap or not. we're eventually going to want this to be a thing here too.

bdach: > in stable, users are shown regardless of whether they are playing a beatmap or not

I am further confused because again, to the best of my knowledge, that's what we already have on lazer...?

https://github.com/ppy/osu-server-spectator/blob/63bb290bb9d01d268409b296618caba22b46a790/osu.Server.Spectator/Hubs/Metadata/MetadataHub.cs#L82-L86

`GetAllStates()` here is going to list all users connected to the metadata hub, which has nothing to do with whether they're playing beatmaps or not.

peppy: Wow, I guess it is. I was sure it was still only for actively playing.

smoogipoo: > In terms of observability, a ddog gauge on the number of active groups could prove useful.

This doesn't look easy to do, because SignalR itself doesn't expose any tracking variable. It looks like you can override `Hub.Groups` with a custom `IGroupManager` which could implement our own tracking, but it sounds difficult to do properly - have to consider SignalR's automatic cleanup on disconnection, SignalR not telling us the groups which a user's in, concurrency, etc...

> Also, something like https://stackoverflow.com/questions/12651169/signalr-and-large-number-of-groups (not sure how up-to-date it is) may prove to be an annoying snag.

This looks to be more users with many groups, rather than groups with many users or many groups in general. While not super applicable to us anyway, I'm crossing my fingers it won't ever become an issue - the only reference I can find is this [`connection.GroupNames`](https://github.com/dotnet/aspnetcore/blob/7ba02e8ac71eeba562476953d5ff6030b8eda765/src/SignalR/server/Core/src/DefaultHubLifetimeManager.cs#L46-L48) member that doesn't look to be serialised anywhere.

smoogipoo:
Superseded by https://github.com/ppy/osu-server-spectator/pull/255

PR #255, opened 2025-01-07
Always broadcast user presence to users' friends

smoogipoo:
Supersedes https://github.com/ppy/osu-server-spectator/pull/252
See client side changes in https://github.com/ppy/osu/pull/31444

Note that this doesn't update when the user adds/removes friends. That's left as a future task for when we figure out a method to signal changes to/from osu-web (redis, polling the database, or otherwise).

### DIFF
--- a/osu.Server.Spectator/Entities/EntityStore.cs
+++ b/osu.Server.Spectator/Entities/EntityStore.cs
@@ -110,6 +110,39 @@ namespace osu.Server.Spectator.Entities
             throw new TimeoutException("Could not allocate new entity after multiple retries. Something very bad has happened");
         }
 
+        /// <summary>
+        /// Attempts to retrieve an existing entity with a lock for use.
+        /// </summary>
+        /// <param name="id">The ID of the requested entity.</param>
+        /// <returns>An existing <see cref="ItemUsage{T}"/> which allows reading or writing the item, or <c>null</c> if no entity exists. This should be disposed after usage.</returns>
+        public async Task<ItemUsage<T>?> TryGetForUse(long id)
+        {
+            TrackedEntity? item;
+
+            lock (entityMapping)
+            {
+                if (!entityMapping.TryGetValue(id, out item))
+                {
+                    DogStatsd.Increment($"{statsDPrefix}.get-notfound");
+                    return null;
+                }
+            }
+
+            try
+            {
+                await item.ObtainLockAsync();
+            }
+            // this may be thrown if the item was destroyed between when we retrieved the item usage and took the lock.
+            catch (InvalidOperationException)
+            {
+                DogStatsd.Increment($"{statsDPrefix}.get-notfound");
+                return null;
+            }
+
+            DogStatsd.Increment($"{statsDPrefix}.get");
+            return new ItemUsage<T>(item);
+        }
+
         public async Task Destroy(long id)
         {
             TrackedEntity? item;
--- a/osu.Server.Spectator/Hubs/Metadata/MetadataHub.cs
+++ b/osu.Server.Spectator/Hubs/Metadata/MetadataHub.cs
@@ -29,7 +29,7 @@ namespace osu.Server.Spectator.Hubs.Metadata
         private readonly IScoreProcessedSubscriber scoreProcessedSubscriber;
 
         internal const string ONLINE_PRESENCE_WATCHERS_GROUP = "metadata:online-presence-watchers";
-        private static string friend_presence_watchers(int userId) => $"metadata:online-presence-watchers:{userId}";
+        internal static string FRIEND_PRESENCE_WATCHERS_GROUP(int userId) => $"metadata:online-presence-watchers:{userId}";
 
         internal static string MultiplayerRoomWatchersGroup(long roomId) => $"metadata:multiplayer-room-watchers:{roomId}";
 
@@ -77,7 +77,7 @@ namespace osu.Server.Spectator.Hubs.Metadata
                 using (var db = databaseFactory.GetInstance())
                 {
                     foreach (int friendId in await db.GetUserFriendsAsync(usage.Item.UserId))
-                        await Groups.AddToGroupAsync(Context.ConnectionId, friend_presence_watchers(friendId));
+                        await Groups.AddToGroupAsync(Context.ConnectionId, FRIEND_PRESENCE_WATCHERS_GROUP(friendId));
                 }
             }
         }
@@ -201,7 +201,7 @@ namespace osu.Server.Spectator.Hubs.Metadata
             using (var db = databaseFactory.GetInstance())
             {
                 foreach (int friendId in await db.GetUserFriendsAsync(state.UserId))
-                    await Groups.RemoveFromGroupAsync(state.ConnectionId, friend_presence_watchers(friendId));
+                    await Groups.RemoveFromGroupAsync(state.ConnectionId, FRIEND_PRESENCE_WATCHERS_GROUP(friendId));
             }
         }
 
@@ -210,7 +210,7 @@ namespace osu.Server.Spectator.Hubs.Metadata
             if (userPresence?.Status == UserStatus.Offline)
                 userPresence = null;
 
-            return Clients.Groups(friend_presence_watchers(userId), ONLINE_PRESENCE_WATCHERS_GROUP).UserPresenceUpdated(userId, userPresence);
+            return Clients.Groups(FRIEND_PRESENCE_WATCHERS_GROUP(userId), ONLINE_PRESENCE_WATCHERS_GROUP).UserPresenceUpdated(userId, userPresence);
         }
     }
 }
--- a/osu.Server.Spectator/Hubs/Metadata/MetadataHub.cs
+++ b/osu.Server.Spectator/Hubs/Metadata/MetadataHub.cs
@@ -66,11 +66,7 @@ namespace osu.Server.Spectator.Hubs.Metadata
                         versionHash = versionHash.Substring(versionHash.Length - 82, 32);
                 }
 
-                usage.Item = new MetadataClientState(Context.ConnectionId, Context.GetUserId(), versionHash)
-                {
-                    UserStatus = UserStatus.Online
-                };
-
+                usage.Item = new MetadataClientState(Context.ConnectionId, Context.GetUserId(), versionHash);
                 await broadcastUserPresenceUpdate(usage.Item.UserId, usage.Item.ToUserPresence());
                 await Clients.Caller.DailyChallengeUpdated(dailyChallengeUpdater.Current);
 
--- a/osu.Server.Spectator/Hubs/Metadata/MetadataHub.cs
+++ b/osu.Server.Spectator/Hubs/Metadata/MetadataHub.cs
@@ -78,9 +78,9 @@ namespace osu.Server.Spectator.Hubs.Metadata
                         await Groups.AddToGroupAsync(Context.ConnectionId, FRIEND_PRESENCE_WATCHERS_GROUP(friendId));
 
                         // Check if the friend is online, and if they are, broadcast to the connected user.
-                        using (var friendUsage = await GetStateFromUser(friendId))
+                        using (var friendUsage = await TryGetStateFromUser(friendId))
                         {
-                            if (friendUsage.Item != null && shouldBroadcastPresenceToOtherUsers(friendUsage.Item))
+                            if (friendUsage?.Item != null && shouldBroadcastPresenceToOtherUsers(friendUsage.Item))
                                 await Clients.Caller.UserPresenceUpdated(friendId, friendUsage.Item.ToUserPresence());
                         }
                     }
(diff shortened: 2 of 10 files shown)


osu.Server.Spectator/Hubs/Metadata/MetadataHub.cs
```diff
@@ -186,14 +197,20 @@ protected override async Task CleanUpState(MetadataClientState state)
             await base.CleanUpState(state);
             await broadcastUserPresenceUpdate(state.UserId, null);
             await scoreProcessedSubscriber.UnregisterFromAllMultiplayerRoomsAsync(state.UserId);
+
+            using (var db = databaseFactory.GetInstance())
+            {
+                foreach (int friendId in await db.GetUserFriendsAsync(state.UserId))
+                    await Groups.RemoveFromGroupAsync(state.ConnectionId, FRIEND_PRESENCE_WATCHERS_GROUP(friendId));
+            }
```
peppy, line 205: Ignoring the fact that this isn't reliable because the database friends may have changed since the original connection (imagine a user removed all their friends before quitting), I'm 99% sure we don't need to do this. We don't seem to do this anywhere else and so far it hasn't been an issue. One would *hope* signalr does this for you.

Reference:  https://learn.microsoft.com/en-us/aspnet/core/signalr/groups?view=aspnetcore-9.0

> Group membership isn't preserved when a connection reconnects. The connection needs to rejoin the group when it's re-established. It's not possible to count the members of a group, since this information isn't available if the application is scaled to multiple servers.

```suggestion
```

peppy (commented):
Seems pretty logical

osu.Server.Spectator/Hubs/Metadata/MetadataHub.cs
```diff
@@ -65,9 +66,19 @@ public override async Task OnConnectedAsync()
...
+                    UserStatus = UserStatus.Online
+                };
+
                 await broadcastUserPresenceUpdate(usage.Item.UserId, usage.Item.ToUserPresence());
                 await Clients.Caller.DailyChallengeUpdated(dailyChallengeUpdater.Current);
+
+                using (var db = databaseFactory.GetInstance())
+                {
+                    foreach (int friendId in await db.GetUserFriendsAsync(usage.Item.UserId))
+                        await Groups.AddToGroupAsync(Context.ConnectionId, FRIEND_PRESENCE_WATCHERS_GROUP(friendId));
```
peppy, line 80: This doesn't send the initial "online" presence to the user in the case a friend is online. I think this is required.

osu.Server.Spectator/Hubs/Metadata/MetadataHub.cs
```diff
@@ -65,9 +66,19 @@ public override async Task OnConnectedAsync()
                         versionHash = versionHash.Substring(versionHash.Length - 82, 32);
                 }
 
-                usage.Item = new MetadataClientState(Context.ConnectionId, Context.GetUserId(), versionHash);
+                usage.Item = new MetadataClientState(Context.ConnectionId, Context.GetUserId(), versionHash)
+                {
+                    UserStatus = UserStatus.Online
+                };
```
peppy, line 72: This isn't going to work for the "appear offline" state. The initial broadcast of "online"-ness needs to not occur until the client updates its status via `UpdateStatus`. I'm already doing some work around this because it's broken in other ways, though. I think this PR can get away with just not having this here (in testing it still works fine).


```suggestion
                usage.Item = new MetadataClientState(Context.ConnectionId, Context.GetUserId(), versionHash);
```

osu.Server.Spectator/Hubs/Metadata/MetadataHub.cs
smoogipoo: Indeed unnecessary, removed.

osu.Server.Spectator/Hubs/Metadata/MetadataHub.cs
smoogipoo: Makes sense, I only added this because I noticed a test failure. Have adjusted.

smoogipoo: This was an intentional choice, I haven't tested how stable does it. Basically you want to receive a giga notification on startup if 100 of your friends are online?

smoogipoo: I think I understand your concern now, have applied a fix in https://github.com/ppy/osu-server-spectator/pull/255/commits/fbdf5918ff46b56e8a2c742b9c83303c9e6d8f78.

peppy: > Edit: Stable doesn't notify either.

It definitely does for me.. unless I'm hallucinating.

https://github.com/user-attachments/assets/558dfcdd-b190-4eda-9f39-c312038ab12c

Maybe you had this turned off in settings?

osu.Server.Spectator/Entities/EntityStore.cs
```diff
@@ -110,6 +110,39 @@ public async Task<ItemUsage<T>> GetForUse(long id, bool createOnMissing = false)
...
+                if (!entityMapping.TryGetValue(id, out item))
+                {
+                    DogStatsd.Increment($"{statsDPrefix}.get-notfound");
+                    return null;
+                }
+            }
+
+            try
+            {
+                await item.ObtainLockAsync();
```
peppy, line 133: For this particular case, do we actually need a lock? Wondering if it can be avoided..

osu.Server.Spectator/Hubs/Metadata/MetadataHub.cs
```diff
@@ -222,7 +238,11 @@ private Task broadcastUserPresenceUpdate(int userId, UserPresence? userPresence)
             // we never want appearing offline users to have their status broadcast to other clients.
             Debug.Assert(userPresence?.Status != UserStatus.Offline);
 
-            return Clients.Group(ONLINE_PRESENCE_WATCHERS_GROUP).UserPresenceUpdated(userId, userPresence);
+            return Task.WhenAll
+            (
+                Clients.Group(ONLINE_PRESENCE_WATCHERS_GROUP).UserPresenceUpdated(userId, userPresence),
+                Clients.Group(FRIEND_PRESENCE_WATCHERS_GROUP(userId)).FriendPresenceUpdated(userId, userPresence)
```
peppy, line 244: I guess clients are going to get double status updates for friends, but probably fine for now.

peppy (commented):
Looks mostly good

osu.Server.Spectator/Entities/EntityStore.cs
smoogipoo: Yes, because `ItemUsage.Item` enforces that the item must be locked:

https://github.com/ppy/osu-server-spectator/blob/718fb5a44739cecbc3a7a243dd0699d7fb75266a/osu.Server.Spectator/Entities/EntityStore.cs#L195-L201

A deadlock concern is valid here, saved by the fact that `EntityStore` will continuously retry retrieval of the lock over 5 seconds. So I think it'll be fine after all.

I've tried to resolve this in a more "proper" way, and have come up with the following. I'm not sure if it's over-engineered (the hub part), and likely not over-engineered _enough_:

```diff
diff --git a/osu.Server.Spectator/Entities/EntityStore.cs b/osu.Server.Spectator/Entities/EntityStore.cs
index ebeb4dc..42b7af2 100644
--- a/osu.Server.Spectator/Entities/EntityStore.cs
+++ b/osu.Server.Spectator/Entities/EntityStore.cs
@@ -6,6 +6,7 @@ using System.Collections.Generic;
 using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
+using JetBrains.Annotations;
 using osu.Framework.Extensions.ObjectExtensions;
 using StatsdClient;
 
@@ -32,13 +33,19 @@ namespace osu.Server.Spectator.Entities
         {
             get
             {
-                lock (entityMapping)
+                using (GetRetrievalLock())
                     return entityMapping.Count;
             }
         }
 
         public string EntityName => typeof(T).Name;
 
+        /// <summary>
+        /// Locks this <see cref="EntityStore{T}"/> for retrieval of an entity.
+        /// </summary>
+        [MustDisposeResource]
+        public EntityStoreLock GetRetrievalLock() => new EntityStoreLock(entityMapping);
+
         /// <summary>
         /// Retrieves an entity.
         /// </summary>
@@ -49,7 +56,7 @@ namespace osu.Server.Spectator.Entities
         /// <returns>The entity.</returns>
         public T? GetEntityUnsafe(long id)
         {
-            lock (entityMapping)
+            using (GetRetrievalLock())
                 return !entityMapping.TryGetValue(id, out var entity) ? null : entity.GetItemUnsafe();
         }
 
@@ -68,7 +75,7 @@ namespace osu.Server.Spectator.Entities
             {
                 TrackedEntity? item;
 
-                lock (entityMapping)
+                using (GetRetrievalLock())
                 {
                     if (!entityMapping.TryGetValue(id, out item))
                     {
@@ -119,7 +126,7 @@ namespace osu.Server.Spectator.Entities
         {
             TrackedEntity? item;
 
-            lock (entityMapping)
+            using (GetRetrievalLock())
             {
                 if (!entityMapping.TryGetValue(id, out item))
                 {
@@ -147,7 +154,7 @@ namespace osu.Server.Spectator.Entities
         {
             TrackedEntity? item;
 
-            lock (entityMapping)
+            using (GetRetrievalLock())
             {
                 if (!entityMapping.TryGetValue(id, out item))
                     // was not tracking.
@@ -173,7 +180,7 @@ namespace osu.Server.Spectator.Entities
         /// </summary>
         public KeyValuePair<long, T>[] GetAllEntities()
         {
-            lock (entityMapping)
+            using (GetRetrievalLock())
             {
                 return entityMapping
                        .Where(kvp => kvp.Value.GetItemUnsafe() != null)
@@ -187,7 +194,7 @@ namespace osu.Server.Spectator.Entities
         /// </summary>
         public void Clear()
         {
-            lock (entityMapping)
+            using (GetRetrievalLock())
             {
                 entityMapping.Clear();
             }
@@ -195,7 +202,7 @@ namespace osu.Server.Spectator.Entities
 
         private void remove(long id)
         {
-            lock (entityMapping)
+            using (GetRetrievalLock())
             {
                 entityMapping.Remove(id);
 
@@ -290,5 +297,28 @@ namespace osu.Server.Spectator.Entities
                 if (shouldBeLocked && !isLocked) throw new InvalidOperationException("Attempted to access a tracked entity without holding a lock");
             }
         }
+
+        [MustDisposeResource]
+        public class EntityStoreLock : IDisposable
+        {
+            private readonly object lockObject;
+
+            public EntityStoreLock(object lockObject)
+            {
+                this.lockObject = lockObject;
+                Monitor.Enter(lockObject);
+            }
+
+            public void Dispose()
+            {
+                try
+                {
+                    Monitor.Exit(lockObject);
+                }
+                catch
+                {
+                }
+            }
+        }
     }
 }
diff --git a/osu.Server.Spectator/Hubs/Metadata/MetadataHub.cs b/osu.Server.Spectator/Hubs/Metadata/MetadataHub.cs
index 9a3e12c..65ddd0b 100644
--- a/osu.Server.Spectator/Hubs/Metadata/MetadataHub.cs
+++ b/osu.Server.Spectator/Hubs/Metadata/MetadataHub.cs
@@ -52,6 +52,8 @@ namespace osu.Server.Spectator.Hubs.Metadata
         {
             await base.OnConnectedAsync();
 
+            var friendUserIds = new List<int>();
+
             using (var usage = await GetOrCreateLocalUserState())
             {
                 string? versionHash = null;
@@ -76,16 +78,31 @@ namespace osu.Server.Spectator.Hubs.Metadata
                     foreach (int friendId in await db.GetUserFriendsAsync(usage.Item.UserId))
                     {
                         await Groups.AddToGroupAsync(Context.ConnectionId, FRIEND_PRESENCE_WATCHERS_GROUP(friendId));
-
-                        // Check if the friend is online, and if they are, broadcast to the connected user.
-                        using (var friendUsage = await TryGetStateFromUser(friendId))
-                        {
-                            if (friendUsage?.Item != null && shouldBroadcastPresenceToOtherUsers(friendUsage.Item))
-                                await Clients.Caller.FriendPresenceUpdated(friendId, friendUsage.Item.ToUserPresence());
-                        }
+                        friendUserIds.Add(friendId);
                     }
                 }
             }
+
+            // Broadcast online friend states to the newly connected user.
+
+            // We're using multiple user states here...
+            using (UserStates.GetRetrievalLock())
+            {
+                using var userUsage = await TryGetStateFromUser(Context.GetUserId());
+
+                if (userUsage?.Item == null)
+                    return;
+
+                foreach (int friendId in friendUserIds)
+                {
+                    using var friendUsage = await TryGetStateFromUser(friendId);
+
+                    if (friendUsage?.Item == null || !shouldBroadcastPresenceToOtherUsers(friendUsage.Item))
+                        continue;
+
+                    await Clients.Caller.FriendPresenceUpdated(friendId, friendUsage.Item.ToUserPresence());
+                }
+            }
         }
 
         private async Task logLogin(ItemUsage<MetadataClientState> usage)

```

peppy (approved):

PR #256, opened 2025-01-08
Avoid sending user presence updates when nothing has changed

peppy:
Noticed while testing that every client will send three presence updates on fresh connect to main menu.

This is testing the "user just starts game". If a user is, for instance, at song select, it will still trigger two updates (second being for the activity). The goal here is to optimise the most common flow to reduce overhead incurred on both client and server.

### DIFF
--- a/osu.Server.Spectator/Hubs/Metadata/MetadataHub.cs
+++ b/osu.Server.Spectator/Hubs/Metadata/MetadataHub.cs
@@ -110,6 +110,10 @@ namespace osu.Server.Spectator.Hubs.Metadata
             using (var usage = await GetOrCreateLocalUserState())
             {
                 Debug.Assert(usage.Item != null);
+
+                if (usage.Item.UserActivity == null && activity == null)
+                    return;
+
                 usage.Item.UserActivity = activity;
 
                 await broadcastUserPresenceUpdate(usage.Item.UserId, usage.Item.ToUserPresence());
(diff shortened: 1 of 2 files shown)


osu.Server.Spectator/Hubs/Metadata/MetadataHub.cs
```diff
@@ -110,6 +110,10 @@ public async Task UpdateActivity(UserActivity? activity)
             using (var usage = await GetOrCreateLocalUserState())
             {
                 Debug.Assert(usage.Item != null);
+
+                if (usage.Item.UserActivity == null && activity == null)
+                    return;
```
bdach, line 115: I'm a little weirded out by the fact that this method explicitly checks for both nulls while the other one checks equality... Like I'd expect

```diff
diff --git a/osu.Server.Spectator/Hubs/Metadata/MetadataHub.cs b/osu.Server.Spectator/Hubs/Metadata/MetadataHub.cs
index b861357..a77159c 100644
--- a/osu.Server.Spectator/Hubs/Metadata/MetadataHub.cs
+++ b/osu.Server.Spectator/Hubs/Metadata/MetadataHub.cs
@@ -111,12 +111,11 @@ namespace osu.Server.Spectator.Hubs.Metadata
             {
                 Debug.Assert(usage.Item != null);
 
-                if (usage.Item.UserActivity == null && activity == null)
-                    return;
-
-                usage.Item.UserActivity = activity;
-
-                await broadcastUserPresenceUpdate(usage.Item.UserId, usage.Item.ToUserPresence());
+                if (usage.Item.UserActivity != activity)
+                {
+                    usage.Item.UserActivity = activity;
+                    await broadcastUserPresenceUpdate(usage.Item.UserId, usage.Item.ToUserPresence());
+                }
             }
         }
 

```

to mirror the other method, and it does seem to pass tests - then again, `UserActivity` is a class and it doesn't implement equality members so maybe equality checking is a bit useless here. Is _that_ the reason why these are explicit null checks?

peppy: Yeah precisely as you say – equality is not implemented so I wanted to be very explicit about this only avoiding the null-to-not-null transition.

bdach (approved):

PR #257, opened 2025-01-08
Ensure user state is never conveyed for "appear offline" users

peppy:
Noticed during testing that "appear offline" doesn't actually work under the surface. If someone was to watch packets, they would be able to see invisible users and track their activity.

This change ensures that

- For fresh logins where the user already has their client set to "appear offline", nothing will be conveyed to other users.
- For existing logins, after a user changes their state to "appear offline" they will no longer convey anything to other users.

Of important note, I think this is only the start of the issue, as spectator server is probably still revealing play state to anyone that subscribes for it.. Fixing this either requires merging hubs (has been discussed IRL for other reasons) or possibly having the client conveying its "online" state when beginning a play session.

(Not touching on that today, will open an issue)

- [x] Depends on https://github.com/ppy/osu-server-spectator/pull/256 for mergeability

### DIFF
--- a/osu.Server.Spectator/Hubs/Metadata/MetadataHub.cs
+++ b/osu.Server.Spectator/Hubs/Metadata/MetadataHub.cs
@@ -68,7 +68,6 @@ namespace osu.Server.Spectator.Hubs.Metadata
                 usage.Item = new MetadataClientState(Context.ConnectionId, Context.GetUserId(), versionHash);
 
                 await logLogin(usage);
-                await broadcastUserPresenceUpdate(usage.Item.UserId, usage.Item.ToUserPresence());
                 await Clients.Caller.DailyChallengeUpdated(dailyChallengeUpdater.Current);
             }
         }
@@ -116,7 +115,8 @@ namespace osu.Server.Spectator.Hubs.Metadata
 
                 usage.Item.UserActivity = activity;
 
-                await broadcastUserPresenceUpdate(usage.Item.UserId, usage.Item.ToUserPresence());
+                if (shouldBroadcastPresentToOtherUsers(usage.Item))
+                    await broadcastUserPresenceUpdate(usage.Item.UserId, usage.Item.ToUserPresence());
             }
         }
 
@@ -129,7 +129,14 @@ namespace osu.Server.Spectator.Hubs.Metadata
                 if (usage.Item.UserStatus != status)
                 {
                     usage.Item.UserStatus = status;
-                    await broadcastUserPresenceUpdate(usage.Item.UserId, usage.Item.ToUserPresence());
+
+                    if (status == UserStatus.Offline)
+                    {
+                        // special case of users that already broadcast that they are online switching to "appear offline".
+                        await broadcastUserPresenceUpdate(usage.Item.UserId, null);
+                    }
+                    else
+                        await broadcastUserPresenceUpdate(usage.Item.UserId, usage.Item.ToUserPresence());
                 }
             }
         }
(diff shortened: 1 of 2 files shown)


osu.Server.Spectator/Hubs/Metadata/MetadataHub.cs
```diff
@@ -68,7 +68,6 @@ public override async Task OnConnectedAsync()
                 usage.Item = new MetadataClientState(Context.ConnectionId, Context.GetUserId(), versionHash);
 
                 await logLogin(usage);
-                await broadcastUserPresenceUpdate(usage.Item.UserId, usage.Item.ToUserPresence());
```
bdach, line 71: Just to confirm intent, this is deleted because it's basically conveying nothing, correct? I'm not really sure why I put that there in the first place but I guess that makes sense.

osu.Server.Spectator/Hubs/Metadata/MetadataHub.cs
```diff
@@ -205,16 +212,36 @@ public async Task EndWatchingMultiplayerRoom(long id)
...
         {
-            if (userPresence?.Status == UserStatus.Offline)
-                userPresence = null;
+            // we never want appearing offline users to have their status broadcast to other clients.
+            Debug.Assert(userPresence?.Status != UserStatus.Offline);
 
             return Clients.Group(ONLINE_PRESENCE_WATCHERS_GROUP).UserPresenceUpdated(userId, userPresence);
         }
+
+        private bool shouldBroadcastPresentToOtherUsers(MetadataClientState state)
```
bdach, line 228: Method name is typo'd:

```suggestion
        private bool shouldBroadcastPresenceToOtherUsers(MetadataClientState state)
```

(don't apply directly for obvious reasons)

osu.Server.Spectator/Hubs/Metadata/MetadataHub.cs
```diff
@@ -129,7 +129,14 @@ public async Task UpdateStatus(UserStatus? status)
                 if (usage.Item.UserStatus != status)
                 {
                     usage.Item.UserStatus = status;
-                    await broadcastUserPresenceUpdate(usage.Item.UserId, usage.Item.ToUserPresence());
+
+                    if (status == UserStatus.Offline)
```
bdach, line 133: Any reason why this is a local check that isn't using the helper method? (`shouldBroadcastPresentToOtherUsers()` I mean)

osu.Server.Spectator/Hubs/Metadata/MetadataHub.cs
peppy: Yeah, it should never fire anything.

osu.Server.Spectator/Hubs/Metadata/MetadataHub.cs
peppy: It's actually a different condition – this case we explicitly *want* to sent the null status just once when a user transitions to "appear offline". See comment just below, hopefully.

smoogipoo (approved):
I'm not entirely convinced on littering the rest of the code with `if (shouldBroadcast()` as opposed to handling it directly inside the method with perhaps a `force` param, but I don't care enough to justify spending time making a review on that.

peppy:
> as opposed to handling it directly inside the method with perhaps a force param, but I don't care enough to justify spending time making a review on that.

i tried this and it didn't read as well as how i have it now. if we end up having this in more places, we could reconsider of course.

PR #260, opened 2025-01-16
Fix connection errors on self-friends

smoogipoo:
Noticed this when doing some hacky self-testing (manual insert into `phpbb_zebra`). We actually have/had such cases in the production database.


bdach (approved):
Error or no, even business-logic-/functionality-wise doing anything on "self-friends" like this makes no sense to me, so sure.

PR #261, opened 2025-01-16
Always broadcast local user's state back to themselves

smoogipoo:


osu.Server.Spectator/Hubs/Metadata/MetadataHub.cs
```diff
@@ -131,8 +131,13 @@ public async Task UpdateActivity(UserActivity? activity)
...
 
-                if (shouldBroadcastPresenceToOtherUsers(usage.Item))
-                    await broadcastUserPresenceUpdate(usage.Item.UserId, usage.Item.ToUserPresence());
+                await Task.WhenAll
+                (
+                    shouldBroadcastPresenceToOtherUsers(usage.Item)
+                        ? broadcastUserPresenceUpdate(usage.Item.UserId, usage.Item.ToUserPresence())
+                        : Task.CompletedTask,
+                    Clients.Caller.UserPresenceUpdated(usage.Item.UserId, usage.Item.ToUserPresence())
+                );
```
peppy, line 140: I really prefer doing this as 2x`await` with explicit `if` for readability (I don't think performance concerns are valid for these kinds of usages?).

smoogipoo: In that case I'll also remove the one in broadcast, yes?

peppy: The one below? Yeah I guess. The inline ternary if in here is specifically annoying to my eye.

peppy: Let's leave this for now. In an IDE it's bearable legibility wise, just looks bad on github review.

peppy (approved):

PR #265, opened 2025-01-23
Add support for creating, joining, and parting osu!web rooms via interop

smoogipoo:
- [x] Depends on https://github.com/ppy/osu-web/pull/11811
- Depends on https://github.com/ppy/osu/pull/31637 for new method added to the interface.

This server needs the ability to manage rooms for matchmaking. Because I intend to use the existing `MultiplayerHub` for matchmaking messages, I took a bit of a detour and made it the single endpoint for creating multiplayer rooms so that the client will no longer need to juggle two endpoints for doing certain things.

I see a future where we move playlists, daily challenge, and even the lounge to this server, but I'm taking small meaningful steps because matchmaking is the goal for now.

### DIFF
--- a/osu.Server.Spectator/Hubs/Multiplayer/MultiplayerHub.cs
+++ b/osu.Server.Spectator/Hubs/Multiplayer/MultiplayerHub.cs
@@ -48,7 +48,7 @@ namespace osu.Server.Spectator.Hubs.Multiplayer
         public async Task<MultiplayerRoom> CreateRoom(MultiplayerRoom room)
         {
             Log($"{Context.GetUserId()} creating room");
-            return await JoinRoomWithPassword(await legacyIO.CreateRoom(Context.GetUserId(), room), room.Settings.Password);
+            return await JoinRoomWithPassword(await legacyIO.CreateRoomAsync(Context.GetUserId(), room), room.Settings.Password);
         }
 
         public Task<MultiplayerRoom> JoinRoom(long roomId) => JoinRoomWithPassword(roomId, string.Empty);
@@ -63,7 +63,7 @@ namespace osu.Server.Spectator.Hubs.Multiplayer
                     throw new InvalidStateException("Can't join a room when restricted.");
             }
 
-            await legacyIO.JoinRoom(roomId, Context.GetUserId());
+            await legacyIO.JoinRoomAsync(roomId, Context.GetUserId());
 
             using (var userUsage = await GetOrCreateLocalUserState())
             {
@@ -894,7 +894,7 @@ namespace osu.Server.Spectator.Hubs.Multiplayer
 
         private async Task leaveRoom(MultiplayerClientState state, bool wasKick)
         {
-            await legacyIO.PartRoom(state.CurrentRoomID, state.UserId);
+            await legacyIO.PartRoomAsync(state.CurrentRoomID, state.UserId);
 
             using (var roomUsage = await getLocalUserRoom(state))
                 await leaveRoom(state, roomUsage, wasKick);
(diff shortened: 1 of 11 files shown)


osu.Server.Spectator/Services/LegacyIO.cs
smoogipoo, line 1: The base implementation here was pulled from [`osu-server-beatmap-submission`](https://github.com/ppy/osu-server-beatmap-submission/blob/master/osu.Server.BeatmapSubmission/Services/LegacyIO.cs) with local adjustments for this server's usage (return value and `HubException` handling).

osu.Server.Spectator/Hubs/Multiplayer/MultiplayerHub.cs
```diff
@@ -720,8 +731,8 @@ private async Task removeDatabaseUser(MultiplayerRoom room, MultiplayerRoomUser
 
         protected override async Task CleanUpState(MultiplayerClientState state)
         {
-            await leaveRoom(state, true);
             await base.CleanUpState(state);
+            await leaveRoom(state, true);
```
smoogipoo, line 735: Meaningless change because the base implementation is empty. I just thought I'd make this consistent with other implementations, seeing as `leaveRoom` can throw an exception.

bdach (commented):
Initial pass.

Looks mostly quaint although I do wonder how much discussion on general direction happened in places I can't see or hear.

osu.Server.Spectator/Services/LegacyIO.cs
```diff
@@ -0,0 +1,180 @@
...
+
+        private static string hmacEncode(string input, byte[] key)
+        {
+            byte[] byteArray = Encoding.ASCII.GetBytes(input);
+
+            using (var hmac = new HMACSHA1(key))
+            {
+                byte[] hashArray = hmac.ComputeHash(byteArray);
+                return hashArray.Aggregate(string.Empty, (s, e) => s + $"{e:x2}", s => s);
+            }
```
turbedi, line 99:
```suggestion
            byte[] hashArray = HMACSHA1.HashData(key, byteArray);
            return hashArray.Aggregate(string.Empty, (s, e) => s + $"{e:x2}", s => s);
```
A bit cleaner IMO (and potentially faster)

bdach:
Something is going to need to happen to fix the test failures here. Easiest suggestion would be to not have a static ctor in `AppSettings` and instead just have getters delegate directly to `Environment.GetEnvironmentVariable()`.

osu.Server.Spectator/Services/LegacyIO.cs
```diff
@@ -0,0 +1,180 @@
...
+        {
+            await runLegacyIO(HttpMethod.Put, $"multiplayer/rooms/{roomId}/users/{userId}");
+        }
+
+        public async Task PartRoom(long roomId, int userId)
+        {
+            await runLegacyIO(HttpMethod.Delete, $"multiplayer/rooms/{roomId}/users/{userId}");
+        }
+
+        private class CreateRoomRequest : Room
```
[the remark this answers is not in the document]
peppy: I agree that this is going to be a potential point of failure if we forget to add any changes to fields/properties going forward. But also, after careful reading it seems this isn't only adding the host `user_id` but also populating `CurrentPlaylistItem`.

@smoogipoo might be worth adding some inline commentary as to why this is required. also curious if the `PlaylistItem` clone is just for sanity of if there's an actual reason that it's required.

peppy:
@smoogipoo I made some naming and minor code structure changes, please check you're okay with them.

smoogipoo:
Your changes look fine.

osu.Server.Spectator/Services/LegacyIO.cs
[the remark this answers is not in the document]
smoogipoo: Exactly.

peppy (approved):

PR #269, opened 2025-02-11
Use base ctor to remove unnecessary property copies

smoogipoo:
- Depends on https://github.com/ppy/osu/pull/31637 for new ctor.

See discussion in https://github.com/ppy/osu-server-spectator/pull/265#discussion_r1928348536. Turns out I added these copies to the base class in the osu-side PR (https://github.com/ppy/osu/pull/31637).

### DIFF
--- a/osu.Server.Spectator/Services/LegacyIO.cs
+++ b/osu.Server.Spectator/Services/LegacyIO.cs
@@ -152,16 +152,8 @@ namespace osu.Server.Spectator.Services
             /// Creates a <see cref="Room"/> from a <see cref="MultiplayerRoom"/>.
             /// </summary>
             public RoomWithHostId(MultiplayerRoom room)
+                : base(room)
             {
-                RoomID = room.RoomID;
-                Host = room.Host?.User;
-                Name = room.Settings.Name;
-                Password = room.Settings.Password;
-                Type = room.Settings.MatchType;
-                QueueMode = room.Settings.QueueMode;
-                AutoStartDuration = room.Settings.AutoStartDuration;
-                AutoSkip = room.Settings.AutoSkip;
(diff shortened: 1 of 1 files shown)


osu.Server.Spectator/Services/SharedInterop.cs
```diff
@@ -178,16 +178,8 @@ private class RoomWithHostId : Room
             /// Creates a <see cref="Room"/> from a <see cref="MultiplayerRoom"/>.
             /// </summary>
             public RoomWithHostId(MultiplayerRoom room)
+                : base(room)
             {
-                RoomID = room.RoomID;
-                Host = room.Host?.User;
```
bdach, line 183: This is technically different than [what the base ctor that's being called in the new code does](https://github.com/ppy/osu/blob/bd590786710afeff63c40f65361b290d42f61e2a/osu.Game/Online/Rooms/Room.cs#L358). Did you test that this doesn't cause any weird issues? I'd like to think that it doesn't but this is the sort of nightmare model juggling stuff that breaks more often than I'd want to see.

smoogipoo: Yeah, this is fine.

First, this value is [overwritten](https://github.com/ppy/osu-web/blob/0e49d567e73da910d9340cd08b87143ce8efc570/app/Models/Multiplayer/Room.php#L608) by osu!web. 

Second, `.User` should actually always be null in spectator because it's only used osu-side as a post-processing variable, so the base ctor is more correct in this case, though it doesn't matter because of the above. Now that I say it out loud, the `.User` property should be removed from this model altogether, but that's for another time.

bdach (approved):
just gonna merge and hope that we notice breakage (if any) on staging

PR #271, opened 2025-02-13
Don't use "legacy IO" terminology

peppy:
Given we're using more of this going forward, and this is what osu-web tends to call it these days, let's stick with nomenclature that makes sense.

### DIFF
--- a/osu.Server.Spectator/AppSettings.cs
+++ b/osu.Server.Spectator/AppSettings.cs
@@ -59,7 +59,7 @@ namespace osu.Server.Spectator
             DatabaseUser = Environment.GetEnvironmentVariable("DB_USER") ?? "osuweb";
             DatabasePort = Environment.GetEnvironmentVariable("DB_PORT") ?? "3306";
 
-            SharedInteropDomain = Environment.GetEnvironmentVariable("SHADERD_INTEROP_DOMAIN") ?? "http://localhost:8080";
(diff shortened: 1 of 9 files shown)


bdach:
Should I be renaming this in BSS too? What about osu-queue-score-statistics?

Kinda weirded out that now and here is where this is being brought up.

peppy:
Yep, let's rename them all as time permits.

smoogipoo:
This one has some special logic to deal with exceptions, as well as supporting return values/raw string data (to use Newtonsoft.JSON here). Will need to keep that in mind.

osu.Server.Spectator/AppSettings.cs
```diff
@@ -59,7 +59,7 @@ static AppSettings()
             DatabaseUser = Environment.GetEnvironmentVariable("DB_USER") ?? "osuweb";
             DatabasePort = Environment.GetEnvironmentVariable("DB_PORT") ?? "3306";
 
-            LegacyIODomain = Environment.GetEnvironmentVariable("LEGACY_IO_DOMAIN") ?? "http://localhost:8080";
+            SharedInteropDomain = Environment.GetEnvironmentVariable("SHADERD_INTEROP_DOMAIN") ?? "http://localhost:8080";
```
smoogipoo, line 62:
```suggestion
            SharedInteropDomain = Environment.GetEnvironmentVariable("SHARED_INTEROP_DOMAIN") ?? "http://localhost:8080";
```

smoogipoo (changes requested):

smoogipoo (approved):

PR #273, opened 2025-02-27
Stability improvements for multiplayer room interop

smoogipoo:
As mentioned in https://github.com/ppy/osu/pull/31637#discussion_r1969294493

- Delaying adding users to rooms until after the exclusive lock has been acquired. This would prevent races caused by the single _other_ user leaving the room during the whole process.
- Removing users from rooms if adding fails. This could due to various conditions like the [room having ended](https://github.com/ppy/osu-server-spectator/blob/2be697979fb8b2fec7986f1da78d7c320975bd6d/osu.Server.Spectator/Hubs/Multiplayer/MultiplayerHub.cs#L202-L206).

### DIFF
--- a/osu.Server.Spectator/Hubs/Multiplayer/MultiplayerHub.cs
+++ b/osu.Server.Spectator/Hubs/Multiplayer/MultiplayerHub.cs
@@ -87,8 +87,6 @@ namespace osu.Server.Spectator.Hubs.Multiplayer
                 {
                     try
                     {
-                        await sharedInterop.AddUserToRoomAsync(Context.GetUserId(), roomId, password);
-
                         if (roomUsage.Item == null)
                         {
                             newRoomFetchStarted = true;
@@ -138,8 +136,6 @@ namespace osu.Server.Spectator.Hubs.Multiplayer
                     {
                         try
                         {
-                            await sharedInterop.RemoveUserFromRoomAsync(Context.GetUserId(), roomId);
-
                             if (userUsage.Item != null)
                             {
                                 // the user was joined to the room, so we can run the standard leaveRoom method.
@@ -166,6 +162,15 @@ namespace osu.Server.Spectator.Hubs.Multiplayer
 
                         throw;
                     }
+                }
+
+                try
+                {
+                    await sharedInterop.AddUserToRoomAsync(Context.GetUserId(), roomId, password);
+                }
+                catch (Exception ex)
+                {
+                    Error("Failed to add user to the databased room", ex);
                 }
 
                 var settings = new JsonSerializerSettings
@@ -919,10 +924,17 @@ namespace osu.Server.Spectator.Hubs.Multiplayer
 
         private async Task leaveRoom(MultiplayerClientState state, bool wasKick)
         {
-            await sharedInterop.RemoveUserFromRoomAsync(state.UserId, state.CurrentRoomID);
-
             using (var roomUsage = await getLocalUserRoom(state))
                 await leaveRoom(state, roomUsage, wasKick);
+
+            try
+            {
+                await sharedInterop.RemoveUserFromRoomAsync(state.UserId, state.CurrentRoomID);
+            }
+            catch (Exception ex)
+            {
+                Error("Failed to remove user from the databased room", ex);
+            }
         }
 
         private async Task leaveRoom(MultiplayerClientState state, ItemUsage<ServerMultiplayerRoom> roomUsage, bool wasKick)
--- a/osu.Server.Spectator/Hubs/Multiplayer/MultiplayerHub.cs
+++ b/osu.Server.Spectator/Hubs/Multiplayer/MultiplayerHub.cs
@@ -67,6 +67,8 @@ namespace osu.Server.Spectator.Hubs.Multiplayer
                     throw new InvalidStateException("Can't join a room when restricted.");
             }
 
+            ServerMultiplayerRoom? room = null;
+
             using (var userUsage = await GetOrCreateLocalUserState())
             {
                 if (userUsage.Item != null)
@@ -80,8 +82,6 @@ namespace osu.Server.Spectator.Hubs.Multiplayer
 
                 // track whether this join necessitated starting the process of fetching the room and adding it to the room store.
                 bool newRoomFetchStarted = false;
-
-                ServerMultiplayerRoom? room = null;
 
                 using (var roomUsage = await Rooms.GetForUse(roomId, true))
                 {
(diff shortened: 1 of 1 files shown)


osu.Server.Spectator/Hubs/Multiplayer/MultiplayerHub.cs
```diff
@@ -89,6 +87,8 @@ public async Task<MultiplayerRoom> JoinRoomWithPassword(long roomId, string pass
                 {
                     try
                     {
+                        await sharedInterop.AddUserToRoomAsync(Context.GetUserId(), roomId, password);
```
peppy, line 90: Slightly concerned that this seems to be holding user and room locks over a web request. One of the more recent stability improvements to bancho saw it having a "way out" when database or external requests failed (and all LIO requests are done on a separate thread / queue).

It's fine to go ahead with this until it (potentially) becomes a problem. Just raising concern that this is creating a true cross-component dependency where both components (osu-web shared io octane and spectator-server) being alive to continue functioning. We do have shared IO in its own octane pool, so there's some degree of separation from main web (although that's only octane level, not reverse proxy / nginx).

smoogipoo: The main concern I have is a situation where...

User 1 -> Creates room
User 2 -> Begins joining the room
User 1 -> Leaves room (closes room)
User 2 -> Added to channel via interop
User 2 -> Joining room failed

Without a cross dependency or otherwise lock on rooms, I'm not sure how to best do this. I'm open to suggestions...

That said, I do see another issue here, which is any osu-web errors are potentially going to leave the server in a bad state. I'll block this temporarily to double check that...

> both components (osu-web shared io octane and spectator-server) being alive to continue functioning

The only way I see around this 100% is to completely separate the two. I.e. server-spectator shall do _everything_ related to creating rooms - validating input data, creating chat channel (perhaps that can be via interop since it's not 'core'), and adding rows to the DB.

Which isn't a _bad_ direction imo, but I wouldn't go down that path unless everyone was 100% on board with it.

smoogipoo: To be clear if it makes things any easier to digest/think about, the only thing adding/removing users from rooms entails osu-web side is adding them to the chat channel, and likely a _mild_ form of validation (some of which is duplicated here, like restrictions).

smoogipoo: Given the above, and if we want to reduce dependence between services, how do you feel about this:

```diff
diff --git a/osu.Server.Spectator/Hubs/Multiplayer/MultiplayerHub.cs b/osu.Server.Spectator/Hubs/Multiplayer/MultiplayerHub.cs
index 8535b5f..081a08b 100644
--- a/osu.Server.Spectator/Hubs/Multiplayer/MultiplayerHub.cs
+++ b/osu.Server.Spectator/Hubs/Multiplayer/MultiplayerHub.cs
@@ -87,8 +87,6 @@ namespace osu.Server.Spectator.Hubs.Multiplayer
                 {
                     try
                     {
-                        await sharedInterop.AddUserToRoomAsync(Context.GetUserId(), roomId, password);
-
                         if (roomUsage.Item == null)
                         {
                             newRoomFetchStarted = true;
@@ -138,8 +136,6 @@ namespace osu.Server.Spectator.Hubs.Multiplayer
                     {
                         try
                         {
-                            await sharedInterop.RemoveUserFromRoomAsync(Context.GetUserId(), roomId);
-
                             if (userUsage.Item != null)
                             {
                                 // the user was joined to the room, so we can run the standard leaveRoom method.
@@ -168,6 +164,15 @@ namespace osu.Server.Spectator.Hubs.Multiplayer
                     }
                 }
 
+                try
+                {
+                    await sharedInterop.AddUserToRoomAsync(Context.GetUserId(), roomId, password);
+                }
+                catch (Exception ex)
+                {
+                    Error("Failed to add user to databased room", ex);
+                }
+
                 var settings = new JsonSerializerSettings
                 {
                     // explicitly use Auto here as we are not interested in the top level type being conveyed to the user.
@@ -919,10 +924,17 @@ namespace osu.Server.Spectator.Hubs.Multiplayer
 
         private async Task leaveRoom(MultiplayerClientState state, bool wasKick)
         {
-            await sharedInterop.RemoveUserFromRoomAsync(state.UserId, state.CurrentRoomID);
-
             using (var roomUsage = await getLocalUserRoom(state))
                 await leaveRoom(state, roomUsage, wasKick);
+
+            try
+            {
+                await sharedInterop.RemoveUserFromRoomAsync(state.UserId, state.CurrentRoomID);
+            }
+            catch (Exception ex)
+            {
+                Error("Failed to leave databased room", ex);
+            }
         }
 
         private async Task leaveRoom(MultiplayerClientState state, ItemUsage<ServerMultiplayerRoom> roomUsage, bool wasKick)

```

bdach: I personally would much prefer that last patch rather than what is in this PR right now. While somewhat dicey, I'd say it's much more acceptable levels dicey given the overall low importance of the web-side operation succeeding, compared to holding locks over web requests - especially locks that aren't specific to a single user, like the room lock here.

peppy: Looks good to me too.

osu.Server.Spectator/Hubs/Multiplayer/MultiplayerHub.cs
```diff
@@ -166,6 +164,15 @@ public async Task<MultiplayerRoom> JoinRoomWithPassword(long roomId, string pass
...
                 }
 
+                try
+                {
+                    await sharedInterop.AddUserToRoomAsync(Context.GetUserId(), roomId, password);
+                }
+                catch (Exception ex)
+                {
+                    Error("Failed to add user to the databased room", ex);
+                }
```
bdach, line 174: I'm not sure if I misread or just didn't notice when looking at the patch yesterday, but this block executes while the lock on `userUsage` is still held, and I'm not sure how much of a problem to make of it - I'd prefer if this was outside of any locks, but given it's a lock for a specific user only, I could maybe live with it?

smoogipoo: Makes sense, I've applied this change as well.

peppy (approved):

PR #274, opened 2025-03-03
Retry interop calls only for 5XX responses

smoogipoo:
Fixes the delayed response brought up in https://github.com/ppy/osu/pull/31637#issuecomment-2693174318

### DIFF
--- a/osu.Server.Spectator/Services/SharedInterop.cs
+++ b/osu.Server.Spectator/Services/SharedInterop.cs
@@ -94,8 +94,21 @@ namespace osu.Server.Spectator.Services
             }
             catch (Exception e)
             {
-                if (e is SharedInteropRequestFailedException interopException && !interopException.AllowRetry)
-                    throw;
+                if (e is SharedInteropRequestFailedException interopException)
+                {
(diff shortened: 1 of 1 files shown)


bdach:
Dunno if I agree with this, I'd make the retry conditional on the response code?

Like a 500 or a 504 you can go and retry, but a 429 makes no sense to retry because it's an application-generated response code that isn't likely to be transient.

smoogipoo:
Makes sense. I've force pushed the changes.

osu.Server.Spectator/Services/SharedInterop.cs
```diff
@@ -93,6 +94,9 @@ private async Task<string> runCommand(HttpMethod method, string command, dynamic
             }
             catch (Exception e)
             {
+                if (e is SharedInteropRequestFailedException interopException && !interopException.AllowRetry)
```
peppy, line 97: i probably would have just inlined the status code comparisons here rather than separate `AllowRetry` property, easier to read. seems fine otherwise.

peppy (approved):

bdach (approved):

PR #276, opened 2025-03-04
Invoke interop to remove user from room when kicked

smoogipoo:
Fixes https://github.com/ppy/osu/issues/22279

One method calls another method >_<...

Because this is now in the inner method, migrating the existing implementation would hold onto both user and room locks. This was previously a contentious topic in https://github.com/ppy/osu-server-spectator/pull/273, so I've tried my best to address it here.  

I realise what I've done may be ugly, but I don't know of a better way to do this. The problem is if we use something akin to our `FireAndForget()` extension, the `HttpClient` is disposed because it's a transient service. To get around this, I've added a `runCommandScoped()` which creates a local scope to run the requested command in the background, following [this suggestion](https://stackoverflow.com/a/59878652) on StackOverflow.

Do let me know of any suggestions on how to improve this.

### DIFF
--- a/osu.Server.Spectator/Hubs/Multiplayer/MultiplayerHub.cs
+++ b/osu.Server.Spectator/Hubs/Multiplayer/MultiplayerHub.cs
@@ -165,14 +165,7 @@ namespace osu.Server.Spectator.Hubs.Multiplayer
                 }
             }
 
-            try
-            {
-                await sharedInterop.AddUserToRoomAsync(Context.GetUserId(), roomId, password);
-            }
-            catch (Exception ex)
-            {
-                Error("Failed to add user to the databased room", ex);
-            }
+            await sharedInterop.AddUserToRoomAsync(Context.GetUserId(), roomId, password);
 
             var settings = new JsonSerializerSettings
             {
@@ -978,14 +971,7 @@ namespace osu.Server.Spectator.Hubs.Multiplayer
             else
                 await Clients.Group(GetGroupId(room.RoomID)).UserLeft(user);
 
-            try
-            {
-                await sharedInterop.RemoveUserFromRoomAsync(state.UserId, state.CurrentRoomID);
-            }
-            catch (Exception ex)
-            {
-                Error("Failed to remove user from the databased room", ex);
-            }
+            await sharedInterop.RemoveUserFromRoomAsync(state.UserId, state.CurrentRoomID);
         }
 
         internal Task<ItemUsage<ServerMultiplayerRoom>> GetRoom(long roomId) => Rooms.GetForUse(roomId);
--- a/osu.Server.Spectator/Hubs/Multiplayer/MultiplayerHub.cs
+++ b/osu.Server.Spectator/Hubs/Multiplayer/MultiplayerHub.cs
@@ -926,15 +926,6 @@ namespace osu.Server.Spectator.Hubs.Multiplayer
         {
             using (var roomUsage = await getLocalUserRoom(state))
                 await leaveRoom(state, roomUsage, wasKick);
-
-            try
-            {
-                await sharedInterop.RemoveUserFromRoomAsync(state.UserId, state.CurrentRoomID);
-            }
-            catch (Exception ex)
-            {
-                Error("Failed to remove user from the databased room", ex);
-            }
         }
 
         private async Task leaveRoom(MultiplayerClientState state, ItemUsage<ServerMultiplayerRoom> roomUsage, bool wasKick)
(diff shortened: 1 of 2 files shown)


osu.Server.Spectator/Services/SharedInterop.cs
```diff
@@ -121,6 +130,24 @@ private async Task<string> runCommand(HttpMethod method, string command, dynamic
...
+        private void runCommandScoped(HttpMethod method, string command, dynamic? postObject = null)
+        {
+            _ = Task.Run(async () =>
+            {
+                using (var scope = scopeFactory.CreateScope())
+                {
+                    var scopedInterop = (SharedInterop)scope.ServiceProvider.GetRequiredService<ISharedInterop>();
+                    await scopedInterop.runCommand(method, command, postObject).ConfigureAwait(false);
+                }
+            });
```
bdach, line 148: Two questions:

- I can understand fire-and-forget running this, but is this at least going to get logged if/when it fails? Does this even fail safely as written (i.e. won't it kill the entire service or something because unobserved)i?
- I have to wonder about playing the `IServiceScopeFactory` stuff - does `SharedInterop` even need to have a transient lifetime? I assume that was taken from new BSS, but there it doesn't really matter what lifetime it has, so I just picked transient as the most convenient / least likely to blow up in my face, but in here, it may actually be worth making it singleton lifetime and skip all of the scope games?

osu.Server.Spectator/Services/SharedInterop.cs
```diff
@@ -140,17 +167,22 @@ public async Task<long> CreateRoomAsync(int hostUserId, MultiplayerRoom room)
             })));
         }
 
-        public async Task AddUserToRoomAsync(int userId, long roomId, string password)
+        public Task AddUserToRoomAsync(int userId, long roomId, string password)
```
bdach, line 170: Minor note, no action expected: Feels a smidge weird to have this with an `Async` suffix and returning `Task` if this method is just gonna be a synchronous fire-and-forget now. Like I wouldn't expect an async method to fire-and-forget.

smoogipoo: I think in general "-Async" is for the consumer not the implementation, because removing the `async` keyword indicates no functional difference to the consumer.

osu.Server.Spectator/Services/SharedInterop.cs
smoogipoo: So um... This PR kinda became a bit of a mess. Sorry about that - I can remake if requested.

Looks like making it a singleton is probably the more acceptable path, I've taken that route. In particular, note that we still can't use the `FireAndForget` helper and log the error because `MultiplayerHub.Error()` references `MultiplayerHub.Context` which throws on disposal (typical SignalR transitive hub stuff).

In any case, LIO errors are logged internally with a fix in https://github.com/ppy/osu-server-spectator/pull/276/commits/2972e369d637125aca5fc1727e77f528023bab92 (I considered this before...) I have tested that throwing an exception doesn't kill the server (unobserved generally don't anyway).

osu.Server.Spectator/Services/SharedInterop.cs
```diff
@@ -94,6 +100,8 @@ private async Task<string> runCommand(HttpMethod method, string command, dynamic
             }
             catch (Exception e)
             {
+                bool allowRetry = true;
```
bdach, line 103: @peppy please check that you're ok with this since you [previously objected](https://github.com/ppy/osu-server-spectator/pull/274#discussion_r1977088539) to similar stuff

osu.Server.Spectator/Hubs/Multiplayer/MultiplayerHub.cs
```diff
@@ -167,11 +167,11 @@ public async Task<MultiplayerRoom> JoinRoomWithPassword(long roomId, string pass
 
             try
             {
-                await sharedInterop.AddUserToRoomAsync(Context.GetUserId(), roomId, password);
+                // Run in background so we don't hold locks on user/room states.
+                _ = sharedInterop.AddUserToRoomAsync(Context.GetUserId(), roomId, password);
             }
-            catch (Exception ex)
+            catch
```
bdach, line 173: I'd probably leave an inline comment inside the catch block here (and the other one) that any failures are logged internally to `SharedInterop` because this looks very not good at face value.

osu.Server.Spectator/Services/SharedInterop.cs
peppy: I dunno it reads pretty bad to me but I guess if we have to log this way there's no real avoiding it.

peppy (approved):

PR #278, opened 2025-03-14
Fix cleared star rating when current item is duplicated

smoogipoo:
- [x] Depends on https://github.com/ppy/osu/pull/32372

Notice how after I finish the current playlist item, the SR range display reads `{ 0.0 -> 4.14 }`:

https://github.com/user-attachments/assets/433f8001-f5ea-4802-8cf0-e9ea237ac8cd


peppy:
@smoogipoo I've tagged `osu` with `2025.403.0` which can be used here to unblock (check back later today for package creation).

bdach:
Test failure is irrelevant (already occurs on `master`) and could be fixed by https://github.com/ppy/osu/pull/32666

bdach (approved):

PR #280, opened 2025-03-18
Update packages and switch to using `BeatmapStatusWatcher` for metadata updates

peppy:
More importantly, pulls in https://github.com/ppy/osu/pull/32299.


smoogipoo (approved):
Lgtm

bdach (approved):

PR #282, opened 2025-03-20
Fix intermittent invalid state while removing playlist items

smoogipoo:
Fixes issue pointed out in https://github.com/ppy/osu/pull/32460

I don't like the structure of all of this code, but it is what it is for now...


bdach:
While it's an ordering fix and ordering fixes are generally dodgy I can get behind this without question for the following reasons:

- Fixing this "properly" would require redesigning the client-server API here to be more atomic which is not straightforward
- The fix is low impact and generally shouldn't impact coherency anywhere else
- The original positioning of the hub call is kinda dodgy to begin with (performing network calls when holding a database connection open?)

@peppy heads up that this is moderate priority for deploy to production if the multiplayer rewrite series is to proceed (https://github.com/ppy/osu/pull/32460 and https://github.com/ppy/osu/pull/32458 are both blocked on this)

bdach (approved):

PR #284, opened 2025-04-22
Log room events to new database table

bdach:
Follow-up to https://github.com/ppy/osu-web/pull/12120

This will be powering web pages which will show full room history, eventually addressing https://github.com/ppy/osu-web/issues/10455.


peppy:
I think this is fine to go ahead with.

One thing to note in passing is that bancho is engineered in a way such that log events like this are done so on a background queue, rather than blocking the ongoing flow of multiplayer. It allows bancho to continue to operate even if the database falls over.

For something like this, I'd propose *not* `await`ing on the database execution for said reasons. But we already do this in many other places so I'd see applying a change like this as a later pass. Probably after things fall over and we decide we want that.

PR #285, opened 2025-06-13
Report user counts for tachyon builds

bdach:
In theory this should suffice.


smoogipoo (approved):
Looks fine in principle

PR #288, opened 2025-06-27
Fix server-side replays for beatmaps of version <5 being offset by 24ms from client-side replays

bdach:
Addresses https://github.com/ppy/osu/discussions/33691.

Tested manually full-stack.

### DIFF
--- a/osu.Server.Spectator/Hubs/Spectator/SpectatorHub.cs
+++ b/osu.Server.Spectator/Hubs/Spectator/SpectatorHub.cs
@@ -151,7 +151,7 @@ namespace osu.Server.Spectator.Hubs.Spectator
                     // Score may be null if the BeginPlaySession call failed but the client is still sending frame data.
                     // For now it's safe to drop these frames.
                     // Note that this *intentionally* skips the `endPlaySession()` call at the end of method.
-                    if (score == null || scoreToken == null)
+                    if (score == null || scoreToken == null || usage.Item?.Beatmap == null)
                         return;
 
                     await processScore(usage.Item!);
@@ -173,7 +173,7 @@ namespace osu.Server.Spectator.Hubs.Spectator
 
         private async Task processScore(SpectatorClientState item)
         {
-            Debug.Assert(item.Score != null && item.ScoreToken != null);
+            Debug.Assert(item.Score != null && item.ScoreToken != null && item.Beatmap != null);
 
             Score score = item.Score;
             long scoreToken = item.ScoreToken.Value;
@@ -193,7 +193,7 @@ namespace osu.Server.Spectator.Hubs.Spectator
             // even though in theory the rank could be recomputed after every replay frame.
             score.ScoreInfo.Rank = StandardisedScoreMigrationTools.ComputeRank(score.ScoreInfo);
 
-            await scoreUploader.EnqueueAsync(scoreToken, score, item?.Beatmap ?? new database_beatmap());
+            await scoreUploader.EnqueueAsync(scoreToken, score, item.Beatmap);
             await scoreProcessedSubscriber.RegisterForSingleScoreAsync(Context.ConnectionId, Context.GetUserId(), scoreToken);
         }
 
--- a/osu.Server.Spectator/Hubs/Spectator/SpectatorHub.cs
+++ b/osu.Server.Spectator/Hubs/Spectator/SpectatorHub.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Linq;
 using System.Threading.Tasks;
 using Microsoft.Extensions.Logging;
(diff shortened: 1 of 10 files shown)


osu.Server.Spectator/Hubs/Spectator/SpectatorHub.cs
```diff
@@ -168,8 +161,17 @@ public async Task EndPlaySession(SpectatorState state)
...
 
-        private async Task processScore(Score score, long scoreToken)
+        private async Task processScore(SpectatorClientState? item)
         {
+            Score? score = item?.Score;
+            long? scoreToken = item?.ScoreToken;
+
+            // Score may be null if the BeginPlaySession call failed but the client is still sending frame data.
+            // For now it's safe to drop these frames.
+            // Note that this *intentionally* skips the `endPlaySession()` call at the end of method.
```
smoogipoo, line 171: This comment is no longer accurate. Have you checked the effects of this?

osu.Server.Spectator/Hubs/Spectator/SpectatorHub.cs
```diff
@@ -168,8 +171,13 @@ public async Task EndPlaySession(SpectatorState state)
...
         }
 
-        private async Task processScore(Score score, long scoreToken)
+        private async Task processScore(SpectatorClientState item)
         {
+            Debug.Assert(item.Score != null && item.ScoreToken != null);
+
+            Score score = item.Score;
+            long scoreToken = item.ScoreToken.Value;
+
```
smoogipoo, line 180: Was this change necessary? Just trying to reduce complexity for reviewers' eyes (mine).

```diff
diff --git a/osu.Server.Spectator/Hubs/Spectator/SpectatorHub.cs b/osu.Server.Spectator/Hubs/Spectator/SpectatorHub.cs
index e43d9eb..83b2e69 100644
--- a/osu.Server.Spectator/Hubs/Spectator/SpectatorHub.cs
+++ b/osu.Server.Spectator/Hubs/Spectator/SpectatorHub.cs
@@ -3,7 +3,6 @@
 
 using System;
 using System.Collections.Generic;
-using System.Diagnostics;
 using System.Linq;
 using System.Threading.Tasks;
 using Microsoft.Extensions.Logging;
@@ -147,14 +146,15 @@ namespace osu.Server.Spectator.Hubs.Spectator
                 {
                     Score? score = usage.Item?.Score;
                     long? scoreToken = usage.Item?.ScoreToken;
+                    database_beatmap? beatmap = usage.Item?.Beatmap;
 
                     // Score may be null if the BeginPlaySession call failed but the client is still sending frame data.
                     // For now it's safe to drop these frames.
                     // Note that this *intentionally* skips the `endPlaySession()` call at the end of method.
-                    if (score == null || scoreToken == null)
+                    if (score == null || scoreToken == null || beatmap == null)
                         return;
 
-                    await processScore(usage.Item!);
+                    await processScore(score, scoreToken.Value, beatmap);
                 }
                 finally
                 {
@@ -171,13 +171,8 @@ namespace osu.Server.Spectator.Hubs.Spectator
             await endPlaySession(Context.GetUserId(), state);
         }
 
-        private async Task processScore(SpectatorClientState item)
+        private async Task processScore(Score score, long scoreToken, database_beatmap beatmap)
         {
-            Debug.Assert(item.Score != null && item.ScoreToken != null);
-
-            Score score = item.Score;
-            long scoreToken = item.ScoreToken.Value;
-
             // Do nothing with scores on unranked beatmaps.
             var status = score.ScoreInfo.BeatmapInfo!.Status;
             if (status < min_beatmap_status_for_replays || status > max_beatmap_status_for_replays)
@@ -193,7 +188,7 @@ namespace osu.Server.Spectator.Hubs.Spectator
             // even though in theory the rank could be recomputed after every replay frame.
             score.ScoreInfo.Rank = StandardisedScoreMigrationTools.ComputeRank(score.ScoreInfo);
 
-            await scoreUploader.EnqueueAsync(scoreToken, score, item?.Beatmap ?? new database_beatmap());
+            await scoreUploader.EnqueueAsync(scoreToken, score, beatmap);
             await scoreProcessedSubscriber.RegisterForSingleScoreAsync(Context.ConnectionId, Context.GetUserId(), scoreToken);
         }
 

```

Is there a case where `item.Beatmap` can be null and we should use `new database_beatmap()`? I sure hope not.

smoogipoo (changes requested):

osu.Server.Spectator/Hubs/Spectator/SpectatorHub.cs
bdach: I mostly dislike passing three properties from an object into a method if they're already enclosed in one object that can be passed directly. but if you see it as an issue then I can change this.

the null beatmap fallback is just mostly because something needs to be done about it being maybe-null from NRT annotations. I can change it into an assertion in `processScore()` if it helps.

smoogipoo: I'd be okay with that, but `EndPlaySession()` has to also check for null otherwise the asserts are illogical to me.

smoogipoo (approved):

peppy (approved):

Issue #289, opened 2025-07-02
Stored replays are missing `ClientVersion` lazer value


smoogipoo:
Lazer encodes a [small blob](https://github.com/ppy/osu/blob/master/osu.Game/Scoring/Legacy/LegacyReplaySoloScoreInfo.cs) containing some extra data. This is currently missing the `ClientVersion` value as it's not being set by osu-server-spectator:

<img width="508" alt="Image" src="https://github.com/user-attachments/assets/187dee3c-145a-429f-848d-e2b05db7a7e7" />

Example score: https://osu.ppy.sh/scores/3497894741

bdach:
can probably be filled out pretty easily somewhere around

https://github.com/ppy/osu-server-spectator/blob/bab7c8c8db1803bbd19decb418dae1672b6986b6/osu.Server.Spectator/Hubs/ScoreUploader.cs#L106-L107

using `build_id` on the score; only possible concern would be the extra db query overhead (could just do a permanent memory cache keyed by `build_id`)

PR #290, opened 2025-07-02
Write client version to server-side replays

bdach:
Closes https://github.com/ppy/osu-server-spectator/issues/289.

As far as I'm aware the data in the `version` column in `osu_builds` should match [the client-side implementation](https://github.com/ppy/osu/blob/b1435d35e56eed08a57d1909fa0b16e67bd9c2a2/osu.Game/OsuGameBase.cs#L121-L139) but it may be worth double checking.

### DIFF
--- a/osu.Server.Spectator/Hubs/Metadata/MetadataHub.cs
+++ b/osu.Server.Spectator/Hubs/Metadata/MetadataHub.cs
@@ -180,7 +180,7 @@ namespace osu.Server.Spectator.Hubs.Metadata
 
             using var db = databaseFactory.GetInstance();
 
-            MultiplayerRoomStats stats = (await cache.GetOrCreateAsync<MultiplayerRoomStats>(id.ToString(), e =>
+            MultiplayerRoomStats stats = (await cache.GetOrCreateAsync<MultiplayerRoomStats>($@"{nameof(MultiplayerRoomStats)}#{id}", e =>
             {
                 e.SlidingExpiration = TimeSpan.FromHours(24);
                 return Task.FromResult(new MultiplayerRoomStats { RoomID = id });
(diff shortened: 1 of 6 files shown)


osu.Server.Spectator/Hubs/Metadata/MetadataHub.cs
```diff
@@ -180,7 +180,7 @@ public async Task<MultiplayerPlaylistItemStats[]> BeginWatchingMultiplayerRoom(l
 
             using var db = databaseFactory.GetInstance();
 
-            MultiplayerRoomStats stats = (await cache.GetOrCreateAsync<MultiplayerRoomStats>(id.ToString(), e =>
+            MultiplayerRoomStats stats = (await cache.GetOrCreateAsync<MultiplayerRoomStats>($@"{nameof(MultiplayerRoomStats)}#{id}", e =>
```
bdach, line 183: Reason for this weird change is that I wanted to use a memory cache for the build lookups to reduce redundancy and https://github.com/ppy/osu-server-spectator/blob/bab7c8c8db1803bbd19decb418dae1672b6986b6/osu.Server.Spectator/Startup.cs#L60 registers the cache instance as a singleton:

Image

so cache keys have to be globally unique lest some Really Dumb Stuff Happens™

osu.Server.Spectator/Hubs/ScoreUploader.cs
```diff
@@ -103,6 +107,17 @@ private async Task readLoop()
                         if (!dbScore.passed)
                             continue;
 
+                        if (dbScore.build_id != null)
```
peppy, line 110: I would have just joined on the existing `db.GetScoreFromTokenAsync` query, but I guess this will work.

peppy (approved):

PR #291, opened 2025-07-24
Expire remaining playlist items on ending match rather than deleting them

bdach:
Should close https://github.com/ppy/osu-web/issues/12301.

As per @smoogipoo this was [never strictly necessary and only done for space saving](https://discord.com/channels/90072389919997952/1327149041511043134/1397166770627936277).

The other direction would be to delete rows from all other related tables but I think not deleting is better because of history preservation, and I'm not sure the space gains from this special case will be relevant.


peppy (approved):

peppy:
I'll deploy this alongside the next major lazer release, unless something comes up earlier.

PR #292, opened 2025-07-30
Fix game completed / aborted events doubling up on each other

bdach:
Should address https://discord.com/channels/188630481301012481/188630616286167041/1398583488147619852.

I'm going to admit, I'm not *that* confident in this change, the state flow in `MultiplayerHub` is quite complex and `updateRoomStateAsRequired()` has many callers - but maybe it works...? At least the test coverage added here says that it might...

A side effect of this is that e.g. if in a game every player quits out individually, the game gets marked as aborted and not completed. Which I think makes sense.


peppy (approved):
Looks logical to me

PR #300, opened 2025-08-26
Make `MatchTypeImplementation` async

smoogipoo:


smoogipoo:
> Not sure I understand this.

For example the HeadToHead implementation, joining users fire `Hub.NotifyMatchUserStateChanged` without awaiting.

> Is there a specific reason async is required in some future PR?

Messages are supposed to fire in sequence, hence `await`. For example matchmaking will fire both state change and countdown events when the stage changes.

But even more so than that, we're dealing with [_async-coloured_](https://journal.stuffwithstuff.com/2015/02/01/what-color-is-your-function/) methods here. Not only the hub context but database operations are async, and we cannot simply jump between async (hub) -> sync (MTI) -> async (database) easily. Furthermore, there are no variations of some sync SignalR methods like `Groups.AddToGroupAsync()`, so path to make everything sync doesn't make sense either.

Basically, there is nothing that makes sense other than async.

PR #301, opened 2025-08-26
Refactor and rename multiplayer playlist (ex-`MultiplayerQueue`)

smoogipoo:
- [x] Depends on https://github.com/ppy/osu-server-spectator/pull/300

Split from https://github.com/ppy/osu-server-spectator/pull/298

- `MatchTypeImplementation` made into an interface (`IMatchController`).
- `MultiplayerQueue` merged into/as class `StandardMatchController`.
  - New base class of `HeadToHeadMatchController` and `TeamVersusMatchController`.
- Renamed `Room.MatchTypeImplementation` -> `Room.Controller`.

Matchmaking will implement its own controller, and some of the interface methods (e.g. `AddPlaylistItem()`) as no-ops.


osu.Server.Spectator/Hubs/Multiplayer/ServerMultiplayerRoom.cs
```diff
@@ -23,24 +23,28 @@ public MatchTypeImplementation MatchTypeImplementation
             private set => matchTypeImplementation = value;
         }
 
-        public readonly MultiplayerQueue Queue;
+        public IPlaylistImplementation PlaylistImplementation => MatchTypeImplementation.Playlist;
```
peppy, line 26: Can we call this just `Playlist`? What is the meaning behind "Implementation" here?

smoogipoo: `Playlist` would hide the base class' member, and I don't know the best way to do this:

https://github.com/ppy/osu-server-spectator/blob/18bd8d164d364bb90fdb6b10328b0168218f64f3/osu.Server.Spectator.Tests/Multiplayer/MatchTypeRoomEventHookTests.cs#L22-L32

osu.Server.Spectator/Hubs/Multiplayer/MatchTypeImplementation.cs
```diff
@@ -18,9 +20,9 @@ protected MatchTypeImplementation(ServerMultiplayerRoom room, IMultiplayerHubCon
             Hub = hub;
         }
 
-        public virtual Task Initialise()
+        public virtual async Task Initialise()
         {
-            return Task.CompletedTask;
+            await Playlist.Initialise();
```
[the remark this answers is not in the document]
smoogipoo: Right now I just need it so that matchmaking can set its mostly-no-op implementation.

This can definitely be done better. One idea I had was to merge `IPlaylistImplementation` and `MatchTypeImplementation` and let it be implemented by a `MultiplayerMatchType` (which `HeadToHead` and `TeamVersus` would subclass).

If you'd like to see that here, then I can try it out.

Also, I really don't like that this is a `class` and not an `interface`. There's really no need for it to be one.

osu.Server.Spectator/Hubs/Multiplayer/ServerMultiplayerRoom.cs
peppy: It would be nice is `ServerMultiplayerRoom` could just implement `IPlaylistImplementation` and delegate to the implementation internally. But having `Initialize` in that interface feels really bad..

smoogipoo: I've gone ahead and implemented the idea in https://github.com/ppy/osu-server-spectator/pull/301#discussion_r2300265527 in the last two commits. Please see if you agree with this structure more.

smoogipoo:
Lgtm 👍

PR #309, opened 2025-09-05
Allow matchmaking rooms of any size

smoogipoo:
Turns out I didn't have to hack this one as much as I'd thought. `MATCHMAKING_ROOM_SIZE` environment variable set to special value "0" will allow rooms of any size.

Additionally, I've added environment variables to control update rates. Suggested settings for staging:

```
MATCHMAKING_ROOM_SIZE = 0
MATCHMAKING_QUEUE_UPDATE_RATE = 10
MATCHMAKING_ALLOW_SKIP = true
```

https://github.com/user-attachments/assets/f09474d2-1a3a-4637-b1ce-eceb89eb3d16


osu.Server.Spectator/Hubs/Multiplayer/Matchmaking/MatchmakingMatchController.cs
```diff
@@ -135,7 +135,7 @@ public async Task HandleUserJoined(MultiplayerRoomUser user)
             switch (state.Stage)
             {
                 case MatchmakingStage.WaitingForClientsJoin:
-                    if (++joinedUserCount == room_size)
+                    if (++joinedUserCount >= room_size)
```
smoogipoo, line 138: I think this should be safe...

The original change I was going to go with was storing the room settings to an `IMemoryCache` object, and then pulling those settings in this class. This is because otherwise I have no way to pass the MM room size between classes due to the database round-trip.

smoogipoo: That is true, though this process should happen fairly quick (~1 second) so I don't know if it's a problem at least for testing.

This code only starts triggering after every player has accepted the invitation, so they've already signaled their intent to join and the only thing remaining is the client to invoke `JoinRoom()`.

peppy: So they can still join during warmup?

smoogipoo: With room size zero, yes.

peppy (approved):

PR #312, opened 2025-09-07
Add `MATCHMAKING_MINIMUM_TIME_IN_QUEUE`

smoogipoo:
This was brought up as an alternative to `MATCHMAKING_QUEUE_UPDATE_RATE`. I suggest not changing the value at all, even for production.


smoogipoo:
Not sure about the implementation right now, may/will revise.

PR #331, opened 2025-10-17
Refactor quick play environment variables for consistency

smoogipoo:
I used two different conventions here - first specifying the fallback values at the point of parsing (which is how we usually do it), and the better (imo) version which is specifying the static values as fallbacks.

This updates them all to use the second version for readability purposes.

### DIFF
--- a/osu.Server.Spectator/AppSettings.cs
+++ b/osu.Server.Spectator/AppSettings.cs
@@ -8,38 +8,38 @@ namespace osu.Server.Spectator
     public static class AppSettings
     {
         public static bool SaveReplays { get; set; }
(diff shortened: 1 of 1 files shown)


smoogipoo:
Just to make safe incremental changes and not potentially break things, I guess... The others use a different format still, which is bools parsed as `== "1"`, which is not how `bool.TryParse()` operates.

If @peppy also agrees with these changes, I can change all the others. Note that we'll have to update some env vars though (e.g. probably `SAVE_REPLAYS = 1` -> `SAVE_REPLAYS = true`).

peppy:
Yeah let's standardise others to use this style.

peppy (approved):

PR #334, opened 2025-10-20
Record queue wait times per-pool

smoogipoo:
@peppy I assume `matchmaking_pools.name` is set appropriately for each pool?


peppy:
Yes

image

PR #335, opened 2025-10-20
Adjust quick play stage times

smoogipoo:
Round warmup: 10s -> 20s

- Gives more time for players to participate in the room post-round.

Selection: 20s -> 60s

- Gives more time for players to select beatmaps.
- Playing a bit on the safe side with this, which might be a little excessive. Better would be to both give a way for players to select a random beatmap _and_ to ffwd the timer when all players have a selection, but I'm not sure what it would take to get that to work right now.

Beatmap finalised: 5s -> 7s

- With the previous value the beatmap would start being downloaded slightly before the animation completes.


peppy:
Will redeploy with this tomorrow.

PR #337, opened 2025-10-21
Adjust quick play statsd counters

smoogipoo:
Hopefully the last time I'll do anything with these counters... Sorry.

- Removed the groups counters because they're not really useful.
- Tagged the users counter.
- Adjusted to more canonical names following conventions in `ScoreProcessedSubscriber`.


osu.Server.Spectator/Hubs/Multiplayer/Matchmaking/Queue/MatchmakingQueue.cs
```diff
@@ -61,6 +61,18 @@ public MatchmakingQueue(matchmaking_pool pool)
...
         }
 
+        /// <summary>
+        /// Retrieves the count of users in this queue.
+        /// </summary>
+        public int Count
+        {
+            get
+            {
+                lock (queueLock)
```
peppy: arguable that you need to lock for count, but sure.

peppy (approved):

PR #340, opened 2025-10-24
Reduce round start and user pick stage times

smoogipoo:
These both feel a little bit too long now. They're still increased from their initial values (start: 10 -> 15 and pick: 20 -> 40).


peppy:
Will be deployed monday if that's okay.

PR #349, opened 2025-10-29
Fix room not cleared on disconnections

smoogipoo:
The only path that was clearing the room was via `LeaveRoom` (the client invocation). We need to do it in all cases.

### DIFF
--- a/osu.Server.Spectator/Hubs/Multiplayer/MultiplayerHub.cs
+++ b/osu.Server.Spectator/Hubs/Multiplayer/MultiplayerHub.cs
@@ -339,14 +339,7 @@ namespace osu.Server.Spectator.Hubs.Multiplayer
                         if (targetUserUsage.Item.CurrentRoomID == null)
                             throw new InvalidOperationException();
 
-                        try
-                        {
-                            await leaveRoom(targetUserUsage.Item, roomUsage, true);
-                        }
-                        finally
-                        {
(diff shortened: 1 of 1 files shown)


bdach (commented):
It's style points, but with `ClearRoom()` now moved to the innermost `leaveRoom()` that everything else calls, you can get rid of another call:

```diff
diff --git a/osu.Server.Spectator/Hubs/Multiplayer/MultiplayerHub.cs b/osu.Server.Spectator/Hubs/Multiplayer/MultiplayerHub.cs
index cc981f6d52f..9db92119a6e 100644
--- a/osu.Server.Spectator/Hubs/Multiplayer/MultiplayerHub.cs
+++ b/osu.Server.Spectator/Hubs/Multiplayer/MultiplayerHub.cs
@@ -339,14 +339,7 @@ namespace osu.Server.Spectator.Hubs.Multiplayer
                         if (targetUserUsage.Item.CurrentRoomID == null)
                             throw new InvalidOperationException();
 
-                        try
-                        {
-                            await leaveRoom(targetUserUsage.Item, roomUsage, true);
-                        }
-                        finally
-                        {
-                            targetUserUsage.Item.ClearRoom();
-                        }
+                        await leaveRoom(targetUserUsage.Item, roomUsage, true);
                     }
                 }
             }

```

because `leaveRoom()` will clear the room anyway inside of it.

PR #352, opened 2025-10-30
Improve randomised beatmap selection algorithm

smoogipoo:
The old algorithm didn't provide provide a nice distribution of beatmaps. This is because it cut off the available pool of beatmaps for the easy/difficult set based on the standard deviation.

I ran both algorithms across a selection of 1000 times over a dataset of 1000 fake beatmaps, forming 4 datasets:
1. Negative bias: player rating skews negatively
2. Positive bias: player rating skews positively
3. Outer bias: player rating is skewed away from the average
4. Inner bias: player rating is skewed towards the average

In the following tests, the x-axis "beatmap" doubles as the rating of that particular beatmap, where 500 is the "centre" rating.

| case | old | new |
| --- | --- | --- |
| negative bias | image | image |
| positive bias | image | image |
| outer bias | image | image |
| inner bias | image | image |

Of importance:
- negative biased shows a healthy selection of difficult beatmaps
- positive biased shows a healthy selection of easy beatmaps
- inner bias tightens towards the centre but still shows a healthy selection of difficult and easy beatmaps

Generated with the following test: https://gist.github.com/smoogipoo/63ae3c6d78382caa88a735eadb6884c7

### DIFF
--- a/osu.Server.Spectator/Hubs/Multiplayer/Matchmaking/Queue/MatchmakingBeatmapSelector.cs
+++ b/osu.Server.Spectator/Hubs/Multiplayer/Matchmaking/Queue/MatchmakingBeatmapSelector.cs
@@ -41,22 +41,20 @@ namespace osu.Server.Spectator.Hubs.Multiplayer.Matchmaking.Queue
         /// <param name="ratings">The lobby user ratings.</param>
         public matchmaking_pool_beatmap[] GetAppropriateBeatmaps(EloRating[] ratings)
         {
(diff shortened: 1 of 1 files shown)


bdach (approved):
In that case let's ✨  𝓂𝑜𝓋𝑒 𝒻𝒶𝓈𝓉 𝒶𝓃𝒹 𝒷𝓇𝑒𝒶𝓀 𝓉𝒽𝒾𝓃𝑔𝓈 ✨

PR #353, opened 2025-10-30
Initialise a default beatmap rating

smoogipoo:
- Related to https://github.com/ppy/osu-web/pull/12505

I don't want us to have to keep updating the table in post as we add/modify beatmaps.


bdach (approved):
Not sure what sort of deploy timing you want on this, so will hold with merge

PR #357, opened 2025-10-31
Fix quick play rooms retaining a host

smoogipoo:
Fixes https://github.com/ppy/osu/issues/35517

### DIFF
--- a/osu.Server.Spectator/Hubs/Multiplayer/MultiplayerHub.cs
+++ b/osu.Server.Spectator/Hubs/Multiplayer/MultiplayerHub.cs
@@ -133,7 +133,7 @@ namespace osu.Server.Spectator.Hubs.Multiplayer
                                     throw new InvalidPasswordException();
                             }
 
-                            if (isNewRoom)
+                            if (isNewRoom && room.Settings.MatchType != MatchType.Matchmaking)
                                 room.Host = roomUser;
 
                             userUsage.Item.SetRoom(roomId);


bdach (commented):
Would like test coverage for this, even if it's something along the lines of

```diff
diff --git a/osu.Server.Spectator.Tests/Matchmaking/MatchmakingMatchControllerTests.cs b/osu.Server.Spectator.Tests/Matchmaking/MatchmakingMatchControllerTests.cs
index cab7dd37b83..8c48cd04d81 100644
--- a/osu.Server.Spectator.Tests/Matchmaking/MatchmakingMatchControllerTests.cs
+++ b/osu.Server.Spectator.Tests/Matchmaking/MatchmakingMatchControllerTests.cs
@@ -67,6 +67,7 @@ namespace osu.Server.Spectator.Tests.Matchmaking
 
             // Join the first user.
             await Hub.JoinRoom(ROOM_ID);
+            Assert.Null(Rooms.GetEntityUnsafe(ROOM_ID)?.Host);
 
             // Check that the room is waiting for users.
             await verifyStage(MatchmakingStage.WaitingForClientsJoin);
@@ -78,6 +79,7 @@ namespace osu.Server.Spectator.Tests.Matchmaking
             // Join the second user.
             SetUserContext(ContextUser2);
             await Hub.JoinRoom(ROOM_ID);
+            Assert.Null(Rooms.GetEntityUnsafe(ROOM_ID)?.Host);
 
             for (int i = 1; i <= 2; I++)
             {

```

Also I find the following points a bit dodgy:

- the host is specified web-side in the shared interop call:

  https://github.com/ppy/osu-server-spectator/blob/f04afdba48fc43e8e4b9d48e4c39cdd8be720670/osu.Server.Spectator/Services/SharedInterop.cs#L145-L151
  
  but then spectator server makes no real effort to validate anything against that. That said, I can see it being annoying, so I wouldn't really expect any real action here.
- The `TransferHost()` operation doesn't have an explicit protection against being called in quick play rooms (although with this change, it will *implicitly* fail because the room will have null host, so no user will be able to pass the check as to whether they are host).

So I guess this change is *functionally* fine but I find it a bit lacking. Your call as to how far defensively speaking you want to carry this.

PR #358, opened 2025-11-03
Disable JSON protocol

smoogipoo:
See internal conversation starting with https://discord.com/channels/90072389919997952/1327149041511043134/1434563875499409579

This is explicitly listed as a supported action [here](https://learn.microsoft.com/en-us/aspnet/core/signalr/configuration?view=aspnetcore-9.0&tabs=dotnet#configure-server-options):

> Protocols supported by this hub. By default, all protocols registered on the server are allowed. Protocols can be removed from this list to disable specific protocols for individual hubs.

### DIFF
--- a/osu.Server.Spectator/Startup.cs
+++ b/osu.Server.Spectator/Startup.cs
@@ -27,6 +27,10 @@ namespace osu.Server.Spectator
         {
             services.AddSignalR(options =>
                     {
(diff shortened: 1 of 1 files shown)


bdach (approved):
This drops any client that wants to use json rather rudely, but this is better than the alternative which is causing a whole bunch of havoc server-side, sooooooo...

PR #375, opened 2025-11-20
Further iterate on beatmap selection algorithm

smoogipoo:
Follows on from https://github.com/ppy/osu-server-spectator/pull/352

Sorry, this one is also going to be a staring at graphs game and which one looks nicer.

The point here is multi-fold:
- As more beatmaps are added (and eventually filtering the [entire ranked section](https://discord.com/channels/188630481301012481/1440912440224120882/1440920448358682795)), I grow increasingly unconfident in the existing algorithm and desire something simpler - hence using a standard gaussian distribution.
- To not average out the ratings.

Imagining two players at 1200 and 1800 rating, the current algorithm will average them to around 1500. The new algorithm will instead define two peaks at each player's rating and merge them.

| old | new |
| --- | --- |
| Screenshot 2025-11-20 at 16 29 56 | image |

Data generation: https://gist.github.com/smoogipoo/64e8541f9662ed418a7093289764532a

### DIFF
--- a/osu.Server.Spectator/Hubs/Multiplayer/Matchmaking/Queue/MatchmakingBeatmapSelector.cs
+++ b/osu.Server.Spectator/Hubs/Multiplayer/Matchmaking/Queue/MatchmakingBeatmapSelector.cs
@@ -48,13 +48,17 @@ namespace osu.Server.Spectator.Hubs.Multiplayer.Matchmaking.Queue
         public matchmaking_pool_beatmap[] GetAppropriateBeatmaps(EloRating[] ratings)
         {
             return ratings.SelectMany(r =>
-                              beatmaps.OrderByDescending(b =>
+                          {
+                              // 80 is a safe value that is the lower limit of EloSystem.
+                              double ratingSig = Math.Max(80, r.Sig);
+                              return beatmaps.OrderByDescending(b =>
                               {
                                   double beatmapRating = b.rating ?? 1500;
-                                  // The clamp acts to both prevent division-by-zero and to give all beatmaps a chance.
-                                  double weight = Math.Clamp(Math.Exp(-Math.Pow(beatmapRating - r.Mu, 2) / (2 * r.Sig * r.Sig)), 0.1, 1);
+                                  // The clamp attempts to ensure all beatmaps are given some chance of being selected.
+                                  double weight = Math.Clamp(Math.Exp(-Math.Pow(beatmapRating - r.Mu, 2) / (2 * ratingSig * ratingSig)), 0.1, 1);
                                   return Math.Pow(Random.Shared.NextDouble(), 1.0 / weight);
-                              }).Take(PoolSize))
+                              }).Take(PoolSize);
+                          })
                           .OrderBy(_ => Random.Shared.NextDouble())
                           .Take(PoolSize)
                           .ToArray();


osu.Server.Spectator/Hubs/Multiplayer/Matchmaking/Queue/MatchmakingBeatmapSelector.cs
```diff
@@ -48,75 +47,17 @@ public static async Task<MatchmakingBeatmapSelector> Initialise(matchmaking_pool
...
-                windowSize++;
-            }
-
-            return result.Take(countToAdd).ToArray();
+            return ratings.SelectMany(r =>
+                              beatmaps.OrderByDescending(b =>
+                              {
+                                  double beatmapRating = b.rating ?? 1500;
+                                  // The clamp acts to both prevent division-by-zero and to give all beatmaps a chance.
+                                  double weight = Math.Clamp(Math.Exp(-Math.Pow(beatmapRating - r.Mu, 2) / (2 * r.Sig * r.Sig)), 0.1, 1);
```
bdach: How often is `r.Sig` going to be 0? I assume that's a variance-type quantity? Is it going to affect only brand new players that never touched quick play before? Or is it something that is going to be e.g. reset every time the pool changes?

The reason I ask this, and maybe I'm overthinking this, but assuming `r.Sig` of zero:

- `2 * r.Sig * r.Sig` is zero
- `-Math.Pow(beatmapRating - r.Mu, 2)` is always negative
- therefore the entire expression under the exp is `-∞`
- therefore `Exp(-∞)`, taking the limit interpretation, is 0
- therefore `weight` is `0.1` due to the clamp
- therefore the random variable sampled below in this case will be raised to the tenth power
- therefore (and this is where I'm no longer certain because I *suck* at statistics) this makes the sampling no longer uniform. Instead it is *maybe* a [beta distribution](https://www.wolframalpha.com/input?i=beta+distribution+alpha%3D1%2F10+beta%3D1+)?

Why does that matter? Because that distribution is *skewed* and may no longer give all beatmaps in the pool an equal chance to be sampled. Although at the end there, there is also sorting by that maybe-beta-distributed random variable factored into it which is another layer of statistics slash probability that I'm not even going to attempt going into because I'm too stupid and would probably have to just run simulations instead.

smoogipoo: `r.Sig` is never zero. It starts at [350](https://github.com/ppy/osu-server-spectator/blob/d0b4f54ee86d4c2fa3ebb0e7dc582d91ada47330/osu.Server.Spectator/Hubs/Multiplayer/Matchmaking/Elo/EloRating.cs#L18-L19) and decreases to [80](https://github.com/ppy/osu-server-spectator/blob/d0b4f54ee86d4c2fa3ebb0e7dc582d91ada47330/osu.Server.Spectator/Hubs/Multiplayer/Matchmaking/Elo/EloSystem.cs#L28C23-L28C31) over time. But it's a good idea to `Math.Max()` just in case.

bdach: The "division-by-zero" mention in the comment was a bit of a misdirect then...

smoogipoo: Maybe not div-by-0 in the error sense directly, but div by an extremely-low-near-zero number. Have also clarified (removed) this mention.

PR #379, opened 2025-11-25
Implement featured artist pool for quick play

smoogipoo:
Doesn't need to be merged or deployed immediately - it's intended for the 3rd week of the upcoming "release" of quick play, which intends to use the full ranked beatmap pool for quick play.

This removes writes to `matchmaking_pool_beatmaps.selection_count`. I think we should extract this data from `matchmaking_room_events` instead, as this isn't going to be supported with the ranked beatmap pool anyway. There's probably not enough data in there at the moment to do this exactly...

If there are no rows in `matchmaking_pool_beatmaps` for the pool, it filters from `osu_beatmaps`:
- Featured artist
- Ranked status
- Within 1 min and 4 minutes in length
- Of the correct ruleset and variant (in the case of mania)

### DIFF
--- a/osu.Server.Spectator/Database/DatabaseAccess.cs
+++ b/osu.Server.Spectator/Database/DatabaseAccess.cs
@@ -677,12 +677,19 @@ namespace osu.Server.Spectator.Database
             string variantString = string.Empty;
 
             if (rulesetId == 3)
-                variantString = "AND diff_size = @Variant";
+                variantString = "AND b.diff_size = @Variant";
 
-            return (await connection.QueryAsync<database_beatmap>("SELECT beatmap_id, checksum, difficultyrating FROM `osu_beatmaps` "
-                                                                  + "WHERE playmode = @RulesetId "
-                                                                  + "AND approved BETWEEN 1 AND 2 "
-                                                                  + "AND hit_length BETWEEN 60 AND 240 "
+            // - From the featured artist listing
+            // - Non-converted beatmaps
+            // - Ranked status
+            // - Between 1 and 4 minutes in length
+            // - With the correct keymode (if mania)
+            return (await connection.QueryAsync<database_beatmap>("SELECT b.beatmap_id, b.checksum, b.difficultyrating FROM `osu_beatmaps` b "
+                                                                  + "JOIN `osu_beatmapsets` s ON s.beatmapset_id = b.beatmapset_id "
+                                                                  + "WHERE s.track_id IS NOT NULL "
+                                                                  + "AND b.playmode = @RulesetId "
+                                                                  + "AND b.approved BETWEEN 1 AND 2 "
+                                                                  + "AND b.hit_length BETWEEN 60 AND 240 "
                                                                   + variantString,
                 new
                 {


osu.Server.Spectator/Hubs/Multiplayer/Matchmaking/Queue/MatchmakingQueueBackgroundService.cs
```diff
@@ -231,6 +235,17 @@ private async Task refreshQueues()
             lastQueueRefreshTime = DateTimeOffset.Now;
         }
 
+        private Task refreshPools()
+        {
+            if (DateTimeOffset.Now - lastPoolRefreshTime < TimeSpan.FromMinutes(5))
+                return Task.CompletedTask;
```
bdach: Not sure how to feel about this refresh period. I'd increase it significantly, go to like every-hour range or so, probably. Main concern would be the amount of db scraping this is going to do by retrieving all beatmaps.

No idea how the query itself performs in a production setting. At least the select is column-limited to something like (back-of-napkin) 23 bytes per beatmap so maybe it won't return *egregious* amounts of data (but may well be in megabytes range still).

osu.Server.Spectator/Database/DatabaseAccess.cs
```diff
@@ -659,16 +659,25 @@ public async Task<matchmaking_pool_beatmap[]> GetMatchmakingPoolBeatmapsAsync(ui
             })).ToArray();
         }
 
-        public async Task IncrementMatchmakingSelectionCount(matchmaking_pool_beatmap[] beatmaps)
```
peppy: Bit weird to remove this but leave the column around. Originally we discussed having one extra column `present_count` so we could get an idea of the ratio of presents vs user selects (where `selection_count` might become a decimal denoting how many users from the full room chose it, or where `present_count` is incremented by `room_size` each present to get a fair statistic).

osu.Server.Spectator/Hubs/Multiplayer/Matchmaking/Queue/MatchmakingBeatmapSelector.cs
```diff
@@ -28,17 +34,24 @@ public MatchmakingBeatmapSelector(matchmaking_pool_beatmap[] beatmaps)
...
+                // If there are no beatmaps specified in the pool, use all global ranked beatmaps.
+                if (beatmaps.Length == 0)
+                {
+                    database_beatmap[] globalBeatmaps = await db.GetMatchmakingGlobalPoolBeatmapsAsync(pool.ruleset_id, pool.variant_id);
+                    beatmaps = globalBeatmaps.Select(b => new matchmaking_pool_beatmap
+                    {
+                        beatmap_id = b.beatmap_id,
+                        checksum = b.checksum,
+                        difficultyrating = b.difficultyrating
+                    }).ToArray();
```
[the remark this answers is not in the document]
smoogipoo: I adjusted the types here to `uint`s to match the DB. That means I can't use -1.

I've set `pool_id` to the incoming pool ID, but I think 0 is already a "tombstone value" for `id`? Given that it's an autoincrement column.

osu.Server.Spectator/Hubs/Multiplayer/Matchmaking/Queue/MatchmakingQueueBackgroundService.cs
```diff
@@ -185,6 +187,8 @@ public async Task ExecuteOnceAsync()
                 logger.LogError(ex, "Failed to refresh the matchmaking queue.");
             }
 
+            await refreshPools();
```
bdach: Should probably try-catch this so that the entire background service doesn't fall over if anything in this dies.

smoogipoo:
On master the full pool is queried every time, so if that's an issue then yeah perhaps a combination of this PR with its refresh period plus setting the pool manually.

Also the pool size with this query is about 10000 items.

The simplification is that we wouldnt need the special case for when to query global beatmaps, and could keep the existing `selection_count` stuff.

peppy (approved):

PR #383, opened 2025-12-05
Don't kick multiplayer users on disconnect

smoogipoo:
The server-side counterpart to https://github.com/ppy/osu/pull/35889

Rather than kicking them, mark them as a normal leave instead. Nothing appears to fall over, and this is probably better messaging.


bdach (approved):
worth a try

PR #393, opened 2025-12-07
Mark early abandoned users + devalue the match

smoogipoo:
- [x] Depends on https://github.com/ppy/osu-server-spectator/pull/391
- [x] Depends on https://github.com/ppy/osu/pull/35923

Works best in combination with https://github.com/ppy/osu-server-spectator/pull/392.

### DIFF
--- a/osu.Server.Spectator/Hubs/Multiplayer/Matchmaking/MatchmakingMatchController.cs
+++ b/osu.Server.Spectator/Hubs/Multiplayer/Matchmaking/MatchmakingMatchController.cs
@@ -186,7 +186,16 @@ namespace osu.Server.Spectator.Hubs.Multiplayer.Matchmaking
         {
             anyPlayerQuit = true;
 
-            if (isMatchComplete())
+            // Mark users that leave the match early as having abandoned the match.
+            if (hasGameplayRoundsRemaining())
+            {
+                state.Users.GetOrAdd(user.UserID).AbandonedAt = DateTimeOffset.UtcNow;
+                state.RecordScores([], placement_points); // Empty update to adjust placements.
+                await hub.NotifyMatchRoomStateChanged(room);
+            }
+
+            // Attempt to conclude the match in advance so users don't have to keep playing rounds by themselves.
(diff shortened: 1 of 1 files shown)


osu.Server.Spectator/Hubs/Multiplayer/Matchmaking/MatchmakingMatchController.cs
```diff
@@ -459,17 +461,58 @@ private bool anyUsersReady()
...
+        {
+            return state.CurrentRound < totalRounds() || state.Stage <= MatchmakingStage.Gameplay;
+        }
+
+        /// <summary>
+        /// Whether this room is in head-to-head mode and should use best-of scoring.
+        /// </summary>
+        private bool isHeadToHead()
+        {
+            return state.Users.Count == 2;
```
peppy: I think there needs to be a `AppSettings.MatchmakingHeadToHeadIsBestOf` usage here still (or somewhere). Or that setting should just be removed?

osu.Server.Spectator/Hubs/Multiplayer/Matchmaking/MatchmakingMatchController.cs
```diff
@@ -459,17 +461,58 @@ private bool anyUsersReady()
...
+        private bool isBestOf()
+        {
+            return state.Users.Count == 2 && AppSettings.MatchmakingHeadToHeadIsBestOf;
+        }
+
+        /// <summary>
+        /// The total number of rounds playable.
+        /// In <see cref="isBestOf">best-of</see> mode, the number of rounds is reduced to that of a best-of scenario.
+        /// </summary>
+        private int totalRounds()
```
peppy: I'm still not sure.

It seems like the idea behind `BestOf` is that there's a set max rounds, but it can end early if a single user has too many points.

But `totalRounds()` is used directly in `hasGameplayRoundsRemaining()`, which has no knowledge of the points situation and will always use the minimum rounds number, making it still incorrect. Am I missing something?

PR #396, opened 2025-12-10
Fix broken best-of condition

smoogipoo:


smoogipoo:
Yes, the problem is 4 results can be W,W,L,L which equals 54 points (15 * 2 + 12 * 2), which is greater than the 15*3 point condition.

peppy (approved):

PR #398, opened 2025-12-11
Award 1/0 placement points in best-of mode

smoogipoo:
Maybe fixes https://github.com/ppy/osu/issues/35941?


peppy:
Is this basically a workaround where now losses give the same as fail scores?

Put another way, if I asked for the best of numbers to be adjusted where the second is non-zero, this would regress the "fix"? If so I think this needs to be mentioned inline.

(i bring this up because my first thought in response to this was that it's a bit sad that we're going back to a 1:1 mapping of points to win count)

PR #409, opened 2025-12-23
Fix missing scores not being counted

smoogipoo:
Fixes https://github.com/ppy/osu/issues/35941

Previously, these scores were completely missing and would "award" 0 points. Now they are considered as 0 total score, and awarded the appropriate number of points.

This is a fix I'm backporting from ranked play: https://github.com/smoogipoo/osu-server-spectator/blob/a8b735abc040db3242d714e45c9eef0491fc001c/osu.Server.Spectator/Hubs/Multiplayer/Matchmaking/RankedPlayMatchController.cs#L304-L316


smoogipoo:
Indeed it would.

peppy (approved):

PR #410, opened 2025-12-24
Fix `LoadRequested` sent to all users in multiplayer rooms

smoogipoo:
This is part 2 to fixing https://github.com/ppy/osu/issues/36045, and mentioned as bullet point 1 in https://github.com/ppy/osu/pull/36121:

> The multiplayer server sends `LoadRequested` to all users in the room regardless of their beatmap availability. This is an issue on its own that will be fixed separately.

I'm not 100% sure on this change in principle, and there are 3 paths that we can go down here:
1. This change.
2. Re-introduce the server-side "gameplay" group rather than using `.Users()`. This seems like it'd be a more complex change.
3. Continue sending `LoadRequested` to all users in the room and handle this locally in the client.

### DIFF
--- a/osu.Server.Spectator.Tests/Multiplayer/MultiplayerFlowTests.cs
+++ b/osu.Server.Spectator.Tests/Multiplayer/MultiplayerFlowTests.cs
@@ -5,6 +5,7 @@ using System.Threading.Tasks;
 using Microsoft.AspNetCore.SignalR;
 using Moq;
 using osu.Game.Online.Multiplayer;
+using osu.Game.Online.Rooms;
 using osu.Server.Spectator.Database.Models;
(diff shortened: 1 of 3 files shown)


smoogipoo:
The number of test changes here makes me feel a bit uneasy, as if option 3 from the OP is the better one to go down... Let's treat this PR as an RFC.

bdach:
> The number of test changes here makes me feel a bit uneasy

Without real full-stack testing I'm kind of unbothered by anything in here? Unsure whether this is because of lack of bothering capacity or there's really not much to see here.

peppy:
@smoogipoo shall we give this another try? needs some conflict resolution.

PR #466, opened 2026-04-14
Delay closing tournament rooms for 30min after last user left

bdach:
Closes https://github.com/ppy/osu-server-spectator/issues/460

## [Add support for delayed closing of tournament mode multiplayer rooms](https://github.com/ppy/osu-server-spectator/commit/09898c7f2641cd3ad854c8f4e74bc22a74dcdc46)

The way this works is:

1. When the last user leaves a tournament mode room, the room controller will set an end date for the room in the database to 30 minutes in the future. This is persisted to the database, which makes the rooms disappear on their own in the client-side room listing because that's handled by osu-web which will respect the new end date.
2. If any user joins the room before its end date elapses, the end date is cleared to `null` again.
3. If point (2) never occurs before the end date elapses, `ServerMultiplayerRoom.CanJoinRoom()` now checks the end date, which ensures that after that date passes, no user can join the room anymore.
4. A new job, `MultiplayerRoomLifetimeBackgroundService`, is called into existence. Every minute, it looks for tournament rooms with end dates in the past. If it finds any, it destroys their server-side memory state, and also dissociates any referees from said rooms to complete the cleanup.

Of particular note, this took some non-minor reshuffling of how referee user state is managed because of the expectations here.

- Referees join refereeing rights:
  - On creation and subsequent joining of a tournament room.
  - On being added as a referee by another tournament room's host.
- Referees lose refereeing rights:
  - On being removed as a referee by the tournament room's host.
  - On the room being force-closed by themselves or another referee.
  - On the room being cleaned up by the background job.

The above lists are exhaustive, e.g. those are the *only* conditions that cause granting / losing referee rights. No other interaction causes either.

Obviously, the referee privileges are still server-side memory state and will get lost on a redeploy / restart.

Additionally note that the referee hub does not provide a host change operation and I'm kind of fine with that until someone really needs it for whatever reason - adding it *is* possible, but would require further safeties in terms of per-user open room quotas and similar that I'd rather not get into if I don't need to.

## [Remove `LeaveRoom()` referee hub operation](https://github.com/ppy/osu-server-spectator/commit/36bf482955c7cf7cc723f179798214931537c844)

Now that it doesn't shed referee permissions, it's kind of no real use anymore. It was being very awkward in terms of behaviours given the revised concept of referee permission management so I just removed it.

## [Require being room host for adding / removing referees](https://github.com/ppy/osu-server-spectator/commit/2acf6a659d56002c0eb43cfaabe6f4d30f8aced9)

This is partly to match bancho, partly to reduce undesirable scenarios that could cause weird breakage. For instance, imagine this sequence of steps:

1. User A creates room
2. User A adds user B as referee
3. User B removes user A as referee from room

Now A is in a sticky state as the room is chalked up to them, goes on their open tournament room quota, but they're locked out of it (they can't join the room, as removal of a referee revokes rights). So just disallow this to begin with. With this change only A can add or remove referees.

## [Add primitive for listing all refereed rooms](https://github.com/ppy/osu-server-spectator/commit/ce4d87e5e63c3130eb5cc0dc0f3d70dc33056f83)

Provided for convenience / information as to which rooms can be recovered on a reconnection.

---

@ilw8 @ThePooN would appreciate if you read through the above description of the changes and see if there's anything you'd find objectionable.

### DIFF
--- a/osu.Server.Spectator.Tests/Multiplayer/RoomParticipationTests.cs
+++ b/osu.Server.Spectator.Tests/Multiplayer/RoomParticipationTests.cs
@@ -218,7 +218,7 @@ namespace osu.Server.Spectator.Tests.Multiplayer
         [Fact]
         public async Task UserJoinPreJoinFailureCleansUpRoom()
         {
-            Database.Setup(db => db.MarkRoomActiveAsync(It.IsAny<MultiplayerRoom>()))
+            Database.Setup(db => db.SetRoomEndDateAsync(It.IsAny<MultiplayerRoom>(), It.IsAny<DateTimeOffset?>()))
                     .ThrowsAsync(new Exception("error"));
 
             await Assert.ThrowsAnyAsync<Exception>(() => Hub.JoinRoom(ROOM_ID));
--- a/osu.Server.Spectator/Database/DatabaseAccess.cs
+++ b/osu.Server.Spectator/Database/DatabaseAccess.cs
@@ -134,13 +134,14 @@ namespace osu.Server.Spectator.Database
                 })).ToArray();
         }
 
-        public async Task MarkRoomActiveAsync(MultiplayerRoom room)
+        public async Task SetRoomEndDateAsync(MultiplayerRoom room, DateTimeOffset? endDate)
         {
             var connection = await getConnectionAsync();
 
-            await connection.ExecuteAsync("UPDATE multiplayer_rooms SET ends_at = null WHERE id = @RoomID", new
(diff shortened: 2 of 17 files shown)


ThePooN:
> would appreciate if you read through the above description of the changes and see if there's anything you'd find objectionable.

Sounds good to me (haven't read the code).

ILW8:
Looks good to me as well

smoogipoo (approved):
LGTM, looks like this turned out well in the end.

peppy (approved):
👌

PR #467, opened 2026-04-15
Ranked Play: Fix late-arriving scores not being considered

### LINKED ISSUE #37296: [ranked play] First person to leave mid-gameplay deals damage when both people leave mid-round regardless of score
### Type

Game behaviour

### Bug description

Title should explain it basically but ill elaborate.

Basically, when both people in a ranked play match leave during a round, the score of the first person is used for dealing damage even if they had less score than the second person. This means that if both people leave around the end of the map, not only can you get punished regardless of how well you did, you will get punished by the whole score of the person that left first, meaning you can both have around 800k score and you get hit by the 800k rather than the score difference between the two.

### Screenshots or videos

https://github.com/user-attachments/assets/b1bbd424-8489-4fd9-8b04-bc0f9c50874c

### Version

latest master build at time of writing

### Logs

[compressed-logs.zip](https://github.com/user-attachments/files/26722294/compressed-logs.zip)

smoogipoo:
Should fix https://github.com/ppy/osu/issues/37296

Waits up to 10 seconds for scores to arrive at osu!web/in the database (can be reduced upon request).

Don't know of a better way to do this...

### DIFF
--- a/osu.Server.Spectator.Tests/RankedPlay/RankedPlayStageImplementationTest.cs
+++ b/osu.Server.Spectator.Tests/RankedPlay/RankedPlayStageImplementationTest.cs
@@ -2,6 +2,7 @@
 // See the LICENCE file in the repository root for full licence text.
 
 using System;
+using System.Collections.Generic;
 using System.Linq;
 using System.Threading.Tasks;
 using Moq;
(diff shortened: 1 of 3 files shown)


peppy (approved):
Seems ok

PR #469, opened 2026-04-18
Populate first placements (wins) statistic

smoogipoo:
https://discord.com/channels/188630481301012481/1440912440224120882/1495053878901608458

I don't really like the structure of this code but I don't have any better ideas right now :(


smoogipoo:
First placements can be updated in the DB directly via:

```sql
UPDATE matchmaking_user_stats u
SET first_placements = (SELECT COUNT(*) FROM matchmaking_user_elo_history WHERE pool_id = u.pool_id AND user_id = u.user_id AND result = 'win')
WHERE pool_id IN (38, 39, 40, 41, 42);
```

peppy (approved):

PR #471, opened 2026-04-20
Fix ranked play matches getting stuck if players fail to load beatmaps

smoogipoo:
When the beatmap fails to load for 2 minutes, the player(s) are given a 100K score penalty and the next round is started. No results screen.


osu.Server.Spectator/Hubs/Multiplayer/Matchmaking/RankedPlay/Stages/GameplayWarmupStage.cs
smoogipoo, line 1: Although it technically shouldn't be possible, I've decided to remove the infinite timeout from this stage too. Just for safety.

bdach (approved):

PR #473, opened 2026-04-21
Give matchmaking rooms better names

smoogipoo:
Example:
```
osu! (RP: Season 0): smgi vs smgi1
```


bdach:
> [Web-side components](https://osu.ppy.sh/users/1040328/quickplay) use their own naming scheme

Looking at [where this actually appears to happen in web source](https://github.com/ppy/osu-web/blob/3b9315a6faf2623306fedea04826c16c3bcd6136/app/Models/Multiplayer/Room.php#L373-L375), this change would break that scheme unless the web check is adjusted to adapt to this change.

Maybe you're aware, maybe not, I'm just not sure since it is not spelled out explicitly.

> server-spectator has no concept of user names

You could very well get them via

https://github.com/ppy/osu-server-spectator/blob/843cb8783929e40fa91b66821f68f108a56a6abf/osu.Server.Spectator/Database/IDatabaseAccess.cs#L33-L36

nanaya:
If anything, proper room name as it looks like being done here seems better

smoogipoo:
Changed it to use usernames, I believe the osu!web code above doesn't need to change to account for older unnamed rooms.

peppy (approved):

PR #474, opened 2026-04-21
Reduce results stage duration

smoogipoo:
One complaint (e.g. [reddit](https://www.reddit.com/r/osugame/comments/1sr1trx/constructive_feedback_on_ranked_play_match_format/), [forums](https://osu.ppy.sh/community/forums/topics/2198397?n=118), Koifishu) I'm hearing is that the results screen takes too long. Its animations should be done within around 5 seconds, so 10 seconds total duration may be appropriate here.

### DIFF
--- a/osu.Server.Spectator/Hubs/Multiplayer/Matchmaking/RankedPlay/Stages/ResultsStage.cs
+++ b/osu.Server.Spectator/Hubs/Multiplayer/Matchmaking/RankedPlay/Stages/ResultsStage.cs
@@ -25,7 +25,7 @@ namespace osu.Server.Spectator.Hubs.Multiplayer.Matchmaking.RankedPlay.Stages
         }
 
         protected override RankedPlayStage Stage => RankedPlayStage.Results;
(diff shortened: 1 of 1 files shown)


bdach:
> Its animations should be done within around 5 seconds, so 10 seconds total duration may be appropriate here.

This means that the user has about 5 seconds to grasp all of the information on the results screen, because most of it isn't visible before the animations finish.

In checking with the game-side test scene which is already using a 10sec countdown, my brain cannot really parse anything on the screen in that time other than the huge rank letters and the health difference maybe.

Possibly skill / cognition issue on my behalf though at this point ¯\\_(ツ)\_/¯

I'd advocate for something as high as 15s maybe.

peppy:
I'd agree with 15 seconds. Was going to propose the same myself.

smoogipoo:
Had a feeling that was going to come up 😆 Incremental changes sounds fine for now.

peppy (approved):

PR #475, opened 2026-04-22
Fix rating update being delayed slightly

smoogipoo:
Here's two common cases where rating wouldn't be updated immediately. In both of these, it would wait until a delayed transition to the ended stage:
- If a player leaves during gameplay.
- If a player leaves during the results stage, even if in the final round.

The resulting effect is that the player would receive the wrong rating if they re-queued immediately. No data has been lost and live ratings are still accurate.

### DIFF
--- a/osu.Server.Spectator.Tests/RankedPlay/Stages/GameplayStageTests.cs
+++ b/osu.Server.Spectator.Tests/RankedPlay/Stages/GameplayStageTests.cs
@@ -69,5 +69,14 @@ namespace osu.Server.Spectator.Tests.RankedPlay.Stages
             Assert.Equal(RankedPlayStage.Gameplay, RoomState.Stage);
             Assert.Equal(0, UserState.Life);
         }
+
+        [Fact]
+        public async Task UserRatingUpdatedImmediatelyOnPlayerLeave()
+        {
+            await Hub.LeaveRoom();
+
+            Assert.Equal(RankedPlayStage.Gameplay, RoomState.Stage);
+            Assert.True(MatchController.UserRatingsUpdated);
+        }
     }
 }
--- a/osu.Server.Spectator.Tests/RankedPlay/Stages/GameplayStageTests.cs
+++ b/osu.Server.Spectator.Tests/RankedPlay/Stages/GameplayStageTests.cs
@@ -62,12 +62,15 @@ namespace osu.Server.Spectator.Tests.RankedPlay.Stages
         }
 
         [Fact]
-        public async Task ContinuesToEndedWhenAnyPlayerLeaves()
+        public async Task ContinuesToResultsWhenAnyPlayerLeaves()
         {
             await Hub.LeaveRoom();
 
             Assert.Equal(RankedPlayStage.Gameplay, RoomState.Stage);
             Assert.Equal(0, UserState.Life);
+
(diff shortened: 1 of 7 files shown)


osu.Server.Spectator/Hubs/Multiplayer/Matchmaking/RankedPlay/RankedPlayMatchController.cs
```diff
@@ -63,6 +66,12 @@ public class RankedPlayMatchController : IMatchController, IMatchmakingMatchCont
         /// </summary>
         private readonly List<RankedPlayCardItem> deck = [];
 
+        /// <summary>
+        /// Indicates whether the final user ratings have been updated.
+        /// Todo: This is public for testing purposes, but should not be.
```
smoogipoo, line 73: This is a todo to be fixed _at some point in the future_, not in this PR. I will probably refactor tests entirely while doing this.

bdach: I would say that that is a good idea because holy [REDACTED] the `EndedStageTests` changes are *something*.

osu.Server.Spectator.Tests/RankedPlay/Stages/GameplayStageTests.cs
```diff
@@ -62,12 +62,12 @@ public async Task ContinuesToResultsOnGameplayCompletion()
         }
 
         [Fact]
-        public async Task ContinuesToEndedWhenAnyPlayerLeaves()
+        public async Task UserRatingUpdatedImmediatelyOnPlayerLeave()
         {
             await Hub.LeaveRoom();
 
             Assert.Equal(RankedPlayStage.Gameplay, RoomState.Stage);
-            Assert.Equal(0, UserState.Life);
```
bdach, line 70: Any reason why this assert was completely dropped? Looks like a useful one to me. Still seems to pass, too.

smoogipoo: Previous commit was titled "Fix broken test" and should've been "**Remove** broken test".

I've added it back in 34c97120d7157c7273972c4f398bee1b0c7674a2, and tried to give it some semblance of testing things correctly. Though that does mean it no longer matches its previous description, which as above was broken anyway.

bdach:
merge conflicts here

osu.Server.Spectator/Hubs/Multiplayer/Matchmaking/RankedPlay/RankedPlayMatchController.cs
```diff
@@ -318,6 +327,89 @@ public async Task ActivateCard(RankedPlayCardItem card)
             LastActivatedCard = card;
         }
 
+        public async Task HandleMatchCompleted()
+        {
+            if (UserRatingsUpdated)
+                return;
+
+            UserRatingsUpdated = true;
```
peppy, line 335: Not super convinced this is the correct naming for this flag, because it's guarding this whole method, not just user ratings.

peppy (approved):
Getting this in because it's probably best to have it in according to smoogi.

peppy (approved):

PR #476, opened 2026-04-22
Implement beatmap rating adjustments

smoogipoo:
- [x] Depends on https://github.com/ppy/osu-web/pull/12930

This changes the meaning of the `matchmaking_pool_beatmaps` table.
- Currently, it represents the full set of beatmaps available for a given pool, with the global beatmaps as a fallback path if the table is empty.
- With this PR, it is used _in addition_ to provide ratings for the global beatmaps.

For this to work, there is an update path that now _inserts_ into the `matchmaking_pool_beatmaps` table whenever beatmap rating is updated.

The way rating adjustments work is: we run the average player against the beatmap such that players are considered to have "won" against the beatmap if the average score is more than 700K.

### DIFF
--- a/osu.Server.Spectator.Tests/Matchmaking/MatchmakingBeatmapSelectorTest.cs
+++ b/osu.Server.Spectator.Tests/Matchmaking/MatchmakingBeatmapSelectorTest.cs
@@ -21,15 +21,15 @@ namespace osu.Server.Spectator.Tests.Matchmaking
             MatchmakingBeatmapSelector beatmapSelector = new MatchmakingBeatmapSelector(
                 Enumerable.Range(1, 1000).Select(i => new matchmaking_pool_beatmap
                 {
-                    id = (uint)i,
+                    beatmap_id = i,
                     rating = 1000,
-                }).ToArray());
+                }).ToArray(), new Mock<IDatabaseFactory>().Object);
 
             matchmaking_pool_beatmap[] result = beatmapSelector.GetAppropriateBeatmaps(50, [new EloRating(1500, 80)]);
             Assert.Equal(50, result.Length);
 
             matchmaking_pool_beatmap[] result2 = beatmapSelector.GetAppropriateBeatmaps(50, [new EloRating(1500, 80)]);
-            Assert.NotEqual(result.OrderBy(b => b.id), result2.OrderBy(b => b.id));
+            Assert.NotEqual(result.OrderBy(b => b.beatmap_id), result2.OrderBy(b => b.beatmap_id));
         }
 
         [Fact]
@@ -38,15 +38,15 @@ namespace osu.Server.Spectator.Tests.Matchmaking
             MatchmakingBeatmapSelector beatmapSelector = new MatchmakingBeatmapSelector(
                 Enumerable.Range(1, 1000).Select(i => new matchmaking_pool_beatmap
                 {
-                    id = (uint)i,
+                    beatmap_id = i,
                     rating = 1500,
-                }).ToArray());
+                }).ToArray(), new Mock<IDatabaseFactory>().Object);
 
             matchmaking_pool_beatmap[] result = beatmapSelector.GetAppropriateBeatmaps(50, [new EloRating(1500, 80)]);
             Assert.Equal(50, result.Length);
 
             matchmaking_pool_beatmap[] result2 = beatmapSelector.GetAppropriateBeatmaps(50, [new EloRating(1500, 80)]);
-            Assert.NotEqual(result.OrderBy(b => b.id), result2.OrderBy(b => b.id));
+            Assert.NotEqual(result.OrderBy(b => b.beatmap_id), result2.OrderBy(b => b.beatmap_id));
         }
 
         [Fact]
@@ -55,15 +55,15 @@ namespace osu.Server.Spectator.Tests.Matchmaking
             MatchmakingBeatmapSelector beatmapSelector = new MatchmakingBeatmapSelector(
                 Enumerable.Range(1, 1000).Select(i => new matchmaking_pool_beatmap
                 {
-                    id = (uint)i,
+                    beatmap_id = i,
                     rating = 2000,
-                }).ToArray());
+                }).ToArray(), new Mock<IDatabaseFactory>().Object);
 
             matchmaking_pool_beatmap[] result = beatmapSelector.GetAppropriateBeatmaps(50, [new EloRating(1500, 80)]);
             Assert.Equal(50, result.Length);
 
             matchmaking_pool_beatmap[] result2 = beatmapSelector.GetAppropriateBeatmaps(50, [new EloRating(1500, 80)]);
-            Assert.NotEqual(result.OrderBy(b => b.id), result2.OrderBy(b => b.id));
+            Assert.NotEqual(result.OrderBy(b => b.beatmap_id), result2.OrderBy(b => b.beatmap_id));
         }
 
         [Fact]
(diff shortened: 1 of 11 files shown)


IceDynamix:
i've been running simulations of map ratings on the [osu! tournament rating](https://otr.stagec.net/) [dataset](https://data.otr.stagec.net). the method i used is similar, except adapted to work with multiple players. my method takes every single participant in the match as an individual team, and the map with the specified score as an individual team, which means for ranked play it would be a ffa match of 3 players, which would avoid the issue of ordering as simulating vs p1 and then p2 might create different results compared to p2 and then p1. i'm not sure whether this approach is better or not.

<details>

<summary>translation of my method into c# code to avoid ambiguity</summary>

```cs
//                            note the arrays ----vv                 -------vv
public void AdjustRating(BeatmapLookupKey key, int[] playerScores, EloRating[] playerRatings)
{
    lock (beatmaps)
    {
        /**/

        PlackettLuce model = new PlackettLuce
        {
            /**/
        };

        double scoreToWin = pool.ruleset_id switch
        {
            /**/
        };

        IRating[] ratings = model.Rate(
                [
                    new Team { Players = [model.Rating(beatmap.rating, beatmap.rating_sig)] },
                    // use loop to add player ratings, simplified for now
                    new Team { Players = [model.Rating(playerRatings[0].rating, playerRatings[0].rating_sig)] },
                    new Team { Players = [model.Rating(playerRatings[1].rating, playerRatings[1].rating_sig)] }
                ],
                scores:
                [
                    scoreToWin,
                    // use loop to add player scores, simplified for now
                    playerScores[0],
                    playerScores[1],
                ]
            )
            .Select(t => t.Players.Single())
            .ToArray();

        beatmap.rating = ratings[0].Mu;
        beatmap.rating_sig = ratings[0].Sigma;

        pendingBeatmapUpdates.Add(key);
    }
}
```

</details>

here's the result of an sql i ran on all recorded nomod scores per ruleset in osu! tournament rating, with 25th/50th/75th percentiles added. note that this is score v2 and its tournament environment, which does not accurately reflect the environment of ranked (sightreading, or for example mania tournaments focus on high accuracy).

| ruleset | avg | stdev | P\_25 | P\_50 | P\_75 | count |
| :--- | :--- | :--- | :--- | :--- | :--- | :--- |
| osu | 516824 | 250657 | 340749.5 | 482699 | 668163.25 | 560624 |
| taiko | 936457 | 167384 | 917355.5 | 976146 | 1010839.25 | 61260 |
| catch | 904497 | 145711 | 838954 | 929469 | 1000000 | 44615 |
| 4k | 960875 | 88081 | 963526 | 982020 | 991303 | 169429 |
| 7k | 931151 | 97792 | 920598 | 957090 | 977793 | 22037 |

<details>

<summary>sql used (postgres)</summary>

```sql
SELECT CASE
           WHEN gs.ruleset = 0 THEN 'osu'
           WHEN gs.ruleset = 1 THEN 'taiko'
           WHEN gs.ruleset = 2 THEN 'catch'
           WHEN gs.ruleset = 4 THEN '4k'
           WHEN gs.ruleset = 5 THEN '7k'
           END AS                                               ruleset
        ,
       ROUND(AVG(gs.score))                                     "avg",
       ROUND(STDDEV(gs.score))                                  "stdev",
       PERCENTILE_CONT(0.25) WITHIN GROUP ( ORDER BY gs.score ) "P_25",
       PERCENTILE_CONT(0.5) WITHIN GROUP ( ORDER BY gs.score )  "P_50", -- <<<<<
       PERCENTILE_CONT(0.75) WITHIN GROUP ( ORDER BY gs.score ) "P_75",
       COUNT(*)                                                 "count"
FROM game_scores gs
         JOIN public.games g ON gs.game_id = g.id
WHERE g.mods = 0
  AND g.verification_status = 4
GROUP BY gs.ruleset
ORDER BY gs.ruleset;
```

</details>

running a similar query on the current ranked plays would get more accurate readings (lazer scoring, sightreading etc). regardless, the stdev seems to be a good indicator for the hp scaling

the clear thresholds i used in my simulation were chosen based on the 25th percentile, but can easily be changed to a different one

```rs
fn clear_threshold(mods: Mods, ruleset: &Ruleset) -> f64 {
    let base = match ruleset {
        Ruleset::Osu => 340_000.0,
        Ruleset::Taiko => 920_000.0,
        Ruleset::Catch => 840_000.0,
        Ruleset::ManiaOther | Ruleset::Mania4k => 960_000.0,
        Ruleset::Mania7k => 920_000.0
    };

    let mod_mult = mods.mod_score_multiplier(ruleset);

    base * mod_mult
}
```

also, might want to update the pr comment to mention ruleset-specific goal scores rather than the 700k number

IceDynamix:
i've published my simulation code at [IceDynamix/otr-processor](https://github.com/IceDynamix/otr-processor), with instructions on how to run the simulation yourself and a csv file of the results. again, this uses different scoring and different rating parameters/heuristics. it also has a few issues like not having access to modded star ratings in this experiment (see the repositories readme).

however, my conclusion on _this specific experiment on the o!tr dataset_, is that it didn't work that well
- tournament pools have an inherent goal of "being unique". it could be tournament customs, or simply avoiding maps that are popular or have been pooled often already, regardless it actively goes against the needs of a rating system which lives off the amount of data coming in
- rating systems work best when the rating is actively used to match players (or in this case maps) of similar rating. but in this case, there isnt really a feedback loop. mappoolers dont try to use this rating to make new pools
- a map being pooled too easy will result in a large number of matches where the map "loses", because the same map is used for a weeks worth of matches. but if there is no instance of the same map being pooled too difficult, that rating will never balance out. so again it's a lack of data combined with the missing context from the mappool

these issues are not present in the context of osu ranked play once fully implemented

smoogipoo:
<details>
<summary>Translated MYSQL query</summary>

```sql
SELECT
    ruleset_id,
    ROUND(AVG(total_score)) AS 'avg',
    ROUND(STDDEV(total_score)) AS 'stddev',
    MAX(CASE WHEN pct_rank <= 0.25 THEN total_score END) AS 'p_25',
    MAX(CASE WHEN pct_rank <= 0.50 THEN total_score END) AS 'p_50',
    MAX(CASE WHEN pct_rank <= 0.75 THEN total_score END) AS 'p_75',
    COUNT(*) AS `count`
FROM
(
    SELECT
        s.ruleset_id,
        s.total_score,
        PERCENT_RANK() OVER (PARTITION BY s.ruleset_id ORDER BY s.total_score) as pct_rank
    FROM multiplayer_rooms r
        JOIN multiplayer_playlist_items p
            ON p.room_id = r.id
        JOIN multiplayer_score_links l
            ON l.playlist_item_id = p.id
        JOIN scores s
            ON s.id = l.score_id
    WHERE r.type = 'ranked_play'
        AND s.passed = 1
) AS ranked_scores
GROUP BY ruleset_id
ORDER BY ruleset_id;
```

</details>

| ruleset\_id | avg | stddev | p\_25 | p\_50 | p\_75 | count |
| :--- | :--- | :--- | :--- | :--- | :--- | :--- |
| 0 | 530812 | 184990 | 397779 | 525877 | 655613 | 571324 |
| 1 | 833320 | 127306 | 752237 | 855062 | 935517 | 8694 |
| 2 | 812961 | 140999 | 723618 | 838287 | 923127 | 3992 |
| 3 | 809792 | 132314 | 728671 | 838605 | 915760 | 59315 |

I'd probably be more inclined to use p_50 here, because p_25 feels very low.

How about something like:
```
osu: 550000
taiko: 850000
catch: 850000
mania: 850000
```

IceDynamix:
values look good. as a sidenote, there is likely a non-negligible difference between 4k and 7k, intuition-wise i think the sql would say something around 800k. i think its fine to keep 850k though

it might be worth to check the stddev of scores set after beatmap ratings are implemented, the stddev should be lower as a result of more balanced mappools

peppy (approved):

PR #478, opened 2026-04-23
Fix updating beatmap ratings using the wrong values

smoogipoo:
The fix here is the first commit. I changed around the way the contest was populated, and forgot to adjust the indices.

The second commit is an attempt to remedy the situation without requiring two server startups by always re-querying the `matchmaking_pool_beatmaps` table. The intentional deploy process here is:
- Deploy this version of server spectator.
- Wait 6 hours for the previous instance to die.
- Clear the `matchmaking_pool_beatmaps` table.

For about 1 hour, the server will then temporarily use the wrong rating to match with players, but I think that's acceptable.


peppy (approved):
Surface level check looks okay

PR #480, opened 2026-04-24
Adjust damage multipliers for taiko/catch/mania

smoogipoo:
The goal here is to keep the osu! damage multiplier the same, and increasing it slightly for the other rulesets which tend to have very equal scores.

osu: `[ 1, 1, 2, 2.5, 3, 3.5, 4, ... ]` (starts at 1, initially increases by 1, and then in 0.5 steppings)
others: `[ 1, 1, 2.5, 3.5, 4.5, 5.5, ... ]` (starts at 1, initially increases by 1.5, then in 1.0 steppings).


peppy (approved):

peppy:
This is very conflicty.

peppy (approved):

PR #483, opened 2026-04-28
Reduce beatmap rating breadth at lower skill levels

pacowoc:
There are widespread complaints from players on repeatedly getting a mixture of maps from a difficulty range that is way too wide. This is simply a value nerf to the issue.


peppy:
Please explain what this math is

smoogipoo (approved):

PR #485, opened 2026-04-28
Attempt to increase beatmap variety at high user rating

smoogipoo:
This uses a 2-pass filtering method.

The entire FA pool looks something like this:
image

Use a normal distribution to filter to 1000 beatmaps (providing relevancy) and then apply a uniform distribution on top (providing variety).

### DIFF
--- a/osu.Server.Spectator/Hubs/Multiplayer/Matchmaking/Queue/MatchmakingBeatmapSelector.cs
+++ b/osu.Server.Spectator/Hubs/Multiplayer/Matchmaking/Queue/MatchmakingBeatmapSelector.cs
@@ -17,6 +17,11 @@ namespace osu.Server.Spectator.Hubs.Multiplayer.Matchmaking.Queue
     public class MatchmakingBeatmapSelector
     {
         /// <summary>
+        /// The approximate number of osu! ruleset beatmaps that should be filtered down to for relevancy purposes.
+        /// </summary>
+        private const double osu_beatmap_proportion = 1 / 12.0;
+
+        /// <summary>
         /// Contains all ranked beatmaps.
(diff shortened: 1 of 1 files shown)


peppy (approved):

peppy (approved):

smoogipoo:
I'll tweak the values a bit before this is merged because apparently a large pool is still ok for mania even given m7k has ~800 beatmaps total.

osu.Server.Spectator/Hubs/Multiplayer/Matchmaking/Queue/MatchmakingBeatmapSelector.cs
```diff
@@ -157,7 +162,10 @@ public matchmaking_pool_beatmap[] GetAppropriateBeatmaps(int count, EloRating[]
                                double weight = Math.Clamp(Math.Exp(-Math.Pow(b.rating - userRatingMu, 2) / (2 * ratingSig * ratingSig)), 1e-6, 1);
                                return Math.Pow(Random.Shared.NextDouble(), 1.0 / weight);
                            })
```
IceDynamix, line 164: the specified sigma doesn't matter anymore because the count likely takes precedence and the sampling is done by the uniform distribution. all the first pass does is order by absolute difference between target rating and map rating (with a chance for outliers). could simplify the first pass like this?
```cs
return beatmaps.Values.OrderBy(b => Math.Abs(b.rating - userRatingMu))
```

peppy:
I'm going to get this in as-is, but the review and follow-up considerations are still valid and can be applied in a second round deploy.

peppy (approved):

PR #486, opened 2026-04-29
Punish early leavers by banning them from the matchmaking queue

smoogipoo:
This still has no messaging, but that can probably be done later? I don't think someone who quits would particularly care about being banned, unless they're being malicious in the first place.


Givikap120:
How this would affect players who had their game crashed or their internet connection to be temporarily lost?

smoogipoo:
We're going to be trying to get in https://github.com/ppy/osu-server-spectator/pull/366 to help with connection issues. Can't really do anything about games crashing...

IceDynamix:
maybe add an exception for duels/unranked? i dont think the user should be punished for leaving those

Userlog2:
This bans early leavers for 10 minutes, right?

pacowoc:
> This bans early leavers for 10 minutes, right?

Yes, it is a flat 10 minutes right after a forfeit

peppy (approved):
let's try this i guess?

Givikap120:
> We're going to be trying to get in #366 to help with connection issues. Can't really do anything about games crashing...

Add reconnection option, so player could just restart the game and reconnect to the match.

vinje123:
> let's try this i guess?

Add an exception if you have played 1 or multiple matches already. If you are down 0-3 to a player that's half your rank and clearly better you should be allowed to forfeit without being punished, it's just a time hostage situation at that point. If you are not allowed to forfeit why is it even an option top left in song select?

failaip12:
> > let's try this i guess?
> 
> Add an exception if you have played 1 or multiple matches already. If you are down 0-3 to a player that's half your rank and clearly better you should be allowed to forfeit without being punished, it's just a time hostage situation at that point. If you are not allowed to forfeit why is it even an option top left in song select?

Score doesn't matter in the slightest in this mode, hp is the most important, you could be losing 6-0 and one shot your opponent. 

In the same way rank doesn't matter, not everyone is consistent enough to win in ranked, if you are, or are good at a lot of skillsets, you can punch way above your weight, just look at ekoro, he is almost down to top 1k in pp, yet he is nr. 12 on ranked play. 

Anyway fix your mental bro.

vinje123:
> > > let's try this i guess?
> > 
> > 
> > Add an exception if you have played 1 or multiple matches already. If you are down 0-3 to a player that's half your rank and clearly better you should be allowed to forfeit without being punished, it's just a time hostage situation at that point. If you are not allowed to forfeit why is it even an option top left in song select?
> 
> Score doesn't matter in the slightest in this mode, hp is the most important, you could be losing 6-0 and one shot your opponent.
> 
> In the same way rank doesn't matter, not everyone is consistent enough to win in ranked, if you are, or are good at a lot of skillsets, you can punch way above your weight, just look at ekoro, he is almost down to top 1k in pp, yet he is nr. 12 on ranked play.
> 
> Anyway fix your mental bro.

Why does it matter to you if someone wants to forfeit a match or not? I'm just saying you should not be punished for forfeiting a match after having played multiple maps already, why is that a bad thing?

failaip12:
It matters cause they can be malicious and manipulate the ladder, and frankly 10 minutes is nothing. 

Maybe they could implement a system where it doesn't give you a punishment if you already played a few maps, but the question is, how many is enough to deter most of the malicious uses, while still providing a benefit to you. 

And worst case, and this is already being done I believe, you can just "quit" by not playing the map/getting 0 score.



Tho I will never understand forefeiting in a game and game mode like this, I've beaten people of much higher rank, and got beaten by people of much lower rank than me.

cihe13375:
I saw in a livestream where a player already knew that they would lose after the last map ( they only had like 20k hp left, and they lose that map by 50k). The player manually quit the room after that map was finished, but before the "result screen animation" was finished. I don't know if they would be banned in this case, but if so I think it's quite unexpected.

vinje123:
> It matters cause they can be malicious and manipulate the ladder, and frankly 10 minutes is nothing.
> 
> Maybe they could implement a system where it doesn't give you a punishment if you already played a few maps, but the question is, how many is enough to deter most of the malicious uses, while still providing a benefit to you.
> 
> And worst case, and this is already being done I believe, you can just "quit" by not playing the map/getting 0 score.
> 
> Tho I will never understand forefeiting in a game and game mode like this, I've beaten people of much higher rank, and got beaten by people of much lower rank than me.

I think there could be no punishment if you have played 2 maps. That would kind of not be any different from if they were wintrading with 1 player just afk no? Assuming the other player can get 500K/1000K on 2 beatmaps.

SuperYosh23:
Tbh i kinda really dont like this...

smoogipoo:
In pretty much every other competitive game I've played it's an all or nothing. Just because you _think_ you've lost doesn't mean that you are, and you are directly influencing rankings by giving up/not playing to your best - the whole system relies on players being incentivised to win.

As mentioned above: "10 minutes is nothing".

PR #489, opened 2026-04-30
Fix being unable to duel with high rating difference

### LINKED ISSUE #37589: Duels don't start sometimes
### Type

Game behaviour

### Bug description

Just tried the new duels feature from #37556

When i tried to duel a friend with way higher elo, after accepting the duel, both of us would be stuck waiting for it to put us in a lobby.
I assume this is because of the elo difference: 1200 to 1800. The game should either let you play with a friend even if the elo mismatches, or it should give an error message.

I don't know if the problem is the elo mismatch, but it seemed like it. We were able to start a duel in mania, where we have the same elo.

### Screenshots or videos

_No response_

### Version

2026.429.0

### Logs

Ill post the logs if asked for in a message

smoogipoo:
Should fix https://osu.ppy.sh/community/forums/topics/2201839
Should fix https://github.com/ppy/osu/issues/37589

This variable didn't exist when I initially implemented in quick play...


TheRealStevie:
Related issue: https://github.com/ppy/osu/issues/37589

peppy (approved):

PR #496, opened 2026-05-10
Speed up queue time by scaling window with player rating

pacowoc:
The gradually increasing window creates unusually long queue times for higher elo players due to the low player volume and huge rating gap between them. This pr aims to remedy this problem with minimal change to the spirit of the system.

The math for this change is the reciprocal of a bell curve (the coefficients are nerfed because a direct reciprocal of the full playerbase elo distribution is way too extreme.)  

https://www.desmos.com/calculator/eouhvb4txz
Graph

### DIFF
--- a/osu.Server.Spectator/Hubs/Multiplayer/Matchmaking/Queue/MatchmakingQueue.cs
+++ b/osu.Server.Spectator/Hubs/Multiplayer/Matchmaking/Queue/MatchmakingQueue.cs
@@ -358,17 +358,11 @@ namespace osu.Server.Spectator.Hubs.Multiplayer.Matchmaking.Queue
 
                 double distance = Math.Min(leftDistance, rightDistance);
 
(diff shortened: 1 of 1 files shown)


smoogipoo (approved):
We can give this a shot though I still think we'll want the static search radius. The idea is two part: first to make high rating players match immediately, but second is to increase the window across the board to introduce a bit of randomness/unfairness and promote rating jumps.

Of course, the first one of those is the most common...

pacowoc:
> We can give this a shot though I still think we'll want the static search radius. The idea is two part: first to make high rating players match immediately, but second is to increase the window across the board to introduce a bit of randomness/unfairness and promote rating jumps.
> 
> Of course, the first one of those is the most common...

My main concern with instant match in high elo is that matches will still be recurrent with 3+ high elo online concurrently since people immediately get paired again no matter what after a match and the two players in a match reenter the pool roughly at the same time

PR #500, opened 2026-05-11
Expand referee hub API documentation

bdach:
A few things missing here, like how disconnections work, or what to do on a server reboot.

Still leaving the "this API is a work in progress" notice at the top of the document until slot support is in.

Also contains a github action bump to fix deprecation warnings.

### DIFF
--- a/.github/workflows/deploy-docs.yml
+++ b/.github/workflows/deploy-docs.yml
@@ -37,4 +37,4 @@ jobs:
       
       - name: Deploy to GitHub Pages
         id: deployment
-        uses: actions/deploy-pages@v4
+        uses: actions/deploy-pages@v5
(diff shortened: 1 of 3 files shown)


smoogipoo (approved):

bdach:
@peppy The changes I made here are not actually live [on this page](https://ppy.sh/osu-server-spectator/referee-hub-api.html) for whatever reason. Do you see any reason why that might be that's not github?

For what it's worth the [pages artifact from the latest successful relevant workflow run](https://github.com/ppy/osu-server-spectator/actions/runs/25777956598) looks correct in that it has the missing changes.

peppy:
@bdach https://ppy.sh/osu-server-spectator/referee-hub-api.html?123 works, so just a caching issue. should purge given time

PR #501, opened 2026-05-12
Allow top-100 players to always be able to pair

smoogipoo:
In a recent mrekk stream, he was stuck in the queue for 2 hours unable to pair with anybody because he's too far ahead in rating. This change is intended to make players within the top-100 be able to eventually pair with each other by increasing the maximum search radius to cover the top-100 window.

### DIFF
--- a/osu.Server.Spectator.Tests/Matchmaking/MatchmakingQueueTest.cs
+++ b/osu.Server.Spectator.Tests/Matchmaking/MatchmakingQueueTest.cs
@@ -231,6 +231,34 @@ namespace osu.Server.Spectator.Tests.Matchmaking
             Assert.Equal(2, bundle.FormedGroups[0].Users.Length);
         }
 
+        [Fact]
+        public void Top100MatchesWithTop1()
+        {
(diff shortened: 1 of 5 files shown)


peppy:
Feels weird to be hacking something in like specifically for a top "100" number, but I'll assume you guys know what you're doing 🤷

peppy (approved):

IceDynamix:
the idea is use a metric that is somewhat agnostic to exact elo numbers, because they are different per mode/queue. more populated modes tend to have higher elo numbers, but at the same time also have more players. all that matters is to find *any* threshold where we consider activel player counts to be too low, which can be a simple constant because the number of players needed per match is also constant. 

elo range of top 100 per mode
- standard: 2131-2558
- taiko: 1897-2303
- catch: 1858-2304
- mania 4k: 1992-2362
- mania 7k: 1797-2352

i agree it's not the most elegant solution, but it's somewhat simple to understand and easy to adjust (just a single constant). so it will work fine until we can come up with a better method or the situation improves (e.g. more players)

PR #502, opened 2026-05-12
Simplify ranked play queue matchmaking

IceDynamix:
Slightly less performant than the code on master, but way more readable and easier to reason changes for.

If performance does become a bottleneck, can implement following change
```cs

IEnumerable<MatchmakingQueueUser> leftSideMatches = users
                                                    .Take(pivotIndex) // Excludes the pivot user
                                                    .Reverse() // Iterate starting from the pivot to the beginning of list
                                                    .TakeWhile(u => Math.Abs(pivotUser.Rating.Mu - u.Rating.Mu) <= searchRadius);

IEnumerable<MatchmakingQueueUser> rightSideMatches = users
                                                     .Skip(pivotIndex + 1) // Excludes the pivot user. If pivot user is the highest rated player, sequence is empty
                                                     .TakeWhile(u => Math.Abs(pivotUser.Rating.Mu - u.Rating.Mu) <= searchRadius);

IEnumerable<MatchmakingQueueUser> allMatches = leftSideMatches.Concat(rightSideMatches);
```

### DIFF
--- a/osu.Server.Spectator/Hubs/Multiplayer/Matchmaking/Queue/MatchmakingQueue.cs
+++ b/osu.Server.Spectator/Hubs/Multiplayer/Matchmaking/Queue/MatchmakingQueue.cs
@@ -337,6 +337,11 @@ namespace osu.Server.Spectator.Hubs.Multiplayer.Matchmaking.Queue
             MatchmakingQueueUser pivotUser = users[pivotIndex];
             HashSet<MatchmakingQueueUser> result = [pivotUser];
 
+            // Speedup user pairing by a coefficient based on rating
+            double ratingBonus = Math.Exp(Math.Pow((pivotUser.Rating.Mu - 1500) / 750, 2));
+            TimeSpan searchTime = Clock.UtcNow - pivotUser.SearchStartTime;
+            double searchRadius = Math.Min(Pool.rating_search_radius_max, Pool.rating_search_radius * ratingBonus * Math.Pow(2, searchTime.TotalSeconds / Pool.rating_search_radius_exp));
+
             int leftIndex = pivotIndex - 1;
             int rightIndex = pivotIndex + 1;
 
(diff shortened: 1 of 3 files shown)


smoogipoo:
Tests are failing on a NaN value, from `rating_search_radius_exp` being initialised to 0. It can be initialised to 300, like `rating_search_radius_max` is:

https://github.com/ppy/osu-server-spectator/blob/132fd097585a3005b1974f6fc1a4cf3b4ca28a5d/osu.Server.Spectator/Database/Models/matchmaking_pool.cs#L32-L40

osu.Server.Spectator/Hubs/Multiplayer/Matchmaking/Queue/MatchmakingQueue.cs
```diff
@@ -330,45 +331,25 @@ private IEnumerable<MatchmakingQueueUser[]> matchUsers()
...
-                if (distance > searchRadius)
-                    break;
-
-                result.Add(leftDistance < rightDistance ? users[leftIndex--] : users[rightIndex++]);
-            }
+            HashSet<MatchmakingQueueUser> result = [pivotUser];
+            result.AddRange(
+                allMatches
+                    .OrderBy(_ => Random.Shared.Next())
+                    .Take(Pool.lobby_size)
```
smoogipoo: I think this should be `Pool.lobby_size - 1`? Concerning that there's no tests for this.

smoogipoo (changes requested):

osu.Server.Spectator/Hubs/Multiplayer/Matchmaking/Queue/MatchmakingQueue.cs
IceDynamix: There is a test `Assert.Equal(2, bundle.FormedGroups[0].Users.Length);`. The test suite simply failed on the earlier assertion

smoogipoo (approved):

Issue #244, opened 2024-09-19
Dropping connection from spectator server mid-play results in truncated online replay


rvnexe:
Hi. I've downloaded a replay file from my profile to send it for someone, but it displays only a few seconds 'till the end.
This zip file contains the original replay file, which game client gave me. And the replay file which I downloaded from osu.ppy.sh website.
[replayfiles.zip](https://github.com/user-attachments/files/17057704/replayfiles.zip)

peppy:
The replay retrieved from web contains frames from 658 frames from 141,060 to 152,154 ms, towards the latter end of the beatmap.

This implies the saved replay was truncated. Likely a failed connection to `osu-server-spectator`.

I'm not sure how we should be handling this yet. Either the whole play should be invalidated, or maybe the client needs to send catch-up data. (low priority cc/ @bdach for visibility on this).

smoogipoo:
We could probably send the whole replay again after completion for validation?

bdach:
I swear I've seen this reported somewhere already and even recall reproducing a connection drop-out doing this but I can't find it now. Maybe a github account deletion sent it to the shadow realm.

peppy:
> We could probably send the whole replay again after completion for validation?

Maybe yeah. It's not a high prio item since this should happen quite rarely, but requires some discussion on validation concerns (ie. whether we want to use real-time packets as score validation piece).

bdach:
Rough initial plan of attack detailed below.

## Happy path
```mermaid
sequenceDiagram
	client->>server: BeginPlaySession(scoreToken: 1234, state)
	server-->>client: ACK
	
	client->>server: SendFrameData(bundle: { sequenceNumber: 1, ... })
	server-->>client: ACK
	
	client->>server: SendFrameData(bundle: { sequenceNumber: 2, ... })
	server-->>client: ACK
	
	Note over client,server: ...
	
	client->>server: SendFrameData(bundle: { sequenceNumber: 100, ... })
	server-->>client: ACK
	
	client->>server: EndPlaySession(state: { bundlesSent: 100, ... })
	server-->>client: ACK
	
	server->>client: CompleteReplay(scoreToken: 1234, bundles: [])
	activate client
	client-->>server: ACK
	activate server
	Note right of client: Client purges all bundles<br/>related to score token id=1234 from local buffers.
	deactivate client
	
	server->>S3: Upload replay
	S3-->>server: 200 OK
[...]

PR #503, opened 2026-05-13
Move current score out of spectator client state

bdach:
RFC. This is the first proposed step to addressing https://github.com/ppy/osu-server-spectator/issues/244.

Up until now, the score that was later used for replay upload was part of `SpectatorClientState`, which is entirely dropped on a disconnection. This change moves the score from that transient user state to a dedicated memory buffer with timed expiry.

This means that if a user reconnects to the spectator hub, they are allowed to continue recording frames to the same replay as long as they still have the relevant score token.

Note that this does nothing about any frame bundles that potentially didn't reach spectator server at all, it just allows to continue recording if a reconnection occurs. I intend to address the dropped frames later, but that will require this PR anyway *on top* of actual protocol changes.

As for technicalities:
- This introduces a new entity store with full locking. Initially I thought it may be overkill because of the overarching locking in `SpectatorHub` but the score eviction loop on a separate thread does kind of necessitate locking for correct operation.
- The timeout is arbitrarily set to 5 minutes. As per inline comment anything below 110 seconds probably doesn't make sense. This is a complete shot in the dark and a tradeoff between reliability and memory footprint. I added some ddog counters to instrument this, maybe we can tweak later as required.
- The score buffer is roped into graceful shutdown. It made sense to me because most of everything in the replay upload flow is. It should *not* cause graceful shutdown delays unless I messed something up.


osu.Server.Spectator/Hubs/ScoreBuffer.cs
```diff
@@ -0,0 +1,154 @@
...
+
+        public async Task UpdateAsync(long scoreTokenId, FrameDataBundle data)
+        {
+            using (var usage = await store.TryGetForUse(scoreTokenId))
+            {
+                var buffered = usage?.Item;
+
+                if (buffered == null)
+                {
+                    usage?.Destroy();
```
peppy: Trying to understand how this can ever be the case. Is this just an assert equivalent, or is there some valid case where this will get hit?

If so, I may suggest writing this as

```
if (usage == null)
 return;

Debug.Assert(usage.Item != null);
```

because it clarifies the null case better.

peppy (commented):
Looks fine apart from one minor consideration.

osu.Server.Spectator/Hubs/ScoreBuffer.cs
bdach: https://github.com/ppy/osu-server-spectator/pull/503/commits/2be3426cba4bac2ce05c3e9672690b4a822f1615

peppy (approved):

PR #504, opened 2026-05-13
Ranked Play: Add base damage and individual multipliers

smoogipoo:
- [x] Depends on https://github.com/ppy/osu/pull/37740

This is a pretty drastic change to the scoring system:

- Global multiplier reduced to `0.5` and increases by `0.5` every round.
- Individual multiplier starts at `0.5` and increases by `0.5` for the winning player.
- The global and individual multipliers are added together when dealing damage.
- Bonus damage is set to a conservative `50000` per win. This is non-multiplied and applies on top of the attack damage.
- Failing to download the beatmap now causes `100000` damage multiplied by the global multiplier.

Because this is all frankensteined into the `RankedPlayDamageInfo` model, this change _is_ backwards compatible with existing clients, except that:
- They don't display the individual multipliers.
- The damage may be a bit confusing, because if there's only a 1000 score difference the client will count up to 51000.

### DIFF
--- a/osu.Server.Spectator/Hubs/Multiplayer/Matchmaking/RankedPlay/RankedPlayMatchController.cs
+++ b/osu.Server.Spectator/Hubs/Multiplayer/Matchmaking/RankedPlay/RankedPlayMatchController.cs
@@ -8,7 +8,6 @@ using System.Threading.Tasks;
 using Microsoft.AspNetCore.Mvc;
 using OpenSkillSharp.Models;
 using OpenSkillSharp.Rating;
-using osu.Framework.Utils;
 using osu.Game.Online.Multiplayer;
 using osu.Game.Online.Multiplayer.MatchTypes.RankedPlay;
 using osu.Game.Online.RankedPlay;
@@ -335,54 +334,26 @@ namespace osu.Server.Spectator.Hubs.Multiplayer.Matchmaking.RankedPlay
         /// Causes a player to take damage.
         /// </summary>
         /// <param name="userId">The user ID of the player taking damage.</param>
-        /// <param name="attackDamage">Amount of damage dealt from an attack.</param>
-        /// <param name="attackMultiplier">Scales the <paramref name="attackDamage"/> damage.</param>
-        /// <param name="bonusDamage">Amount of damage dealt from any other source. Does not scale with <paramref name="attackMultiplier"/>.</param>
+        /// <param name="directDamage">Direct amount of damage before any multipliers are added.</param>
+        /// <param name="multiplier">A multiplier of <paramref name="directDamage"/>.</param>
+        /// <param name="bonusDamage">Damage dealt for winning a round. Does not scale with <paramref name="multiplier"/>.</param>
         /// <returns>A descriptor for the damage taken.</returns>
-        public RankedPlayDamageInfo Damage(int userId, int attackDamage = 0, double attackMultiplier = 1, int bonusDamage = 0)
+        public RankedPlayDamageInfo Damage(int userId, int directDamage = 0, double multiplier = 1, int bonusDamage = 0)
         {
             RankedPlayUserInfo userInfo = State.Users[userId];
-            RankedPlayDamageInfo damageInfo = new RankedPlayDamageInfo();
 
-            if (attackDamage != 0)
+            int totalDamage = (int)Math.Ceiling(directDamage * multiplier) + bonusDamage;
+
+            RankedPlayDamageInfo damageInfo = new RankedPlayDamageInfo
             {
-                damageInfo.Sources.Add(new RankedPlayDamageSource
-                {
-                    Type = RankedPlayDamageType.Attack,
-                    RawValue = attackDamage,
-                    Damage = attackDamage
-                });
-
-                if (!Precision.AlmostEquals(1, attackMultiplier))
-                {
-                    damageInfo.Sources.Add(new RankedPlayDamageSource
-                    {
-                        Type = RankedPlayDamageType.Multiplier,
-                        RawValue = attackMultiplier,
-                        Damage = (int)Math.Ceiling(attackDamage * (attackMultiplier - 1))
-                    });
-                }
-            }
-
-            if (bonusDamage != 0)
-            {
-                damageInfo.Sources.Add(new RankedPlayDamageSource
-                {
-                    Type = RankedPlayDamageType.Bonus,
-                    RawValue = bonusDamage,
-                    Damage = bonusDamage
-                });
-            }
-
-            foreach (var source in damageInfo.Sources)
-            {
-                damageInfo.Damage += source.Damage;
-                if (source.Type != RankedPlayDamageType.Multiplier)
-                    damageInfo.RawDamage += source.Damage;
-            }
-
-            damageInfo.OldLife = userInfo.Life;
-            damageInfo.NewLife = Math.Max(damageInfo.OldLife == 1_000_000 ? 1 : 0, damageInfo.OldLife - damageInfo.Damage);
+                RawDamage = directDamage + bonusDamage,
+                Damage = totalDamage,
+                OldLife = userInfo.Life,
+                NewLife = Math.Max(userInfo.Life == 1_000_000 ? 1 : 0, userInfo.Life - totalDamage),
+                DirectDamage = directDamage,
+                Multiplier = multiplier,
+                BonusDamage = bonusDamage,
+            };
 
             userInfo.Life = damageInfo.NewLife;
 
--- a/osu.Server.Spectator/Hubs/Multiplayer/Matchmaking/RankedPlay/RankedPlayMatchController.cs
+++ b/osu.Server.Spectator/Hubs/Multiplayer/Matchmaking/RankedPlay/RankedPlayMatchController.cs
@@ -8,6 +8,7 @@ using System.Threading.Tasks;
 using Microsoft.AspNetCore.Mvc;
 using OpenSkillSharp.Models;
 using OpenSkillSharp.Rating;
+using osu.Framework.Utils;
 using osu.Game.Online.Multiplayer;
 using osu.Game.Online.Multiplayer.MatchTypes.RankedPlay;
 using osu.Game.Online.RankedPlay;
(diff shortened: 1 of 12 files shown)


osu.Server.Spectator/Hubs/Multiplayer/Matchmaking/RankedPlay/Stages/ResultsStage.cs
```diff
@@ -53,24 +60,36 @@ protected override async Task Begin()
...
-            if (winningScores.Length == 1)
-                State.Users[(int)winningScores.Single().user_id].RoundsWon += 1;
+                int attackDamage = winningTotalScore - (int)losingScore.total_score;
+                double attackMultiplier = State.DamageMultiplier + State.Users[winningUserId.Value].DamageMultiplier;
+
+                State.Users[(int)losingScore.user_id].DamageInfo = Controller.Damage((int)losingScore.user_id, attackDamage, attackMultiplier, BaseDamage);
+            }
+            else
+            {
+                // Tie: both players take the base amount of damage.
```
IceDynamix: assuming both players are on 1hp, this would make it possible to make the match a tie

osu.Server.Spectator/Hubs/Multiplayer/Matchmaking/RankedPlay/RankedPlayMatchController.cs
```diff
@@ -334,28 +335,58 @@ public async Task ActivateCard(RankedPlayCardItem card)
         /// Causes a player to take damage.
         /// </summary>
         /// <param name="userId">The user ID of the player taking damage.</param>
-        /// <param name="amount">The amount of damage (before any multipliers are added) to take.</param>
+        /// <param name="attackDamage">Amount of damage dealt from an attack.</param>
+        /// <param name="attackMultiplier">Scales the <paramref name="attackDamage"/> damage.</param>
+        /// <param name="bonusDamage">Amount of damage dealt from any other source. Does not scale with <paramref name="attackMultiplier"/>.</param>
         /// <returns>A descriptor for the damage taken.</returns>
-        public RankedPlayDamageInfo Damage(int userId, int amount)
+        public RankedPlayDamageInfo Damage(int userId, int attackDamage = 0, double attackMultiplier = 1, int bonusDamage = 0)
```
IceDynamix: "bonus" damage is colloquially called "flat" damage, established in rpgs and other genres. would make it more clear that bonus dmg is not multiplied. but changing the variable name is obviously not that important

osu.Server.Spectator/Hubs/Multiplayer/Matchmaking/RankedPlay/Stages/FinishCardPlayStage.cs
```diff
@@ -40,7 +40,7 @@ protected override async Task Finish()
                 // Subtract 100K HP from every player that failed to load the beatmap in time.
                 // Although this seems unfair, it means that players are not able to purposefully block the others' picks.
                 foreach (var player in Room.Users.Where(p => p.BeatmapAvailability.State != DownloadState.LocallyAvailable))
```
IceDynamix: although i didn't notice before, would make it possible to create a tie when both players are on 1hp and fail to download.

IceDynamix (commented):
allowing ties would damage one of the assumptions of the PlackettLuce model (in short: no ties results in simpler code).

to reason why ties should not just result in +0 -0 rating, consider the situation where a 1000 rated player ties against a 2000 rated player, which would be *very* impressive for the 1000 rated player given the rating gap, so a model that accounts for ties will change rating if the rating difference is large enough.

we are at a crossroads: i need an executive decision on whether you want to allow ties or not. if you do, then give me a bit of time to research the other models. can then be combined with the pr to update model parameters (beta, tau, gamma).

osu.Server.Spectator/Hubs/Multiplayer/Matchmaking/RankedPlay/RankedPlayMatchController.cs
pacowoc: > "bonus" damage is colloquially called "flat" damage, established in rpgs and other genres. would make it more clear that bonus dmg is not multiplied. but changing the variable name is obviously not that important

Agree

smoogipoo: Tbh, I don't think this is working anywhere near like traditional RPGs, because we have to bundle it for the client and last stand calculations. Ideally the bonus parameter would not exist at all and we would just call this method multiple times. I haven't figured that part out yet, but I am pained by it too.

"bonus" feels better to me, because really, "attackDamage" is/should be the flat damage.

smoogipoo:
Ties are fine, however I also agree with your above assertions that ties should be avoided as much as possible. So add91df97442e7b432836c362555e949472ed040:
- Removes damage dealt when no player readies up. This could technically deadlock a game if both players perpetually fail to ready up, but this is an edge case that resolves itself after all 50 cards run out.
- Removes damage dealt when there is no clear winner to a round. This could technically deadlock a game if both players perpetually get 0 score, but this is also an edge case that will resolve after all 50 cards run out.

The client already has support for showing a drawn end screen, and +0/-0 makes the most sense for osu!'s nature where draws are _not_ incentivised (as opposed to chess).

bdach:
@smoogipoo merge conflict resolution attempted in 1b867dec39289e79f5ad7ea4473b54058d731b55, please double check

osu.Server.Spectator/Hubs/Multiplayer/Matchmaking/RankedPlay/RankedPlayMatchController.cs
```diff
@@ -334,28 +335,58 @@ public async Task ActivateCard(RankedPlayCardItem card)
...
-            int oldLife = userInfo.Life;
-            // Last-stand mechanic: fatal attacks at 100% HP bring players 1HP.
-            int newLife = Math.Max(oldLife == 1_000_000 ? 1 : 0, oldLife - damage);
+                if (!Precision.AlmostEquals(1, attackMultiplier))
+                {
+                    damageInfo.Sources.Add(new RankedPlayDamageSource
+                    {
+                        Type = RankedPlayDamageType.Multiplier,
+                        RawValue = attackMultiplier,
+                        Damage = (int)Math.Ceiling(attackDamage * (attackMultiplier - 1))
```
bdach: Is it actually ensured that `attackMultiplier > 1` always? Because it seems to me that that's not necessarily so. One case where it wouldn't be is the 100k damage from failing to load which then gets just `State.DamageMultiplier` applied: https://github.com/ppy/osu-server-spectator/blob/1b867dec39289e79f5ad7ea4473b54058d731b55/osu.Server.Spectator/Hubs/Multiplayer/Matchmaking/RankedPlay/Stages/FinishCardPlayStage.cs#L41 which in the first round could be 0.5x. Which, while the end result will still be *correct*, I imagine will make things look pretty silly in the client because you'll have a source dealing *negative* damage.

Solution would be to either apply both multipliers in the 100k load fail case (which would make it true that the multiplier is always above 1), or to hardcode 1x multiplier in the fail case, or whatever else. Dunno.

osu.Server.Spectator/Hubs/Multiplayer/Matchmaking/RankedPlay/Stages/RoundWarmupStage.cs
```diff
@@ -29,7 +29,10 @@ protected override async Task Begin()
             }
 
             State.CurrentRound++;
-            State.DamageMultiplier = computeDamageMultiplier(State.CurrentRound);
+
+            // Increase tension by increasing the global multiplier.
+            if (State.CurrentRound > 1)
+                State.DamageMultiplier += 0.5;
```
bdach, line 35: I'm guessing the removal of per-ruleset multiplier increment considerations is that introducing the base damage and per-player multiplier should offset it.

osu.Server.Spectator/Hubs/Multiplayer/Matchmaking/RankedPlay/Stages/ResultsStage.cs
```diff
@@ -53,24 +60,30 @@ protected override async Task Begin()
...
-            }
+                // Winner: losing player takes damage.
+                SoloScore losingScore = scores.Single(u => u.user_id != winningUserId);
+
+                int attackDamage = winningTotalScore - (int)losingScore.total_score;
+                double attackMultiplier = State.DamageMultiplier + State.Users[winningUserId.Value].DamageMultiplier;
 
-            SoloScore[] winningScores = scores.Where(u => u.total_score == maxTotalScore).ToArray();
-            if (winningScores.Length == 1)
-                State.Users[(int)winningScores.Single().user_id].RoundsWon += 1;
```
bdach, line 73: I feel like you forgot to add this back in.

osu.Server.Spectator/Hubs/Multiplayer/Matchmaking/RankedPlay/RankedPlayMatchController.cs
smoogipoo: I've responded here, where it's more contextually relevant after recent changes: https://github.com/ppy/osu/pull/37740#discussion_r3246955970

osu.Server.Spectator/Hubs/Multiplayer/Matchmaking/RankedPlay/Stages/ResultsStage.cs
smoogipoo: Indeed, added back in https://github.com/ppy/osu-server-spectator/pull/504/changes/35c90dd489b92b3b7ebbb330a0abb511324b4072

osu.Server.Spectator/Hubs/Multiplayer/Matchmaking/RankedPlay/Stages/RoundWarmupStage.cs
smoogipoo: I figure that this is a new system and we probably want to play it out by ear for a little while. I wouldn't be surprised to see it return at some point.

peppy (approved):

bdach:
Seems fine, ~~unless the chosen solution to address the [game-side concern](https://github.com/ppy/osu/pull/37740#discussion_r3256585808) involves reverting the model changes, in which case a recheck is probably needed.~~

bdach (approved):

peppy (approved):

PR #505, opened 2026-05-14
Ranked Play: Only activate last stand at 100% HP

smoogipoo:
This will conflict with https://github.com/ppy/osu-server-spectator/pull/504, but is arguably more important.

There's been a lot of confusion about this mechanic, here's a small sampling:
- https://www.reddit.com/r/osugame/comments/1tcke0l/what_is_the_point_of_this_mechanic/
- https://www.reddit.com/r/osugame/comments/1tc19b6/the_craziest_thing_ive_seen_in_ranked_play/
- https://www.reddit.com/r/osugame/comments/1tc0tt4/my_friend_just_hit_1_hp_in_ranked/
- https://osu.ppy.sh/community/forums/topics/2198397?n=241

This is... not how I expected it to behave. It shouldn't activate every time, rather it should be for rare situations. The choice of 1M is because any other value would take explaining as to how it works as well as probably a more complex design. Maybe we don't have to explain things if they only occur in the rarest of situations.


bdach (approved):
Seems to do what it says

PR #508, opened 2026-05-17
Pool ranked play based on normally distributed difficulty

IceDynamix:
The expectation of the bell curve in the ranked play intro sequence implies that the maps are picked normally distributed based on their difficulty. Current algorithm generates a normally-distributed number for every map based on map rating, then picks the maps with the highest value. This overweights rating ranges that have more maps. 

Example:
- 1300 rating: 50 maps
- 1500 rating: 10 maps

When pooling for 1400 rating, this example would have a 5x higher chance of pooling 1300 rated maps.

This new implementation follows the expected bell curve, by picking a random rating normally distributed, and _then_ picking the map closest to the generated rating.

Parameters:
- `rating_sig = 100`: Variance in context of normal distribution.

Currently set to 100, but could revert back to logistic curve from previous implementation, or simply increase to widen the breadth of ratings generated. Have to do some testing.

### DIFF
--- a/osu.Server.Spectator/Hubs/Multiplayer/Matchmaking/Queue/MatchmakingBeatmapSelector.cs
+++ b/osu.Server.Spectator/Hubs/Multiplayer/Matchmaking/Queue/MatchmakingBeatmapSelector.cs
@@ -147,18 +147,45 @@ namespace osu.Server.Spectator.Hubs.Multiplayer.Matchmaking.Queue
             // Pick from maps around the minimum rating.
             double userRatingMu = ratings.Select(r => r.Mu).DefaultIfEmpty(1500).Min();
 
-            // Logistic curve that is narrow (50pts) at lower ratings where there are many beatmaps,
-            // and broader (100pts) at higher ratings where there are fewer beatmaps.
-            double ratingSig = 50 + 50 / (1 + Math.Exp(-0.01 * (userRatingMu - 1800)));
+            const double rating_sig = 100;
 
-            return beatmaps.Values.OrderByDescending(b =>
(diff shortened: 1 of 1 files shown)


pacowoc:
You can use the built-in `Array.BinarySearch` to optimize the time complexity, you don’t need to code it yourself.

pacowoc:
Also, there is a potential issue causing some maps in the same elo to appear way less than other maps.

Consider the case:

Map A: 1701
Map B: 1705
Map C: 1711
Map D: 1712
Map E: 1713

Map B would be chosen if the random number is between 1703 and 1708; but map D would only be chosen if the rating is between 1711.5 and 1712.5, leading to potentially unbalanced pooling

IceDynamix:
Map ratings at this scale change from just a single match, they will continuously fluctuate closer and further away over time. The situation in question won't stay that way for more than a day. Compare that to the previous implementation, where unbalanced distribution will take a long time to balance out (if ever).

pacowoc:
My main concern is that at least some of the maps will pop out repeatedly just due to random fluctuations eventually resulting in larger rating gaps. The maps in this situation will certainly be different every day but it will dominate momentarily in every new lobby active around the same time.

smoogipoo:
The biggest advantage of this in my eyes, is that it's more stable in cases where there aren't that many maps. Notice the teal colour in both of the below graphs:

| master | PR |
| --- | --- |
| image | image |

I still don't like the harsh drop to 0 for 2100/2400, but I suppose we'll cross that path with another change later on.

smoogipoo (approved):

PR #509, opened 2026-05-17
Lower chance of osu! standard low elo maps that gained a lot of rating in ranked play pool

IceDynamix:
A stopgap measure to improve player retention by hiding maps that _seem_ too easy (scenario of "5* player getting 3* map"). Most 3* maps tend to be underweighted by 0.5*.

Parameters:
- `min_weight = 0.25` : Minimum pick weight a beatmap can ever reach.
- `target_center = 1300` and `target_radius = 75`: Target rating range over which to modulate the lower bound
- `max_rating_diff = 300`: How large the rating diff needs to be in order to assign the minimum weight

image

To be honest, this plays double duty in being a change I personally dislike. I don't like changes that only affect a single ruleset, and I don't like changes that alter the appearance rate of maps. But it really feels somewhat necessary to improve the future prospects of the game.

### DIFF
--- a/osu.Server.Spectator/Hubs/Multiplayer/Matchmaking/Queue/MatchmakingBeatmapSelector.cs
+++ b/osu.Server.Spectator/Hubs/Multiplayer/Matchmaking/Queue/MatchmakingBeatmapSelector.cs
@@ -146,17 +146,24 @@ namespace osu.Server.Spectator.Hubs.Multiplayer.Matchmaking.Queue
         public matchmaking_pool_beatmap[] GetAppropriateBeatmaps(int count, EloRating[] ratings)
         {
             // Pick from maps around the minimum rating.
-            double userRatingMu = ratings.Select(r => r.Mu).DefaultIfEmpty(1500).Min();
+            double targetRating = ratings.Select(r => r.Mu).DefaultIfEmpty(1500).Min();
 
             const double rating_sig = 100;
 
             HashSet<matchmaking_pool_beatmap> maps = [];
 
-            foreach (double rating in randomNumberSamples(count, userRatingMu, rating_sig))
+            foreach (double rating in randomNumberSamples(count, targetRating, rating_sig))
             {
                 // Could optimize with binary search?
                 var map = beatmaps.Values
                                   .Where(b => !maps.Contains(b))
+                                  .Where(b =>
+                                  {
+                                      double p = mapProbability(b);
+                                      if (1 <= p) return true;
+
+                                      return Random.Shared.NextDouble() <= p;
+                                  })
                                   .MinBy(b => Math.Abs(b.rating - rating));
 
                 if (map == null)
@@ -166,6 +173,11 @@ namespace osu.Server.Spectator.Hubs.Multiplayer.Matchmaking.Queue
             }
 
             return maps.ToArray();
+        }
+
+        private static double mapProbability(matchmaking_pool_beatmap map)
+        {
+            return 1.0;
         }
 
         private static IEnumerable<double> randomNumberSamples(int n, double mu, double sigma)
--- a/osu.Server.Spectator/Hubs/Multiplayer/Matchmaking/Queue/MatchmakingBeatmapSelector.cs
+++ b/osu.Server.Spectator/Hubs/Multiplayer/Matchmaking/Queue/MatchmakingBeatmapSelector.cs
@@ -159,7 +159,7 @@ namespace osu.Server.Spectator.Hubs.Multiplayer.Matchmaking.Queue
                                   .Where(b => !maps.Contains(b))
                                   .Where(b =>
                                   {
-                                      double p = BeatmapProbability(b, targetRating);
+                                      double p = BeatmapWeight(b, targetRating);
                                       return p switch
                                       {
                                           >= 1 => true,
@@ -184,12 +184,12 @@ namespace osu.Server.Spectator.Hubs.Multiplayer.Matchmaking.Queue
         /// <param name="beatmap">Map to calculate probability for</param>
         /// <param name="targetRating">Rating to target the pool for</param>
         /// <returns>Number between 0 and 1 representing the probability</returns>
-        public static double BeatmapProbability(matchmaking_pool_beatmap beatmap, double targetRating)
+        public static double BeatmapWeight(matchmaking_pool_beatmap beatmap, double targetRating)
         {
-            return RatingDiffProbability(beatmap, targetRating);
+            return RatingDiffWeight(beatmap, targetRating);
         }
 
-        public static double RatingDiffProbability(matchmaking_pool_beatmap beatmap, double targetRating)
+        public static double RatingDiffWeight(matchmaking_pool_beatmap beatmap, double targetRating)
         {
             // The goal is to improve user perception of standard pools in lower elo.
             // The lower the target elo, the less likely we want to show massive positive rating outliers.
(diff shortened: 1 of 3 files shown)


osu.Server.Spectator/Hubs/Multiplayer/Matchmaking/Queue/MatchmakingBeatmapSelector.cs
```diff
@@ -145,20 +145,94 @@ .. playerScores
...
+        {
+            // The goal is to improve user perception of standard pools in lower elo.
+            // The lower the target elo, the less likely we want to show massive positive rating outliers.
+
+            const double min_target = 800;
+            const double max_target = 1500;
+            const double max_map_diff = 300;
+            const double min_probability = 0.25;
+
+            if (beatmap.playmode == 0) return 1;
```
Elvendir:
```suggestion
            if (beatmap.playmode != 0) return 1;
```
Deactivate when not osu, I think.

osu.Server.Spectator/Hubs/Multiplayer/Matchmaking/Queue/MatchmakingBeatmapSelector.cs
```diff
@@ -145,20 +145,94 @@ .. playerScores
...
+            if (targetRating >= max_target) return 1;
+
+            double originalRating = matchmaking_pool_beatmap.DifficultyRatingToEloRating(beatmap.difficultyrating);
+            if (originalRating < beatmap.rating) return 1;
```
Elvendir:
```suggestion
            if (originalRating >= beatmap.rating) return 1;
```
Using ```originalRating < beatmap.rating```  means not penalizing when  SR < SR_adjust which is the target.

osu.Server.Spectator/Hubs/Multiplayer/Matchmaking/Queue/MatchmakingBeatmapSelector.cs
```diff
@@ -145,20 +145,94 @@ .. playerScores
...
+
+            // Step 1: Determine the size of the probability interval
+            // Scale the target rating: [min_target, max_target] -> [0, 1]
+            double relativeTargetRating = Math.Clamp((max_target - beatmap.rating) / (max_target - min_target), 0, 1);
+            // Scale [0, 1] -> [min_probability, 1] (inverse)
+            double lowerBound = min_probability + (1 - min_probability) * (1 - relativeTargetRating);
```
Elvendir:
```suggestion
            // Scale the target rating: [min_target, max_target] -> [1, 0]
            double relativeTargetRating = Math.Clamp((max_target - beatmap.rating) / (max_target - min_target), 0, 1);
            // Scale [1, 0] -> [min_probability, 1] 
            double lowerBound = min_probability + (1 - min_probability) * (1 - relativeTargetRating);
```

The comments are confusing as it seems there is a typo in them.

osu.Server.Spectator/Hubs/Multiplayer/Matchmaking/Queue/MatchmakingBeatmapSelector.cs
```diff
@@ -145,20 +145,94 @@ .. playerScores
...
+
+            // Step 2: Determine what point to use in the probability interval
+            // Scale rating diff: [0, 300] -> [0, 1]
+            double relativeMapRatingDiff = Math.Clamp((beatmap.rating - originalRating) / max_map_diff, 0, 1);
+            // Scale [0, 1] -> [lowerBound, 1]  (inverse)
```
Elvendir:
```suggestion
            // Scale [0, 1] -> [1, lowerBound]
```
Would probably make it clearer than `(inverse)` next to it.

pacowoc:
You might need to slightly increase the elo window for this, people are getting full 3* hand even at 1400 elo, and the downweighting at 1400 of this pr as for now isn’t adequate for that elo I think.

IceDynamix:
Changed some parameters and methods. I added a table with values in the PR description.

Givikap120:
Can there be a better explanation of what exactly will change? Because I've reread it multiple times and I still can't understand it.

What is "seem to easy"? Is it about maps that have low star rating but are very underweighted? Why is it a problem that they're shown if they're underweighted?
If 3* is underweighted by 0.5* on average - how this affect maps that are underweighted by way more than 0.5*?
Why exactly underweighted maps should be "hidden" from the player?
What does word "hidden" even means? That they just don't get picked?

What is a target center?
Just a random parameter to base the formula off?
Or it's a real variable that gets changed.

What does minimum chance of 0.25 means? There's no way minimum probability of getting a beatmap is 25% because this would imply that we would have 4 maps at most.

Why for low rating the lower bound is so much lower than for high rating? What does this implies?
Why in 1700 rating case for 300 rating difference the lower bound is 99.6%?
Does this means that in 1700 elo match there's 99.6% probability of map below 1400 rating appearing?

IceDynamix:
> What is "seem to easy"? Is it about maps that have low star rating but are very underweighted? Why is it a problem that they're shown if they're underweighted?
Why exactly underweighted maps should be "hidden" from the player?
What does word "hidden" even means? That they just don't get picked? 

this is change is mostly about player retention and less about actual balancing of the pool for competitiveness. many players are choosing not to play ranked play because "the star rating is too low" and they get many "dead hands" where they are forced to pick low sr / low ar.

the majority of players really don't like playing 3* maps. there are multiple reasons for this. as mentioned, many 3* maps are rather underweighted, but also lower relative quality because of ranking criteria forcing mappers to make hard diffs for spreads, players not being a big fan of low ar, and finally players at this skill range preferring to play closer to their skill ceiling rather than floor.

the expectation for "5* players" (which means they usually play maps that have somewhat accurate or slightly overrated star rating at 5*) is to play 5* maps. players feel "disrespected" when the game give them a majority of 3* maps, _even if_ the difficulty is actually correct for the lobby (expected 550k from dynamic beatmap rating). there is no ui in the frontend that shows the dynamic rating after all and to be honest i'm not 100% sure whether its a good idea to add.

the idea of this pull request is to lower the appearance rate of maps that are much harder than their star rating would suggest. this wont actually change the average *true* difficulty of the pool, just the visual aspect of less maps with lower star rating.

> If 3* is underweighted by 0.5* on average - how this affect maps that are underweighted by way more than 0.5*?

the "initial elo" of a map is calculated from its star rating. we calculate the "underweightedness" of sr based on the elo difference from the current elo and the initial elo. rating gain would imply that the initial star rating is too low. `max_rating_diff = 300` defines the scale of rating difference we want to work with. 300 rating is _a lot_.

> What does word "hidden" even means? That they just don't get picked?
> What does minimum chance of 0.25 means? There's no way minimum probability of getting a beatmap is 25% because this would imply that we would have 4 maps at most.

i renamed the variables from "probability" to "weight" to avoid exactly this confusion. i forgot to fully update the wording in the pull request description, that's on me.

a map with weight 0.5 has half of the chance of being picked than a map with 1 weight. so in a pool where 50% of maps have weight 1 and 50% have weight 0.5, you'll see 2/3 of maps of weight 1 and 1/3 of maps of weight 0.5. it's a relative value rather than an absolute one.

> What is a target center? Just a random parameter to base the formula off? Or it's a real variable that gets changed.

it's a parameter to define a [logistic function](https://en.wikipedia.org/wiki/Logistic_function). the shape of a logistic function looks like this. `target_center` describes at which point the mid-point of the logistic function is, and `target_radius` describes how wide to make the logistic function.

image

> Why for low rating the lower bound is so much lower than for high rating? What does this implies? Why in 1700 rating case for 300 rating difference the lower bound is 99.6%?

we want low elo lobbies to reduce the appearance of underweighted maps. but we don't want this to happen in higher elo lobbies. as the elo goes up, we want to increase the lower bound, so underweighted maps slowly reach the appearance rate of all other maps. the closer the lower bound is to 1, the closer the pooling will be as if this change didn't exist at all.

> Does this means that in 1700 elo match there's 99.6% probability of map below 1400 rating appearing?

no. it just means that no matter how large the rating difference, the underweighted map will have almost the same appearance rate as any other map.

Givikap120:
Okay, to summarize
This change consists of two parts:

First part is decreasing the probability of "underweighted" maps appearing in up to 4 times, where the higher is the "underweightness" (rating difference between SR-based elo and actual elo) - the more probability gets decreased, reaching maximum of 75% probability decrease on 300 elo difference.
This behaviour defined by p_min (0.25) and rating difference (300) for p_min.

Second part makes sure that the change is applied only to lower rating lobbies.
We add additional weighting function to our penalty that makes the penalty decrease via the logistic function as rating goes up.
So on 800 rating the penalty is almost full (0.251 instead of 0.25), whereas on 1700 rating it's almost turned off (0.996).
And the weight is defined by logistic curve parameters such as 1300 for the center of the curve, and 75 for it's steepness.

Is that right?

IceDynamix:
yes that is roughly correct

PR #510, opened 2026-05-18
Add timeout preventing recurrent matchups

smoogipoo:
This makes it so that players cannot be paired within 10 minutes apart. In real-world cases, this will prevent mrekk and Ekoro constantly getting paired.

---

@peppy For staging, please set environment variable `MATCHMAKING_RECENT_MATCHUP_TIMEOUT=0`

Also I recognise the dual split of `memoryCache` between `MatchmakingQueue` and `MatchmakingQueueBackgroundService` is kind of weird. There's two ways of thinking about this:
1. I can move `queue_ban_end_time` (currently in MQBS) into MQ, thereby making it per-pool rather than global. Or...
2. You can think of it as a global ban from queueing altogether, to prevent other abuse.


osu.Server.Spectator/Hubs/Multiplayer/Matchmaking/Queue/MatchmakingQueue.cs
```diff
@@ -222,6 +241,27 @@ public MatchmakingQueueUpdateBundle MarkInvitationDeclined(MatchmakingQueueUser
...
 
+        /// <summary>
+        /// Marks a recent matchup between two users.
+        /// </summary>
+        /// <param name="userId">The first user.</param>
+        /// <param name="opponentId">The second user.</param>
+        public void MarkRecentMatchup(int userId, int opponentId)
+        {
+            cache.Set(recent_matchup(userId, opponentId), true, Clock.UtcNow + RecentMatchupTimeout);
+            cache.Set(recent_matchup(opponentId, userId), true, Clock.UtcNow + RecentMatchupTimeout);
```
peppy: How about just sorting ascending and storing once?

peppy (approved):

PR #511, opened 2026-05-18
Fix duels not appearing in recent history

smoogipoo:
The case where a lobby _does_ exist but a queue _does not_ exist is possible in exactly one scenario: duels. For duels, the queues are ephemeral for the duration of the queue.


peppy:
Hmm, do we want this though? For personal history maybe, but not sure about global?

smoogipoo:
Feel free to close, just though this would maybe be considered a regression.

I agree it may not be what we want.

peppy:
Let's leave be for now, I think adding them in will just cause confusion ("why were those two players matchmade?" "why are the same two players getting multiple matches in a row? i thought that was changed" etc)

PR #513, opened 2026-05-22
Add server-side support for slots in multiplayer rooms

### LINKED ISSUE #405: Add back support for slots in multiplayer rooms
They appear to be significantly important to the tournament crowd that the lack of them is a major drawback to running team tournaments on lazer.

I asked and extra sorting criteria are not likely to help here, people want to enforce their own standards like e.g. top seed in a bracket always having the topmost slot. Enforcing this in arbitrary tournament scenarios isn't feasible.

Of note we don't want to introduce a cap on user count of multiplayer rooms, either. This is probably something that will only apply to selected rooms. I'm not sure of the specifics here, but it'll probably be something like "slots only exist when the room has a limited participant count".

The scope here is very vague, I'm not entirely sure how anything and everything will work here, I'll probably have to start with a PoC and see where things end up.

### LINKED ISSUE #18897: Add better support for tournament management
This covers requirements such as having a referee in a multiplayer room with access to all the commands they need to make a tournament happen, along with ample output of match results and the various metadata that has been consumed via IRC until now.

bdach:
- [x] Depends on NuGet package containing https://github.com/ppy/osu/pull/37741
- Closes https://github.com/ppy/osu-server-spectator/issues/405
- Closes https://github.com/ppy/osu/issues/18897

This includes:

- Enforcement of maximum participant count limit
- Maintenance of slots in room and implementation of related operations (changing slots, locking room)
- New referee API operations for manipulating slots

The best specification for how the slots were intended to work is probably the test suite added here (`MaxParticipantsAndSlotsTest`) so I'd rather recommend reading that instead than restate the tests here.


smoogipoo (approved):
I have stylistic concerns with the implementation in `StandardMatchController`, but I think this is passable.

bdach:
> I have stylistic concerns with the implementation in `StandardMatchController`

Do you mind elaborating on what those are?

osu.Server.Spectator/Hubs/Multiplayer/Standard/StandardMatchController.cs
```diff
@@ -115,24 +130,174 @@ public virtual async Task HandleUserRequest(MultiplayerRoomUser user, MatchUserR
...
+            // participant limit has been unset, no slots in state => nothing to do
+            if (room.Settings.MaxParticipants == null && State.Slots == null)
+                return false;
+
+            // participant limit has been unset, slots in state => discard slots
+            if (room.Settings.MaxParticipants == null && State.Slots != null)
+            {
+                State.Slots = null;
+                return true;
+            }
```
smoogipoo: Since `MaxParticipants == null` is a precondition to the rest of the method, I would write this as:

```
if (MaxParticipants == null)
{
    if (State.Slots == null)
    { }
    else
    { }
}
```

osu.Server.Spectator/Hubs/Multiplayer/Standard/StandardMatchController.cs
```diff
@@ -115,24 +130,174 @@ public virtual async Task HandleUserRequest(MultiplayerRoomUser user, MatchUserR
...
+
+            // at this point max participants are guaranteed to be not null
```
smoogipoo: Sure, but why do all of the following code blocks also have `if (MaxParticipants != null)`?

osu.Server.Spectator/Hubs/Multiplayer/Standard/StandardMatchController.cs
```diff
@@ -115,24 +130,174 @@ public virtual async Task HandleUserRequest(MultiplayerRoomUser user, MatchUserR
...
+                    {
+                        int removedSlotWithUser = Array.FindIndex(oldSlots, State.Slots.Length, j => j != null);
+
+                        if (removedSlotWithUser > 0)
+                        {
+                            State.Slots[i] = oldSlots[removedSlotWithUser];
+                            oldSlots[removedSlotWithUser] = null;
+                        }
+                    }
+                }
```
smoogipoo, line 231: It seems a little bit weird to me that in the path with `Slots == null`, we use `GetNextBestSlot()`, but in this path we're just moving users into arbitrary slots.

I feel like this could be done simply by iterating from `i = State.Slots.Length; i < oldSlots.Length`, and we wouldn't have to do this juggling of `oldSlots[...] = null` and the pseudo-assert below.

osu.Server.Spectator/Hubs/Multiplayer/Standard/StandardMatchController.cs
```diff
@@ -115,24 +130,174 @@ public virtual async Task HandleUserRequest(MultiplayerRoomUser user, MatchUserR
...
+                    if (oldSlots[i] != null)
+                    {
+                        State.Slots[i] = oldSlots[i];
+                        oldSlots[i] = null;
+                    }
+                    else
+                    {
+                        int removedSlotWithUser = Array.FindIndex(oldSlots, State.Slots.Length, j => j != null);
+
+                        if (removedSlotWithUser > 0)
```
smoogipoo: Would prefer this to be `>= 0` since, 0 is a valid "this thing exists" value. Even if it's not technically possible.

smoogipoo (commented):
These are my stylistic concerns. All nits that don't need to be addressed.

osu.Server.Spectator/Hubs/Multiplayer/Standard/TeamVersusMatchController.cs
```diff
@@ -106,8 +94,25 @@ private int getBestAvailableTeam()
...
+
+            int teamSize = slots.Length / State.Teams.Count;
+            int teamStartIndex = userState.TeamID * teamSize;
+
+            int nextEmptySlotInTeam = Array.FindIndex(slots, teamStartIndex, teamSize, item => item == null);
+            if (nextEmptySlotInTeam > 0)
+                return nextEmptySlotInTeam;
+
+            return base.GetNextBestSlot(user, slots);
+        }
```
peppy, line 110:
```suggestion
        protected override int GetNextBestSlot(MultiplayerRoomUser user, int?[] slots)
        {
            if (user.MatchState is TeamVersusUserState teamUserState)
            {
                int teamSize = slots.Length / State.Teams.Count;
                int teamStartIndex = teamUserState.TeamID * teamSize;

                int nextEmptySlotInTeam = Array.FindIndex(slots, teamStartIndex, teamSize, item => item == null);
                if (nextEmptySlotInTeam > 0)
                    return nextEmptySlotInTeam;
            }

            return base.GetNextBestSlot(user, slots);
        }
```

peppy:
I've checked all the other files and have no issues, but did come up with some possibly overlapping concerns in `StandardMatchController` so will wait for response.

osu.Server.Spectator/Hubs/Multiplayer/Standard/StandardMatchController.cs
bdach: Adjusted in https://github.com/ppy/osu-server-spectator/pull/513/commits/632dc1ca6f9e08917d03722fa3ccc6447aea74fb. For whatever reason the compiler is not able to apply the law of excluded middle here, even with the `MaxParticipant == null` guard split to top level. So I still have to leave an assert behind to satisfy the compiler.

peppy (approved):
Looks good to me.

PR #516, opened 2026-05-29
Disable requeueing users when cancelling duels 

smoogipoo:
If a player cancels an active duel it should no longer:
- Globally ban them from matchmaking.
- Return the other person to the queue (in which they'll be the permanent sole searching user).


peppy (approved):
Seems logical

PR #518, opened 2026-06-10
Slowly transition to Raw Beatmap Rating for lower elo pooling

pacowoc:
Players of lower brackets usually expect a hand that is somewhat consistent in star rating, this pr addresses this expectation by weighting maps in lower elos with a combination of the dynamic elo and the initial elo rating assigned to the beatmap. This is to reduce complaints such as "Why did I see 3* maps everywhere in my 4.5* pool" or "Why is there a random 6.9* in my 4.8* pool" without breaking competitive integrity of higher elo pools.
With this change, maps that have either gained or lost (a great majority of the maps in this range are highly positive in elo adjustment due to lower approach rates) a lot of rating will have their priority lowered because the system will slowly prioritize the gap in raw (unadjusted) beatmap rating over DMR as the pool rating decreases.

**How does the math work**
This system weights the original and adjusted map elo of a map with a weight that follows a logistic relationship to the pool rating and use the weighted value for pooling.

image
The weight of dynamic rating in pooling algorithm

For example, when generating a pool for a 1450 elo match, since the weight at 1450 is 0.5

- A 1200 map buffed to 1400 will be seen as 1300 in the eyes of the system
- A 1500 map that is neither buffed or nerfed will still be seen as 1500
- A 1700 map that is nerfed to 1500 will be seen as 1600

As the pool rating decreases, the algorithm will lean more and more towards the original star rating of the map instead of the adjusted one. I know that this might cause maps like the infamous "2* map that has 2000 DMR" to appear in real 2* pools but it is an insignificant portion of the pool anyways.


Layendan:
Subjective change, but I would make it 0.95+ weighting at 1600-1650.

1650 being the top 10% currently which is equivalent to ~diamond in other games.

pacowoc:
> Subjective change, but I would make it 0.95+ weighting at 1600-1650.
> 
> 1650 being the top 10% currently which is equivalent to ~diamond in other games.

changed

PR #520, opened 2026-06-26
Request frames not received from client on play end

bdach:
- [ ] Depends on / server-side counterpart to https://github.com/ppy/osu/pull/38163

Refer to above for the general gist of what this is.

For the server-side part specifically, it may be desired to have some sort of metric on how often this reconciliation flow actually results in needing to re-send frames. Listening to suggestions.

### DIFF
--- a/SampleSpectatorClient/Program.cs
+++ b/SampleSpectatorClient/Program.cs
@@ -44,7 +44,8 @@ namespace SampleSpectatorClient
                         new[]
                         {
                             new LegacyReplayFrame(i, RNG.Next(0, 512), RNG.Next(0, 512), ReplayButtonState.None)
-                        }));
+                        },
+                        null));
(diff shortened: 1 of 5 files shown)


peppy:
This side looks pretty good.

> For the server-side part specifically, it may be desired to have some sort of metric on how often this reconciliation flow actually results in needing to re-send frames. Listening to suggestions.

I do think we'd want to track this yes. First globally (to make sure nothing is going wrong with the system) and potentially per-user or per-score at a later date (to make sure there's no funny business).

bdach:
Blocked for now due to reasons described in https://github.com/ppy/osu/pull/38163#issuecomment-4853215905.

bdach:
See https://github.com/ppy/osu/pull/38163#issuecomment-5177564616

PR #522, opened 2026-07-02
Add accessibility mods to ranked play

smoogipoo:
- [ ] Depends on https://github.com/ppy/osu/pull/38209 for `IMultiplayerClient` model change.

This change has two purposes:
1. Applying mods to users in the room (as expected).
2. Changing the results stage to only care about `TotalScoreWithoutMods`.

### DIFF
--- a/osu.Server.Spectator/Database/Models/SoloScore.cs
+++ b/osu.Server.Spectator/Database/Models/SoloScore.cs
@@ -69,6 +69,7 @@ namespace osu.Server.Spectator.Database.Models
             BuildID = build_id,
             Passed = passed,
             TotalScore = total_score,
+            TotalScoreWithoutMods = ScoreData.TotalScoreWithoutMods ?? 0,
             Accuracy = accuracy,
             UserID = (int)user_id,
             MaxCombo = (int)max_combo,
@@ -84,5 +85,7 @@ namespace osu.Server.Spectator.Database.Models
             PP = pp,
             HasReplay = has_replay
         };
+
+        public long TotalScoreWithoutMods => ScoreData.TotalScoreWithoutMods ?? total_score;
     }
 }
--- a/osu.Server.Spectator/Hubs/Multiplayer/Matchmaking/Queue/MatchmakingQueueBackgroundService.cs
+++ b/osu.Server.Spectator/Hubs/Multiplayer/Matchmaking/Queue/MatchmakingQueueBackgroundService.cs
@@ -18,6 +18,7 @@ using osu.Game.Online.Matchmaking.Requests;
 using osu.Game.Online.Matchmaking.Responses;
 using osu.Game.Online.Multiplayer;
 using osu.Game.Online.Multiplayer.MatchTypes.RankedPlay;
+using osu.Game.Utils;
 using osu.Server.Spectator.Database;
 using osu.Server.Spectator.Database.Models;
 using osu.Server.Spectator.Hubs.Multiplayer.Matchmaking.Elo;
@@ -152,7 +153,7 @@ namespace osu.Server.Spectator.Hubs.Multiplayer.Matchmaking.Queue
                 await lobby.Remove(state);
         }
 
-        public async Task AddToQueueAsync(MultiplayerClientState state, int poolId)
+        public async Task AddToQueueAsync(MultiplayerClientState state, int poolId, APIMod[] mods)
         {
             // Users should only ever be in one queue at a time.
             await RemoveFromQueueAsync(state);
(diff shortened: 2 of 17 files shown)


osu.Server.Spectator/Database/Models/SoloScore.cs
```diff
@@ -69,6 +69,7 @@ public string data
             BuildID = build_id,
             Passed = passed,
             TotalScore = total_score,
+            TotalScoreWithoutMods = ScoreData.TotalScoreWithoutMods ?? 0,
```
bdach, line 72: I was curious where this `.ToScoreInfo()` thing was being used to verify correctness of this fallback.

It appears to be used in

https://github.com/ppy/osu-server-spectator/blob/dc8c7dbf27ac0c69fef99cc142323368d6f44a4f/osu.Server.Spectator/Hubs/Multiplayer/Matchmaking/MatchmakingMatchController.cs#L179

which is a client structure containing [this logic](https://github.com/ppy/osu/blob/d06762f0e346753cb15e4414bf174c7056388366/osu.Game/Online/Multiplayer/MatchTypes/Matchmaking/MatchmakingRoomState.cs#L89-L107).

Now, two questions:
- With this visibility mod stuff enabled, shouldn't that be using `TotalScoreWithoutMods` in all three places it's currently using `TotalScore` in?
- If yes, then shouldn't `TotalScoreWithoutMods` fallback to `total_score` and not 0... like the `TotalScoreWithoutMods` property of this class already does lower down?

osu.Server.Spectator/Hubs/Multiplayer/Matchmaking/Queue/MatchmakingQueueBackgroundService.cs
```diff
@@ -152,7 +153,7 @@ public async Task RemoveFromLobbyAsync(MultiplayerClientState state)
                 await lobby.Remove(state);
         }
 
-        public async Task AddToQueueAsync(MultiplayerClientState state, int poolId)
+        public async Task AddToQueueAsync(MultiplayerClientState state, int poolId, APIMod[] mods)
```
bdach, line 156: Maybe worth checking that the client isn't shoving some random mods in here (cf. what `RankedPlayMatchController.isUserModAllowed()` is doing).

The additions below check that the mods *exist* on the ruleset, but not that they are *allowlisted for queueing*.

And yeah I realise any bogus mods should be stopped later by checks on `allowed_mods` web-side, this is just for defence in depth.

bdach (commented):
In general, spectator server must become .NET 10 if this is to go any further forward - I tried to do a local full-stack test and was foiled by

```
Project `/Users/bdach/Documents/Work/open-source/osu/osu.Game/osu.Game.csproj` cannot be added due to incompatible targeted frameworks between the two projects. Review the project you are trying to add and verify that is compatible with the following targets:
    - net8.0
```

osu.Server.Spectator/Database/Models/SoloScore.cs
smoogipoo: `MatchmakingMatchController` is the quick play system. So... yeah you're probably right that it should be using `TotalScoreWithoutMods`, but also it's not accessible anyway and I'd earlier prefer to pull it from the game entirely.

As for the fallback, it's very weird, because it differs in implementation depending on what project suits your particular flavour.
- [`osu-queue-score-statistics`](https://github.com/ppy/osu-queue-score-statistics/blob/d75d901fe17aa1760e97d4c4382f7ecb8f9ad3a1/osu.Server.Queues.ScoreStatisticsProcessor/Models/SoloScore.cs#L104): `?? 0`
- [`osu-queue-score-statistis`](https://github.com/ppy/osu-queue-score-statistics/blob/d75d901fe17aa1760e97d4c4382f7ecb8f9ad3a1/osu.Server.Queues.ScoreStatisticsProcessor/Commands/Maintenance/RecalculateModMultipliersCommand.cs#L149): ` == 0`
- [`osu`](https://github.com/ppy/osu/blob/98fb49876c0242fcf649e0250d6f6b3458769a9e/osu.Game/Online/API/Requests/Responses/SoloScoreInfo.cs#L223): no fallback, not even nullable.

It seems like there's 3 modes which this value can take, all of which seemingly have correct usages: `null`, `0`, and `!= 0`.

Help?

bdach: If it's quick play stuff then maybe leave a TODO at least or something. The name threw me off.

> As for the fallback, it's very weird, because it differs in implementation depending on what project suits your particular flavour.

I'm not going to pretend this isn't a mess because it is. The abstraction here is leaky for reasons forced by the recent mod multiplier rebalance (https://github.com/ppy/osu-queue-score-statistics/pull/374). Add to that some shared model abuse on top - usually this happens when some component is hard-coupled to `ScoreInfo` which has the property (correctly) not-nullable so you have to shove *something* in there.

The situation is:
- Client-side `TotalScoreWithoutMods` is not (currently) allowed to be null because it's supposed to always be populated
- Server-side this is not so - the only scores that will have it populated are lazer scores with mods (for stable scores it can be losslessly recovered via the score conversion bojangle, and for lazer scores without mods it's redundant)
  - Note that this means that for *lazer* scores it's safe to have `TotalScore` fall back to `TotalScoreWithoutMods`, but for an *arbitrary score* from an unknown client, it's not.

I hope to rectify this somehow as I work through results screen in client via application of interfaces which should hopefully decouple this enough to stop being painful.

smoogipoo: If I were to add a comment, where would I add it? I was following the first implementation there for `osu-queue-score-statistics` but it seems the issue is several layers deeper and in another project (`osu`). Is a client side comment fine?

bdach: Yeah I'd put one in the client, inside `MatchmakingRoomState`, saying something along the lines "if this ever gets revived and also gets free mod support it's probably incorrect".

PR #523, opened 2026-07-06
Introduce new version of spectator API with explicit score token passing

bdach:
RFC.

Related:
- https://github.com/ppy/osu/issues/24609#issuecomment-1779433085
- https://github.com/ppy/osu/pull/38163#issuecomment-4853215905

Compatibility matrix:

| | old client | new client |
| :-: | :-: | :-: |
| **old server** | 🟢[^1] | 🔴[^2] |
| **new server** | 🟢[^1] | 🟢[^3] |

[^1]: Old client continues to use the old version of the API which did not functionally change and is not aware of the existence of the new API. Spectating and replay upload operational.
[^2]: New client will attempt to use the new version of the API which does not exist on old server. Complete failure of spectator and replay upload flows for all users affected.
[^3]: New client uses the new version of the API which exists on new server. Spectating and replay upload operational.

---

One continuing sticking point in attempts to make online flows more reliable is the serialising factor of spectator server in one part of submission, namely replay saving.

Up until now, the active score token ID was part of `SpectatorClientState`, and relied on as ambient server-side state to make `SendFrameData()` and `EndPlaying()` calls work.

However, only being able to store *one* token ID means that the client cannot move on to start a new play without concluding the last one, which is a problem, as the user essentially cannot start a new play until they're sure the old one concluded. This is a large pain point for users with unreliable network connections.

This PR attempts to improve this bottleneck by allowing clients to send spectator data out-of-order. The general rules are as follows:

- The user can have up to `MAX_STARTED_SCORES = 5` scores started at any given time; their tokens are stored to the state. (Value of constant up for negotiation.)
- All methods now explicitly accept a score token. `null` score tokens are generally ignored for the replay code paths, instead only updating the `SpectatorClientState`.
- The `BeginPlayingV2()` call will always overwrite properties on `SpectatorClientState` that are not `ScoreTokens`.
- Clients are now allowed to interleave `SendFrameDataV2()` and `EndPlayingV2()` calls from multiple scores. In cases of those calls, the ambient `SpectatorClientState` is only updated if the score token supplied is `null` or equal to the latest non-`null` score token stored to the state.

Listening to suggestions with respect to collection of metrics (should the total number of scores started be tracked to a counter?)


osu.Server.Spectator/Hubs/Spectator/SpectatorHub.cs
```diff
@@ -183,28 +188,176 @@ public async Task EndPlaySession(SpectatorState state)
             await endPlaySession(Context.GetUserId(), state);
         }
 
-        private async Task processScore(SpectatorClientState item, Score score)
+        #endregion
+
+        #region V2
+
+        public async Task BeginPlaySessionV2(long? scoreToken, SpectatorState state)
```
peppy, line 195: What concerns me is that a client could potentially call this multiple times in a row without ending previous `PlaySessions`. And sure, this won't break anything, but it makes the flow harder to understand because there's no stringent guardrails in place.

What I'm trying to get is a mental understanding of what we are trying to resolve here.

Working backwards from this code, I'm assuming we're trying to resolve/handle cases like this:

- Client connects
- `BeginPlaySession(token1)`
- ...
- `EndPlaySession(token1)` (send fails)
- Client disconnected, reconnects
-  `BeginPlaySession(token2)`
- Server requests missing frames from token1
- Client sends remaining frames and called `EndPlaySession(token1)` again?
-  `EndPlaySession(token2)`

I guess if I change the way I'm thinking about this API such that `server-spectator`'s handling is a "dumb" implementation that just acts as a bucket and only stores pristine data when it is completely provided, I would be okay with what is in this PR as it stands.

My (weak) concern is that as it allows n async scores to be in flight, it doesn't give much structure to the score submission process when reading through things. Again, probably not an issue and just a mindset thing? Any actual handling would be done in the final `processScore` once we have data in a good state.

I do question how a client could get to the point of having five scores in flight. It feels like there would have to be some real luck to make this happen (each connection failure would have to happen after a `Begin` and before an `EndPlaySession`, without any of the other 4 scores being recovered in that time). Maybe it's more understandable and realistic to set this to 2 max for now?

Let me know if you want to voice to discuss this. Hard to give my thoughts on this one in a single textual review pass so I may have failed.

bdach: Would rather continue discussion on a voice call since there's a lot to unpack here and I'm not really following the concerns.

peppy: Let's do a call tomorrow.

peppy: Resolved concerns in call. @bdach good to do a final review pass and merge this at your discretion (require un-drafting).

bdach: Pushed minor changes as discussed in the call:

- Number of simultaneous allowed scores decreased to 2 (https://github.com/ppy/osu-server-spectator/pull/523/commits/c952c83c3433cd5526a3c40bfb9390e560bc7714)
- Drops of scores from buffer due to overflow are now logged to console and sentry at warning level (https://github.com/ppy/osu-server-spectator/pull/523/commits/6059c1d707eeb5a59c5fa82024343f0c63135fbd)

One remaining concern raised in the call is that there's still a serializing factor in that a score needs to be started by a `BeginPlaySession()` call. It is not possible to just start sending frames. This is a little inconvenient in scenarios where e.g. a player needs to re-send two or more complete plays, as they'd have to e.g. call `BeginPlaySession()` twice - in correct chronological order to boot (due to the latest score being considered the "active" one) - and then start re-sending the frames. This is even more crucial as the entirety of `SpectatorClientState` is dropped on disconnect, so the tokens will not survive a connection drop-out.

That part of it I would rather handle later as I work through these connectivity issues one step at a time.

Undrafted.

peppy (approved):

PR #524, opened 2026-07-13
Improve matchmaking debuggability

smoogipoo:
I have these two changes stashed locally, but it's probably better to get them in the tree once and for all?

1. Allows recent matchups to be bypassed, so that two players can continuously queue into each other.
2. Adds a setting to use a debug set of beatmaps (in this case, survival dance repeated 50 times, as a short beatmap).


README.md
```diff
@@ -40,4 +40,5 @@ For advanced testing purposes.
...
 | `MATCHMAKING_POOL_SIZE`              | The number of beatmaps per matchmaking room.                                                                                                                                                                                                   | 50                              |
-| `MATCHMAKING_QUEUE_ALLOW_BANS`       | Whether to issue bans from the matchmaking queue for various abuse.                                                                                                                                                                            | true                            |
\ No newline at end of file
+| `MATCHMAKING_QUEUE_ALLOW_BANS`       | Whether to issue bans from the matchmaking queue for various abuse.                                                                                                                                                                            | true                            |
```
bdach: Maybe worth documenting the "recent matchup" thing being steered by this envvar too?

bdach (approved):

PR #533, opened 2026-09-16
Update to target .NET 10

peppy:


smoogipoo (approved):

bdach:
I guess since we're doing bare metal on production with this project (as far as I know), make sure .NET 10 is available on the target droplet before any deploy attempt.

PR #534, opened 2026-09-16
Dispose database connections after use

smoogipoo:
I haven't tested this yet, but PRing for early review and potentially to deploy if things get awry again before I'm able to test it.

Two important changes here:
- Dispose the db connections. This should enter a [fast path](https://github.com/mysql-net/MySqlConnector/blob/fdab41bc1646ce8ae89bde23c960a3239d90194b/src/MySqlConnector/MySqlConnection.cs#L1204-L1225) where connections are returned to the pool immediately.
- Change every method that uses `IEnumerable<T>` into `T[]`, so that the resulting enumerable of a `QueryAsync()` does not escape the scope of a connection's lifetime.


bdach:
Am I seeing it right that `getConnectionAsync()` is itself not changed in any way?

https://github.com/ppy/osu-server-spectator/blob/cfb89301adda6df537e3a57c6d59c47b06acb871/osu.Server.Spectator/Database/DatabaseAccess.cs#L864-L877

My naive understanding would have it that the `openConnection` field should no longer exist if we're doing this.

smoogipoo:
Oh. I totally missed that. All this time it hadn't registered for me that it was being stored to a local. Gonna close this PR then, because I'd say that's pretty safe-enough-ish?

bdach:
Well as I said there's still a chance of a race between

https://github.com/ppy/osu-server-spectator/blob/cfb89301adda6df537e3a57c6d59c47b06acb871/osu.Server.Spectator/Database/DatabaseAccess.cs#L866

and

https://github.com/ppy/osu-server-spectator/blob/cfb89301adda6df537e3a57c6d59c47b06acb871/osu.Server.Spectator/Database/DatabaseAccess.cs#L871-L872

that could cause some excess connections to get created and drop some connections on the floor, leaving cleanup for GC.

I'm not really sure if this store-to-local approach is better or worse than the transient create-every-time-then-dispose approach. We do both in places. `osu-queue-score-statistics` [takes out one `MySqlConnection` instance per processed item](https://github.com/ppy/osu-queue-score-statistics/blob/d75d901fe17aa1760e97d4c4382f7ecb8f9ad3a1/osu.Server.Queues.ScoreStatisticsProcessor/ScoreStatisticsQueueProcessor.cs#L194) (though it's [somewhat](https://github.com/ppy/osu-queue-processor/blob/5cc515a176efd9f53d73a242de6e9b05d1522566/osu.Server.QueueProcessor/QueueProcessor.cs#L290) [obscured](https://github.com/ppy/osu-queue-processor/blob/5cc515a176efd9f53d73a242de6e9b05d1522566/osu.Server.QueueProcessor/DatabaseAccess.cs#L37-L50) in `osu-queue-processor` innards).

smoogipoo:
That's true, the one case I can imagine it would come up is if there's a `Task.WhenAll()` or similar, but I can't find anything of the kind.

I can do something like this, but I'm just unsure of the path forward at this point:

```diff
diff --git a/osu.Server.Spectator/Database/DatabaseAccess.cs b/osu.Server.Spectator/Database/DatabaseAccess.cs
index e1a7404..8080d0d 100644
--- a/osu.Server.Spectator/Database/DatabaseAccess.cs
+++ b/osu.Server.Spectator/Database/DatabaseAccess.cs
@@ -5,6 +5,7 @@ using System;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Linq;
+using System.Threading;
 using System.Threading.Tasks;
 using Dapper;
 using Microsoft.Extensions.Logging;
@@ -18,9 +19,11 @@ namespace osu.Server.Spectator.Database
 {
     public class DatabaseAccess : IDatabaseAccess
     {
-        private MySqlConnection? openConnection;
+        private readonly SemaphoreSlim connectionSemaphore = new SemaphoreSlim(1, 1);
         private readonly ILogger<DatabaseAccess> logger;
 
+        private MySqlConnection? openConnection;
+
         public DatabaseAccess(ILoggerFactory loggerFactory)
         {
             logger = loggerFactory.CreateLogger<DatabaseAccess>();
@@ -859,21 +862,31 @@ namespace osu.Server.Spectator.Database
         public void Dispose()
         {
             openConnection?.Dispose();
+            connectionSemaphore.Dispose();
         }
 
         private async Task<MySqlConnection> getConnectionAsync()
         {
-            if (openConnection != null)
-                return openConnection;
+            await connectionSemaphore.WaitAsync();
 
-            DapperExtensions.InstallDateTimeOffsetMapper();
+            try
+            {
+                if (openConnection != null)
+                    return openConnection;
 
-            openConnection = new MySqlConnection(
-                $"Server={AppSettings.DatabaseHost};Port={AppSettings.DatabasePort};Database=osu;User ID={AppSettings.DatabaseUser};ConnectionTimeout=5;ConnectionReset=false;Pooling=true;Pipelining=false");
+                DapperExtensions.InstallDateTimeOffsetMapper();
 
-            await openConnection.OpenAsync();
+                openConnection = new MySqlConnection(
+                    $"Server={AppSettings.DatabaseHost};Port={AppSettings.DatabasePort};Database=osu;User ID={AppSettings.DatabaseUser};ConnectionTimeout=5;ConnectionReset=false;Pooling=true;Pipelining=false");
 
-            return openConnection;
+                await openConnection.OpenAsync();
+
+                return openConnection;
+            }
+            finally
+            {
+                connectionSemaphore.Release();
+            }
         }
     }
 }

```

